PR #2, opened 2026-05-04
Formula: match flat release tarball layout (fixes brew install)

thomas-doesburg:
## Problem

`brew install logic-pro-mcp` fails on every release since v3.1.1 (current: v3.1.4) with:

```
Errno::ENOENT: No such file or directory - docs/SETUP.md
```

## Root cause

The release tarball ships every helper file at the archive root, with no `docs/` or `Scripts/` subdirectories:

```
$ tar tzf LogicProMCP-macOS-universal.tar.gz   # v3.1.4
LogicProMCP
SETUP.md
install-keycmds.sh
uninstall-keycmds.sh
keycmd-preset.plist
LogicProMCP-Scripter.js
```

But the formula's `install` block expects `docs/SETUP.md` and `Scripts/*`, so the first `pkgshare.install` raises `Errno::ENOENT` and the install never completes. SHA256 in the formula matches the tarball, so the binary itself is fine — it's purely a path mismatch.

Confirmed against v3.1.1, v3.1.2, v3.1.3, and v3.1.4. No release pipeline change appears to have shipped a `docs/`+`Scripts/` layout — looks like the formula was authored against an earlier intended layout that never made it into the published artifact.

## Fix

Drop the `docs/` and `Scripts/` prefixes from the five `pkgshare.install` calls. The `caveats` reference to `#{pkgshare}/SETUP.md` is unchanged — `pkgshare.install` already lands files in `pkgshare/`, regardless of source-side path.

## Verification

```
$ brew install --formula logic-pro-mcp
==> Pouring logic-pro-mcp-3.1.4...
==> Caveats
Logic Pro MCP Server is installed at /opt/homebrew/opt/logic-pro-mcp/bin/LogicProMCP.
🍺  /opt/homebrew/Cellar/logic-pro-mcp/3.1.4: 9 files, ...

$ ls /opt/homebrew/opt/logic-pro-mcp/share/logic-pro-mcp/
LogicProMCP-Scripter.js  install-keycmds.sh    keycmd-preset.plist
SETUP.md                 uninstall-keycmds.sh

$ LogicProMCP --check-permissions
Accessibility: ...
Automation (Logic Pro): ...
```

Tested on macOS Sonoma 14.x, arm64, Xcode 16.4.

## Alternative considered

Updating the release pipeline to emit `docs/` and `Scripts/` subdirectories instead would also work and would let the formula stay as-is. Going with the formula change because the flat tarball layout has shipped across four releases and any consumers who curl the tarball directly will already depend on the flat paths.

### DIFF
--- a/Formula/logic-pro-mcp.rb
+++ b/Formula/logic-pro-mcp.rb
@@ -28,11 +28,11 @@ class LogicProMcp < Formula
     bin.install "LogicProMCP"
     # Helper assets shipped with the binary so users can complete Logic Pro
     # integration without re-cloning the repo.
-    pkgshare.install "docs/SETUP.md"
-    pkgshare.install "Scripts/install-keycmds.sh"
-    pkgshare.install "Scripts/uninstall-keycmds.sh"
-    pkgshare.install "Scripts/keycmd-preset.plist"
-    pkgshare.install "Scripts/LogicProMCP-Scripter.js"
+    pkgshare.install "SETUP.md"
+    pkgshare.install "install-keycmds.sh"
+    pkgshare.install "uninstall-keycmds.sh"
+    pkgshare.install "keycmd-preset.plist"
+    pkgshare.install "LogicProMCP-Scripter.js"
   end
 
   def caveats


MongLong0214:
Reviewed top-to-bottom. **The diff is correct.**

## Verification

- Downloaded `LogicProMCP-macOS-universal.tar.gz` from the v3.1.4 release directly. Layout is exactly what the PR body claims:
  ```
  LogicProMCP
  SETUP.md
  install-keycmds.sh
  uninstall-keycmds.sh
  keycmd-preset.plist
  LogicProMCP-Scripter.js
  ```
  All five helper assets at the archive root, no `docs/` or `Scripts/` subdirectories. SHA256 (`fb12d1de…f174`) matches the `Formula/logic-pro-mcp.rb:23` pin exactly.

- `Scripts/release.sh:73-74` is the source of the flat layout — it copies `docs/SETUP.md` and `Scripts/*` into the staging dir without preserving prefixes:
  ```bash
  run "cp docs/SETUP.md '$STAGE_DIR/'"
  run "cp Scripts/install-keycmds.sh Scripts/uninstall-keycmds.sh ... '$STAGE_DIR/'"
  ```
  Four releases (v3.1.1 → v3.1.4) have shipped with this layout. Authoring the formula against the source-tree layout (which never makes it into the tarball) is the actual mismatch.

- `caveats` line `#{pkgshare}/SETUP.md` stays valid — `pkgshare.install` writes to `pkgshare/` regardless of the source-side path.

- `test do` block doesn't touch the helper assets, so it's unaffected.

## CI failure on this PR is unrelated

The build step is failing with:
```
swift-sdk/Sources/MCP/Base/Transports/HTTPClientTransport.swift:485:31:
error: missing argument for parameter 'of' in call
   try await withThrowingTaskGroup { group in
```

This is **not caused by this PR.** `swift-sdk 0.11.0+` adopts the short-form `withThrowingTaskGroup { ... }` syntax that requires Swift 6.2; the workflow's `Xcode_16.2.app` pin gives Swift 6.0, which can't compile that form. `main` has been failing CI for 5 consecutive runs since 2026-04-26 with the identical error — verifiable via `gh run list --branch main --limit 5`.

I'm shipping the fix in #6 (CI/release workflows pinned to `Xcode_16.4.app`, plus the Issue #3 / #4 / #5 read-path overhaul). Once #6 lands, this PR's CI will go green automatically on the next push or rebase — no changes required to this branch.

## Suggested merge order

1. Merge #6 (CI hotfix + read-path fixes).
2. This PR rebases onto `main`. CI re-runs → should pass.
3. Merge this PR.

Or merge #6 + this PR back-to-back from `main` since both diffs are directly orthogonal — #6 doesn't touch `Formula/` and this PR doesn't touch the SDK / workflows.

## Long-term suggestion (out of scope here)

The flat layout is a property of `Scripts/release.sh`'s staging step. Any future restructure (e.g. someone moving `Scripts/release.sh` to emit `docs/` + `Scripts/` subdirectories to match the source tree) would silently break Homebrew install again. Worth either:

- Adding a `tar tzf | head` smoke check inside `release.sh` that fails fast if helper assets aren't where the formula expects them, or
- Sourcing the staging path layout from a constant shared with the formula's `def install` block.

Both are separate PRs — not a blocker for this fix.

Approving as-is. Will merge once #6 unblocks CI on `main`.

MongLong0214:
Merged via admin (`9e43863`) since the CI hotfix was queued behind this PR — base CI couldn't pass without #6's Xcode 16.4 / swift-sdk 0.11+ compatibility fix, and rebasing onto main wasn't possible from this OAuth scope (workflow-file changes need elevated permissions). The diff itself was verified: actual v3.1.4 tarball is flat, sha256 matches the formula pin, and your patch is a clean rename. Future `brew install logic-pro-mcp` should work as-is. Thanks for the precise diagnosis + working patch.

PR #54, opened 2026-06-18
fix: accept Logic 12.2 "Tracks header" description in getTrackHeaders

jdpersina:
`logic://tracks` fell through to Tier 2 placeholder synthesis on Logic Pro 12.2 + macOS 26.3 even when MCU was fully registered and the mixer AX poll was healthy. `AXLogicProElements.getTrackHeaders` Strategy 3 — the AXGroup-by-description matcher — accepted only
`desc.lowercased() == "track headers"` or the Korean `"트랙 헤더"`. Live Logic 12.2 on macOS 26 exposes the rail as `AXGroup desc="Tracks header"` (singular "Tracks", lowercase "header"), which lowercases to `"tracks header"` and missed the existing equality by one letter and a word-order flip with no structural change.

Result chain: `getTrackHeaders` returned nil → `StatePoller.pollTracks` cached nothing → `ResourceHandlers.readTracks` synthesized placeholder rows (`source: "ax_live_with_file_count"`, `name: "Track \(idx + 1)"`, `placeholder: true`). Mixer writes worked because the v3.4.5 Logic-12.2 mixer-matcher fix had landed; only the parallel tracks-resource path was still broken.

Strategy 3 now accepts `track header(s)` and `tracks header(s)` plus the existing Korean string. All four pre-existing strategies preserved (no regression risk against older Logic builds or test fixtures).

Verified live on macOS 26.3.1 + Logic Pro 12.2 — `source` flips to `ax_live`, real track names ("Clean Echoes", "lofi chorus", "Stereo Out", …) surface, and `placeholder` is no longer set on returned rows.

Track type classification on non-master strips still returns `type: "unknown"` on this OS combo — separate AX path, out of scope for this fix. Documented as a known limitation in TROUBLESHOOTING.

Changes
- `Sources/LogicProMCP/Accessibility/AXLogicProElements.swift` — expand Strategy 3's accepted description set.
- `Tests/LogicProMCPTests/AXLogicProElementsTests.swift` — `testGetTrackHeadersAcceptsLogic122TracksHeaderDescription` (the Logic 12.2 form) and `testGetTrackHeadersAcceptsLegacyAndKoreanDescriptions` (regression guard pinning the legacy + Korean spellings).
- `CHANGELOG.md` — `[Unreleased] → Fixed` entry.
- `docs/TROUBLESHOOTING.md` — new "Tracks Resource" section with the placeholder-mode symptom, cause, and resolution; plus a note on the separate `type: "unknown"` follow-up.

Verification per CONTRIBUTING.md checklist
- `swift build -c release` clean.
- `swift test` green: 1279 tests pass (1277 prior + 2 new here).
- New behavior covered by unit tests.
- Read-only resource fix — no API.md / live-verify-*.md change needed.


MongLong0214:
Formal tracking issue: #59. This PR is the resolving change set for the Logic 12.2 Tracks Resource / transport AX readback regression.

PR #233, opened 2026-07-01
fix(#222): set_instrument leaves the Library panel un-drifted after a failed nav

### LINKED ISSUE #222: Bug: inventory/apply contract overstates set_instrument-loadable Library paths
## Summary

`logic://library/inventory` exposes Library browse/catalog rows through
`presetsByCategory`, and downstream demo sweeps treated those rows as
`logic_tracks.set_instrument` loadable software-instrument paths.

That contract is too broad. The bug is not "make the current 411 rows pass as
instrument presets." The bug is that the inventory/apply contract overstates which
Library paths are safe and meaningful `set_instrument` targets.

## Evidence

Latest reproduced sweep:

- Runtime: `origin/main` `f30418141046ff16d3ddfbfae4840d93a1f2b22e`
- Describe: `v3.7.4-3-gf304181`
- Server startup: `logic-pro-mcp v3.7.4 - 10 tools, 18 resources, 7 channels`
- Swift MCP SDK: `0.12.1`
- Logic Pro: `12.2`
- Inventory rows flattened from `logic://library/inventory`: 411
- `set_instrument` verified: 32 / 411
- `set_instrument` uncertain: 16 / 411
- `set_instrument` failed: 363 / 411
- `set_instrument` timeouts: 0
- Follow-up `logic_midi.play_sequence`: 411 / 411 OK

Observed failure pattern:

- All 363 hard failures returned `ax_write_failed`.
- The repeated hint was `Library path not fully resolvable: <path>`.
- The 16 uncertain outcomes were readback mismatches around Electric Piano while the
  observed patch stayed `Deluxe Classic`.
- MIDI worked for every attempt, so this is not a transport, CoreMIDI, server-liveness,
  or sequencing failure.

Important correction:

Some of the 32 "verified" rows look like Library browser folder/subcategory rows, for
example `Synthesizer/Bass`, `Orchestral/Brass`, and `World/Stringed`. Therefore selected
row readback matching the requested final segment is not sufficient proof that a leaf
instrument patch loaded.

## Root Cause

There are three different surfaces currently being treated as one:

1. Catalog inventory
   - Broad read-only data.
   - Useful for planning and narration.
   - Not automatically applicable through the live Library panel.

2. Panel inventory
   - AX-backed and panel-authoritative.
   - Still needs row typing before it can be used for mutation.
   - The current live AX probe intentionally stops at depth 2+ to avoid destructive
     probing, so its flattened output cannot be trusted as a complete loadable leaf set.

3. Apply catalog
   - This does not exist clearly enough yet.
   - It should be the only source for demo sweeps and `set_instrument` acceptance.
   - It should contain only paths proven to be `source: panel`, `kind: leaf`, and
     `loadable: true`.

`set_instrument` then compounds the ambiguity by walking visible AX text rows and using
selected-row readback as the State A signal. That cannot distinguish "folder row was
selected" from "leaf patch actually loaded" unless the path was already classified as a
loadable leaf.

## Expected Contract

- `presetsByCategory` remains a browse/catalog view, not an apply guarantee.
- A new apply-catalog field or resource exposes only `source: panel`, `kind: leaf`,
  `loadable: true` paths.
- `logic_tracks.set_instrument` pre-resolves against that apply catalog before live AX
  mutation.
- Disk-only, folder, missing, uninstalled, and ambiguous paths fail fast with typed
  State C instead of generic `ax_write_failed`.
- State A means a classified leaf path was applied and read back, not merely that a
  selected Library row text matched the requested final segment.
- MIDI playback remains a separate transport check and never proves the requested
  instrument loaded.

## Acceptance Target

Define the denominator explicitly:

- Denominator: `count(source == panel && kind == leaf && loadable == true)`.
- Pass condition: `verified == denominator`.
- Required zeroes: `uncertain == 0`, generic `ax_write_failed == 0`, `timeout == 0`.
- Folder, disk-only, unavailable, and ambiguous entries are reported separately with typed
  reasons and excluded from the loadable-instrument denominator.

Public-safe evidence labels:

- `full-instrument-sweep-20260701-main-1904`
- `final-sweep-report.md`
- `all-instruments-list.csv`
- `all-instrument-sweep-evidence.jsonl`
- `all-instrument-sweep-summary.json`

MongLong0214:
## Summary

This started as the no-drift fix for #222, but the full operator sweep exposed two more contract gaps that belong in the same PR:

- `set_instrument` now leaves the Library panel usable after failed navigation and avoids shortcut drift by opening Library through the View menu rather than relying on mutable key-command mappings.
- Library row navigation is hardened for deep Finder-column paths: category selection resets horizontal scroll, intermediate folders can be selected without committing a patch load, the rightmost preset column is searched by scrolling, and the final leaf is committed with row-level click/double-click fallback plus readback verification.
- `scan_library` now scans both the user Logic Library and Logic Pro app-bundle Instrument patch roots, deduplicates relative `.patch` candidates, trims filesystem padding from leaf names, maps current Logic grouped libraries (`Studio Horns`, `Studio Strings`, `World`, multi-channel kits), and reports `candidatePatchCount` / `nonApplicablePatchCount` instead of silently dropping candidates.
- `resolve_path` now distinguishes browse rows from loadable leaves. Folder/category rows return `loadable:false`, and `set_instrument` fails closed with `folder_not_preset` before live AX navigation when the cached path is not a preset leaf.

## Linked issue

- Closes #222

## Risk

- Priority: P2
- Area: demo / resources / Library AX
- User-visible impact: failed or deep Library patch loads no longer poison subsequent attempts; callers can preflight paths and avoid applying folders or non-applicable patch candidates.
- Contract impact: additive `scan_library` fields (`candidatePatchCount`, `nonApplicablePatchCount`) and a new terminal error code (`folder_not_preset`). Existing no-drift State C errors remain typed and machine-classifiable.

## Verification

- [x] `swift test --no-parallel` — 1967 passed
- [x] `swift build -c release`
- [x] `python3 -m py_compile Scripts/live-e2e-test.py`
- [x] `git diff --staged --check`
- [x] LSP error diagnostics on changed Swift files — no errors
  - SourceKit still reports the repo's existing Swift Testing deprecation warnings in test files.
  - Python LSP (`basedpyright`) is unavailable locally; syntax compile was used for the touched Python harness.
- [x] Live Logic Pro corrected full apply sweep — evidence label `issue222-corrected-5713-from-4178`:

```text
status=succeeded
candidatePatchCount=5716
leafCount=5713
nonApplicablePatchCount=3
combinedVerified=5713/5713
failed=0
firstFailure=null
finishedAt=2026-07-04T02:35:22+0900
lastVerified=World/Stringed/Pipa
```

Non-applicable candidate classification from the corrected scan:

```text
Deluxe Classic                         -> non_applicable:no_panel_taxonomy_route
Template/Empty Pad                     -> non_applicable:no_panel_template_route
Template/Empty Quick Sampler           -> non_applicable:no_panel_template_route
```

## Reviewer focus

- `LibraryDiskScanner` denominator changes: user + app-bundle scan, dedupe, non-applicable accounting, and current Logic taxonomy mapping.
- `resolve_path` / `set_instrument` contract: only leaf paths are loadable; folders fail closed before AX writes.
- `LibraryAccessor.selectPath` sequencing: intermediate folders do not commit patch loads; only the final leaf commits and gets verified.


MongLong0214:
Follow-up pushed and PR body updated for the corrected full #222 sweep.

Latest head: `c82dfc1` (`fix(#222): harden library patch loading sweep contract`)

Verification added to the PR:

- `swift test --no-parallel` — 1967 passed
- `swift build -c release`
- `python3 -m py_compile Scripts/live-e2e-test.py`
- LSP error diagnostics on changed Swift files — no errors
- Corrected live Logic Pro apply sweep: `5713/5713` verified, `failed=0`, `firstFailure=null`
- Corrected scan denominator: `candidatePatchCount=5716`, `leafCount=5713`, `nonApplicablePatchCount=3`

No local absolute paths are included in the public evidence; the sweep is referenced by the public-safe label `issue222-corrected-5713-from-4178`.

PR #270, opened 2026-07-08
feat: support Logic Pro Creator Studio variant

### LINKED ISSUE #269: Feature: Logic Pro Creator Studio variant support
### Preflight

- [x] I searched existing issues for related work.

### Priority

P1 - unlocks a high-value Logic Pro workflow

### Primary area

setup-release

### Problem

As a Logic Pro MCP user on Apple Creator Studio, I need the same MCP server to control **Logic Pro Creator Studio** (`com.apple.mobilelogic`, process name `Logic Pro Creator Studio`) — not only the desktop purchase build (`com.apple.logic10`, `Logic Pro`).

Today the server and helper scripts hard-code desktop bundle IDs and process names. Setup Doctor, TCC checks, AppleScript `tell application`, System Events `tell process`, bootstrap/bounce scripts, and PID fallback all assume desktop Logic Pro. Creator Studio users see failed detection, wrong process targeting, or honest-but-unhelpful "Logic not running" states.

### Proposed outcome

The server should auto-detect and target either Logic Pro variant from a single manifest source of truth (`manifest.json` → `supported_logic_pro_variants`).

- Resolution order: `LOGIC_PRO_BUNDLE_ID` override → frontmost known variant → running → installed (desktop before Creator Studio).
- Bundle ID drives AppleScript `tell application id`, health/TCC/install probes, and doctor output.
- Process name drives System Events `tell process` only.
- Python helpers (`Scripts/logic_variants.py`) and Swift target resolution (`LogicProTarget`) share the same manifest metadata.
- Standalone Swift scripts resolve the running Logic app inline without extra infrastructure.
- Docs (`README`, `docs/SETUP.md`) document both variants and the override env var.

### Safety and contract notes

- **Mutating operation?** No new mutating MCP commands; existing commands behave the same once the correct variant is resolved.
- **Required explicit target?** Optional `LOGIC_PRO_BUNDLE_ID` for machines with both variants installed.
- **Readback source:** Existing health/resources/doctor surfaces; variant label included where Logic install is reported.
- **Failure/uncertain state:** Unknown forced bundle ID is reported honestly (`unknown_forced`); process name is not invented for unknown bundles.
- **Destructive confirmation:** Unchanged.

### Done when

- [x] Public behavior is documented (`README`, `docs/SETUP.md`, `manifest.json`, `server.json`).
- [x] Tests cover the new contract (`LogicProTargetTests`, `logic_variants_test.py`, updated bootstrap/bounce tests).
- [ ] Live Logic evidence is attached if the feature writes to Logic (Creator Studio live validation deferred; desktop regression covered by unit tests).

razvangeangu:
## Summary

- Add manifest-driven support for **Logic Pro Creator Studio** (`com.apple.mobilelogic`) alongside desktop Logic Pro (`com.apple.logic10`).
- Introduce `LogicProTarget` / `LogicProVariantPolicy` (Swift) and `logic_variants.py` (Python) for shared bundle/process resolution with optional `LOGIC_PRO_BUNDLE_ID` override.
- Patch existing call sites (Setup Doctor, TCC, channels, bootstrap/bounce scripts, standalone Swift helpers) in place — no new structural layers.

## Linked issue

- Closes #269

## Risk

- Priority: P1
- Area: setup-release
- User-visible impact: Creator Studio users can install and run the same MCP server; desktop behavior unchanged when only desktop Logic is present.
- Contract impact: docs only + honest variant reporting in doctor/health surfaces; no new mutating MCP commands.

## Verification

- [x] `swift build`
- [x] `swift test --filter LogicProTargetTests`
- [x] `swift build -c release`
- [x] Targeted test: `python3 Scripts/logic_variants_test.py`, `python3 Scripts/logic_session_bootstrap_test.py`, `python3 Scripts/logic_key_event_test.py`, `swiftc -typecheck` on `logic_key_event.swift`, `logic_ui_snapshot.swift`, `logic_ax_button_press.swift`
- [x] Live Logic evidence: not required / attached below (Creator Studio live validation deferred; unit tests cover resolution contract)

Evidence:
```text
LogicProTargetTests: 15/15 passed
logic_variants_test.py: 7/7 passed
logic_session_bootstrap_test.py: 25/25 passed
logic_key_event_test.py: 7/7 passed
```

## Release readiness

- [x] Public claims, API docs, troubleshooting, and changelog are updated if behavior changed (`README`, `docs/SETUP.md`, `manifest.json`, `server.json`).
- [x] Mutating or destructive behavior has explicit target validation, confirmation, and readback (unchanged; variant resolution only).
- [x] Known limitations are documented in the PR or docs (live Creator Studio E2E not yet attached).

## Reviewer focus

- Resolution order and honest handling of unknown forced `LOGIC_PRO_BUNDLE_ID`.
- Manifest as single source of truth vs hard-coded bundle IDs in standalone Swift scripts.
- Setup Doctor / TCC redaction for `com.apple.mobilelogic`.


MongLong0214 (changes requested):
Thanks for the PR. I reviewed this as a production-readiness gate and I don't think it is merge-ready yet.

Blocking findings:

1. `Scripts/logic_variants.py`'s JXA process resolver selects the first `System Events` object specifier, not the first existing process. In JXA, `se.processes.byName("missing")` returns a non-null object specifier whose `exists()` is false; the current `candidate !== null` check therefore picks desktop `Logic Pro` before trying `Logic Pro Creator Studio`. Creator Studio-only paths such as save-panel snapshots, controller-learn/free-tempo probes, and bootstrap blocking-dialog checks can still miss the Creator Studio process. The resolver should only select a candidate after an existence/readability check.

2. The Python System Events helpers do not implement the issue's requested target resolution order. Without `LOGIC_PRO_BUNDLE_ID`, `logic_process_osa*` iterates manifest order, so desktop is preferred even when Creator Studio is frontmost; for no-stdout mutations like `key code ...`, a successful empty-output send is treated as "not found" and the same action can be attempted against the next variant too. That can target the wrong app when both variants are running.

I verified the Swift target tests and the Python tests in this PR pass locally, but these two cases are not covered by the test suite. Please add coverage for missing-first JXA process / Creator-Studio-only fallback and for frontmost/running Python resolution before this merges.

MongLong0214:
Adding a bit more context so the requested changes are easier to act on. This is not meant to prescribe the exact implementation; I would be comfortable with any approach that preserves the safety invariants below.

The invariant I am trying to protect is: before any UI read/write is sent through System Events, the code should resolve one concrete Logic target, and it should be clear why that target was selected. The Swift side already seems close to this shape with this order:

1. explicit `LOGIC_PRO_BUNDLE_ID`
2. frontmost known Logic variant
3. running known Logic variant
4. installed known Logic variant
5. fallback/default variant

My concern is that the Python/JXA helpers do not yet follow the same contract, so some helper paths can still behave like “try desktop first, then maybe Creator Studio”. That is risky for UI automation because the wrong app can receive the event when both variants are present.

A possible fix direction:

- Make the JXA resolver test actual process existence, not object-specifier non-nullness. In JXA/System Events, `processes.byName("missing")` can return an object specifier even when the process does not exist, so the useful predicate is closer to `candidate.exists()`.
- Add a Python-side resolver with the same target order as the Swift resolver, or otherwise share/generated the same policy so the two surfaces cannot drift.
- Split query-style helpers from action-style helpers. For reads, non-empty stdout may be a reasonable success condition. For UI actions/key commands, empty stdout can still be a successful action, so the helper should not fall through to the next variant after a zero-exit empty-output action.

Suggested coverage, not necessarily the only acceptable set:

- desktop process missing, Creator Studio process present -> JXA chooses Creator Studio
- both variants running, Creator Studio frontmost -> Python resolver chooses Creator Studio
- action helper returns success with empty stdout -> it does not attempt the second variant
- query helper keeps explicit empty/no-data behavior separate from action success

The exact API shape is flexible. The important part for me is that the final behavior is deterministic, covered by tests, and cannot accidentally send a mutating UI event to both Logic variants.

razvangeangu:
Thanks for the detailed review — agreed on all three blocking points. Pushed fixes:

1. **JXA process selection** — `jxa_find_process_snippet` now requires `candidate.exists()` before selecting a process, so a missing desktop Logic object specifier no longer blocks Creator Studio–only hosts.
2. **Python target resolution** — added `resolve_bundle_id` / `resolve_logic_target` matching the Swift order (`LOGIC_PRO_BUNDLE_ID` → frontmost known → running known → installed known → fallback). System Events query/action helpers and activate/probe paths resolve a single target up front instead of iterating manifest order.
3. **Action vs query** — mutating helpers treat exit-0 / empty stdout as a completed send to that one process and do not try the next variant. Query helpers still return empty/no-data without escalating to another variant.
4. **Coverage** — added unit tests for Creator Studio–only JXA selection, frontmost Creator Studio when both are running, and action/query no-fallthrough on empty stdout.

Local `logic_variants_test` (+ related bounce/bootstrap helpers) pass. Ready for another look.

MongLong0214 (approved):

Issue #287, opened 2026-07-10
ADR-004: Bounded mutation Saga — implemented opt-in surface, not universal rollback


MongLong0214:
# ADR-004 — Bounded mutation Saga and verified compensation

> **SSOT revision: 2026-09-21. Implemented bounded core; keep closed.** This body is the current scope and maintenance contract. #308 is the scheduling index. Prior prototype APIs and blanket foundation-order instructions are superseded.
>
> Source baseline: `d067e91958ad3ff1a1bdfb1e6e39a3e23e786517`. Historical live receipts are not a fresh execution at this baseline.

## Decision

Keep the existing Saga implementation for explicitly supported reversible operations. It coordinates preflight, ordered execution, write-boundary accounting, session-local journaling, verification and selective compensation. It is not an ACID transaction, crash-persistent job system, or universal rollback mechanism for arbitrary Logic actions.

**Configuration is explicit:** `LOGIC_MCP_ADR004_MUTATION_SAGA == "1"` enables the feature at this baseline. It is not default-on. `SystemDispatcher` constructs `MutationSaga` using that default. Do not flip the flag simply to make an old documentation sentence true.

## Actual public interface

The existing interface is under `logic_system`:

- `saga_preflight` and `saga_execute`: `steps` plus `idempotency_key`.
- `saga_status` and `saga_cancel`: `idempotency_key`.

The original `logic_project.plan_mutations`, `execute_mutation_plan`, `get_mutation_session`, `cancel_mutation_session` and `compensate_mutation_session` names are **superseded design sketches**, not additional commands that must now be created.

The journal is session-only. A process restart is not durable recovery. Use the current step parser and reversibility classifications to enumerate supported operations; do not assume a new operation is compensable because its name resembles a supported one.

## Implementation history

[The closure receipt](https://github.com/MongLong0214/logic-pro-mcp/issues/287#issuecomment-5006692283) records a real two-step volume/pan Saga with the second step forced to fail before writing. The first write was compensated and independently read back, the result retained the overall partial-failure disposition, and reuse of the idempotency key did not execute or compensate again. PR #399 supplied the historical fault-injection seam.

This supports a bounded implemented core, not all future plug-in/routing/tempo/pitch workflows. The old broad checkpoint proposal was not delivered as a general managed-copy service. Automatic insertion/routing rollback is not authorized by this closed issue.

## Source and tests

[...]

PR #378, opened 2026-07-17
feat(adr-004): preflight rejects route/confirmation/deadline-budget before step 1 (#287)

MongLong0214:
## Summary

The ADR-004 receipt audit found three of the six required full-plan preflight reject classes missing — most notably **there was no confirmation gate at all** (`.awaitingConfirmation` was advanced through unconditionally):

- **`confirmationRequired`** — the registry's `ConfirmationPolicy` is consulted per step; the saga wire carries no confirmation flow, so plans containing confirmation-requiring steps reject before step 1. Defense-in-depth today (the reversible allowlist is all `.none`); fails closed the day the allowlist grows an l1+ op. Proven reachable via a `project.open` (l2) step.
- **`routeUnavailable`** — an injected route-health probe runs per step before anything executes; the public dispatcher wires the real probe (accessibility-channel health — every reversible-allowlist op executes through it) at both construction sites.
- **`deadlineBudgetExceeded`** — worst-case sum of per-step registry deadlines vs the registry-derived saga-execute budget (13 short steps = 325s > 300s rejects; 12 fit exactly). Never a local constant.

Wire mapping gains the three typed codes; three discriminating tests assert **executor run count stays zero** per class; README ADR-004 updated to six-classes-enforced.

## Verification
Full suite `--no-parallel`: **2,835 passed**; branch is off current main (`934112b`).


MongLong0214:
## Adversarial gate — PROCEED-WITH-FIXES → fixes applied at `fccb4e3`

An independent adversarial review of the first cut returned PROCEED-WITH-FIXES with two required findings, both now fixed:

1. **Fail-open route probe (HIGH)** — the optional `routeAvailable` parameter defaulted to a silent skip; any future `MutationSaga` construction site omitting the probe would drop route checks in production. The probe is now a **required** parameter (unit contexts pass an explicit always-true closure; both public dispatcher sites wire the real accessibility-health probe).
2. **Dishonest deadline model (HIGH)** — summing one registry deadline per step under-counts real wall-clock (before/after readbacks, availability capture, compensation are unmodeled) and `>` allowed a zero-headroom exact fit. The model is now 2× per-step deadline (write + verification readback) with `>=` rejection, and the code comments state plainly that this bounds obvious overcommit without guaranteeing completion — compensation is not reserved and a pathological plan can still time out mid-flight.

Also noted by the review and accepted as documented scope: the confirmation class is defense-in-depth (the current reversible allowlist is all `ConfirmationPolicy.none`, so it is unreachable on the production allowlist until the allowlist grows — the test proves reachability via a non-allowlisted l2 step); a boundary control test now proves the budget gate discriminates (5-step plan passes, 6 rejects).

Full suite `--no-parallel`: **2,836 passed** at `fccb4e3`.

Issue #288, opened 2026-07-10
ADR-005: Operation trace and local support bundle — implemented maintenance contract


MongLong0214:
# ADR-005 — Operation trace and privacy-safe support bundle

> **SSOT revision: 2026-09-21. Implemented; keep closed.** This body replaces the old proposed-state description. #308 owns programme scheduling. Earlier receipts remain evidence for their stated historical commits, not current-head qualification.
>
> Reconciled against source baseline `d067e91958ad3ff1a1bdfb1e6e39a3e23e786517`; no new live benchmark was run for this editorial update.

## Decision and actual delivered scope

Maintain the existing bounded operation-trace store and local support-bundle flow. Tracing is default-on (`LOGIC_MCP_ADR005_OPERATION_TRACE != "0"`); the exact `"0"` is the compatibility kill switch.

The live consumer surface includes result `trace_id`, trace queries and explicit support-bundle export. Query/export/delete behavior remains governed by the existing operation registry and confirmation rules. Do not turn diagnostic recording into automatic audio/project upload or a new external observability platform.

The bounded diagnostic ring is **not** a permanent audit log. The trace-clear audit receipt is a separate facility. A missing diagnostic event alone is not proof that no external effect occurred; retry decisions must use the operation's actual write-boundary/outcome contract.

## History and closure

- The earlier progress receipt left three explicit gaps: live write-boundary truthfulness, measured overhead/promotion, and actual consumer use.
- PR #396 corrected boundary placement; PR #397 (`25ee0e6`) promoted default-on behavior.
- [Final receipt](https://github.com/MongLong0214/logic-pro-mcp/issues/288#issuecomment-5005697612) records the default-absent live result → `get_trace` → local bundle chain, allowlist behavior and bounded storage.
- Its microbenchmark and test totals are historical measurements, not current performance guarantees. Re-measure if this path changes.
- Follow-up privacy hardening and exact-artifact qualification are not reasons to pretend this implemented core is still unbuilt. #284 / #373 own release-wide qualification.

## Source entry points

- [Feature flags](https://github.com/MongLong0214/logic-pro-mcp/blob/d067e91958ad3ff1a1bdfb1e6e39a3e23e786517/Sources/LogicProMCP/Utilities/FeatureFlags.swift).
[...]

PR #382, opened 2026-07-17
feat(adr-001)+fix(adr-005): desktop-only qualification matrix + support-bundle containment (#284, #288)

MongLong0214:
## Summary

Three commits (the branch's earlier commits already landed via #380):

**1. Desktop-only qualification matrix — Creator Studio out of product scope (`5d6bcc3`)**
Owner decision: Creator Studio is permanently unsupported. The required same-artifact matrix is now derived from an explicit `shipVariants = [.desktop]` allowlist → **desktop × {en-US, ko-KR} = 2 axes**. Modeling reviewed adversarially: creator is *excluded from required* rather than perpetually waived (required matrix = ship claims; waivers = temporary inability, never product scope). The `creatorStudio` enum stays for honest not-installed health reporting. A contract test pins the exact 2-axis set (variant re-entry or locale drop fails CI); `required-matrix-axes:2` across workflow + contract. Qualification tests rebased off creator observed-axes with no assertion weakened; the axis-vocabulary test now proves a creator run yields **no promotable creator axis**.

**2. Support-bundle path containment — PRD-011 (`0113d14`)**
`export_support_bundle` `dir` must resolve **strictly inside** `~/Library/Logs/LogicProMCP/support-bundles`. realpath(3)-based resolution (deepest existing prefix + literal tail) defeats dot-dot and symlink escapes and the Foundation `/private`-prefix canonicalization divergence. Discriminating tests: relative/abs-inside accepted; outside, dot-dot, root-itself, symlink-escape rejected.

**3. Adversarial-gate hardening (`ce4b286`)**
The gate review returned PROCEED-WITH-FIXES; all required findings closed: the test root override is **DEBUG-only** (was a production containment bypass — P0), realpath failure now **fails closed** (was fail-open — P1), a **post-create TOCTOU recheck** removes and refuses a bundle directory that escaped between check and creation (P1), and a new promotion-gate proof shows an observed **creator host with no desktop evidence can never promote** (required set is static, never collapses to empty — P0 proof).

## Verification
Full suite `--no-parallel`: **2,845 passed**; qualification gate 38/38, runner 77/77, bundle+gate focused 47/47. CHANGELOG documents the ship-scope change; ADR-001 README section + STATUS §38 record the decision.


MongLong0214:
## Order 4 independent exact-main gate: FAIL

Exact main reviewed: `942693b4411d6b7d467e2ea64a0432321da5e113`.

The aggregate completion receipt cannot promote Order 4 to terminal because the support-bundle lane does not satisfy the controlling security acceptance.

- Publication is still path-based: staging creation, per-file `Data.write`, `renameatx_np` with `AT_FDCWD`, and path-based readback.
- There is no directory-fd-relative creation/write/readback, no final-component `O_NOFOLLOW`, and no opened-directory identity verification.
- The cited tests cover lexical/canonical containment, pre-existing destinations, and competing publishers. They do not exercise a final-component symlink swap or parent-directory replacement between validation and publication.
- The current post-publication readback/removal path is not a confidentiality-safe substitute: an attacker-controlled replacement can expose the bundle before the postcheck reacts.

Required correction: implement directory-fd-relative, no-follow, identity-checked publication under the trusted root; add deterministic adversarial coverage for final-component symlink, parent replacement/rename TOCTOU, outside-root escape, and crash/durability behavior; then rerun focused/full, release artifact/hash, security review, exact-head CI, and independent gate.

The already-verified fault-seam and bounded-saga lanes remain valid. Order 4 as a whole remains non-terminal, and Order 5 implementation stays blocked.

PR #383, opened 2026-07-17
fix(adr-005/prd-015): durable append-only audit receipt for trace deletion (#288)

MongLong0214:
## Summary

Debt PRD-015 (P1): a confirmed `clear_traces` destroyed the in-process trace evidence store with no durable record surviving the clear.

**Receipt (commit 1):** every successful clear first appends one JSON line to `~/Library/Logs/LogicProMCP/audit/trace-clear.log` (O_APPEND, never truncated): schema `logic_pro_mcp_trace_clear_receipt.v1`, `cleared_at`, count, `authorization: mcp_confirmed_param`, `server_version`, `pid` — counts/timestamps only, never trace contents or paths. **Receipt-write failure fails closed**: typed State C `trace_clear_receipt_failed` (terminal), store left intact — evidence is never destroyed without its receipt. Success body returns `receipt_path` + `cleared_trace_count`. The audit-root override is DEBUG-only.

**Adversarial hardening (commit 2):** the gate flagged count dishonesty (count and clear straddled an actor suspension point) — the store now **drains atomically** (`clearReturningCount()` reports what it actually removed), the receipt carries the honest `trace_count_at_receipt`, and a divergence appends one **amendment line** (`amends:true`, `actual_cleared_count`) — the first line is never rewritten. Failure extras disambiguated to `store_cleared:false`; the audit directory is realpath-verified against its intended parent before any append (symlinked final component → fail closed; guard **mutation-probed** — neutering it makes the symlink test fail with the clear succeeding). Unbounded-append acceptance documented: a rotator that can delete lines is itself the evidence-destruction surface this receipt closes.

Authorization deliberately remains the registry's l2 `confirmed` gate (local single-user server); the receipt records the authorization source so future operator-approval integration slots in.

## Verification
Full suite `--no-parallel`: **2,853 passed**; focused TraceClear/OperationTrace/SupportBundle 36/36; merged with main (`834c25b`).

### DIFF
--- a/Sources/LogicProMCP/Dispatchers/SystemDispatcher.swift
+++ b/Sources/LogicProMCP/Dispatchers/SystemDispatcher.swift
@@ -814,6 +814,74 @@ struct SystemDispatcher: OperationTraceDispatching {
             .standardizedFileURL
     }
 
+    /// PRD-015 (ADR-005 #288): the directory the `clear_traces` audit receipt is
+    /// written under, OUTSIDE the in-process trace store it records the clearing
+    /// of. Overridable for tests via a DEBUG-only environment hook (never
+    /// honored in a release binary, mirroring `supportBundleRoot`) so the
+    /// fail-closed / append-only behavior is provable against a temp root.
+    static var auditLogRoot: URL {
+        #if DEBUG
+        if let override = ProcessInfo.processInfo
+            .environment["LOGIC_MCP_AUDIT_LOG_ROOT_OVERRIDE"],
+            !override.isEmpty {
+            return URL(fileURLWithPath: override, isDirectory: true).standardizedFileURL
+        }
+        #endif
+        return FileManager.default.homeDirectoryForCurrentUser
+            .appendingPathComponent("Library/Logs/LogicProMCP/audit", isDirectory: true)
+            .standardizedFileURL
+    }
+
+    /// PRD-015 receipt schema tag. Bumped only on a breaking field change so a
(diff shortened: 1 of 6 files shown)


MongLong0214:
## Order 4 independent exact-main gate: FAIL

Exact main reviewed: `942693b4411d6b7d467e2ea64a0432321da5e113`.

The aggregate completion receipt cannot promote Order 4 to terminal because the trace-clear audit lane still fails the controlling final-component and durability acceptance.

- `trace-clear.log` is opened by path with `O_WRONLY | O_APPEND | O_CREAT`.
- The open omits `O_NOFOLLOW` and is not relative to a verified directory descriptor, so a final-component symlink can redirect the append outside the audit root after the directory containment check.
- The opened file is not identity-checked with `fstat`.
- The append does not call `fsync`/`fdatasync`, so the code cannot claim crash-durable evidence before clearing the trace store.
- The cited symlink test covers the audit directory symlink, not the final `trace-clear.log` component. The ordinary success test does not prove crash durability.

Required correction: open the audit root and final file directory-fd-relative with no-follow semantics, verify identities/types, perform a complete append, sync the file and required directory metadata before clearing, and add deterministic final-component symlink/replace and durability-negative tests. Then rerun focused/full, release artifact/hash, security review, exact-head CI, and independent gate.

The already-verified fault-seam and bounded-saga lanes remain valid. Order 4 as a whole remains non-terminal, and Order 5 implementation stays blocked.

Issue #373, opened 2026-07-16
ADR-001 execution: reconcile current qualification coverage and wire independent release enforcement


MongLong0214:
# ADR-001 execution — Current coverage and independent release enforcement

> **SSOT revision: 2026-09-21. OPEN / selected release work.** This is the operational implementation issue for #284. It includes the outstanding enforcement requirements explicitly transferred from #815. #816 owns externally verifiable publication; #290 owns the atlas-specific evidence adapter. #308 owns scheduling. Do not duplicate this work under a new qualification epic.
>
> Audit anchor: `d067e91958ad3ff1a1bdfb1e6e39a3e23e786517`. The revision reconciles source and recorded history; it is not a new live qualification. Read current main and active related PRs before editing.

## Required result

Use the restored runner/transport/oracles to produce and verify current-operation evidence for the **actual release candidate and advertised configuration**, then make an independently trusted verification result load-bearing before publication. A current correct positive case and adversarial rejection must both work.

Do not wait for every backlog ADR. Future Full Compressor, generic third-party control, Flex Pitch and Smart Tempo are not implicit requirements for qualifying today's supported surface. Existing registered but unsupported operations need explicit truthful dispositions rather than fake capability passes.

## History and obsolete instructions

- #367/#372 established the original foundation and rejected blanket waivers. That is not evidence that today's release workflow is already enforcing qualification.
- #815's [September 12 transfer](https://github.com/MongLong0214/logic-pro-mcp/issues/815#issuecomment-5642450091) moved the outstanding signed-bundle/verifier/release-gate acceptance here. #815 closed the diagnosis, not the implementation.
- September 13 removal and 'retire/supersede' comments were reversed by the owner's [reopen-and-implement decision](https://github.com/MongLong0214/logic-pro-mcp/issues/284#issuecomment-5657124698). #926 / `5519079e` restored the subsystem on September 19. No new from-scratch harness is required.
- [The September 14 progress receipt](https://github.com/MongLong0214/logic-pro-mcp/issues/373#issuecomment-5665996597) reports mutation/restore recipes and 41/115 historical passes. It explicitly contradicts the old header saying Phase-B mode does not exist. It also leaves arm behavior inside the sweep unexplained.
- That receipt's category totals and qualified-operation prose are not a current consistent census; derive disjoint operation sets from the actual registry and evidence. Do not add overlapping historical counts or treat a listed operation as passed when the same receipt says its sweep failed.
[...]

PR #386, opened 2026-07-17
qualification: #373 Phase A — semantic oracles for the 20-op read-only surface

MongLong0214:
## Summary

#373 **Phase A** — the qualification runner's semantic validator covered exactly one operation (`system.health`); the other 20 read-only operations fell through to `protocolSmoke` ("the transport worked, nobody checked the meaning"), so the read-only surface could never qualify and PromotionGate could never pass.

**Data-driven oracle engine** (`SemanticOracles.swift`): declarative constraints (`valueEquals` / `numericRange` / `enumMember` / `nonEmptyArray` / `typedField`) resolved over dotted key paths against the parsed response; absent keys fail closed; a bare-scalar body is addressable as the root path. Booleans are distinguished from numeric `1` via `CFBooleanGetTypeID` so a pinned `true` cannot be forged. Presence-only oracles are forbidden by a strength meta-test: every oracle carries at least one VALUE constraint or is an explicitly reasoned `.custom`.

**Oracle table** (`SemanticOracleTable.swift`): one oracle per read-only registry spec — 13 declarative, 7 `.custom` with documented irreducibility reasons (3 prose bodies, 1 bare-scalar body, 2 response↔readback cross-checks, 1 conditional contract). The ratified ≤5 custom budget was replaced by something stricter: an explicit `sanctionedCustomOperationIDs` allowlist enforced by set-equality meta-test (a new hatch cannot appear — and a stale sanction cannot linger — without failing CI).

**Registry reconciliation (registry is truth)**: the planning list was wrong on two names — `edit.select_all` is registered mutating (excluded), `system.clear_traces` is registered read-only (included). Census meta-tests pin both directions (`readOnlySpecs − oracles == ∅`, `oracles − readOnlySpecs == ∅`) plus the 20-operation surface count.

**Mutation harness**: every oracle must fail each corrupted variant of its fixture payload (key dropped / value flipped out of domain / type swapped). Harness falsification-tested: neutering constraint evaluation to always-true produced 184 "oracle survived mutant" detections; reverted clean.

**Fixture counts**: `passed 1→21`, `protocolSmoke 20→0` (structurally unreachable for read-only ops now), `notQualified 86` unchanged (mutating surface, ADR-001-c). Two legacy tests whose premises this phase voids were inverted to pin the new truth, including: a stale waiver for a now-qualifying operation is rejected as `waiverForPassingCase`.

Known Phase-B gaps (tracked on #373): probe params leave 6 read-only ops unable to reach `passed` live (they error before the oracle is consulted — honest failure, never a false pass); `.crossCheck`/`.implies` constraint cases to shrink the custom hatch 7→3.

**Adversarial gate round (applied pre-merge):** the gate confirmed the deviations and raised 5 P1 / 3 P2, all resolved: `midi.list_ports` cross-check honestly re-labeled same-handler echo/transport-consistency (the resource re-invokes the same handler — no independent MIDI probe exists today); `plugins.get_inventory` had an internal contradiction (allowed State B while requiring the `plugins` array State B omits — would false-fail every honest B) — resolved STRICTER: only State A qualifies (State B is by construction unverified; passing it would launder "couldn't check" into "checked"); `tracks.list_library` fails closed on degraded/cache-miss readback, types preset lists, rejects vacuous empty agreement; the custom-oracle mutation harness now requires typed mutant classes per op (malformed / well-formed-but-wrong / readback-divergence for cross-checkers) instead of a gameable count; `system.permissions` labels match the closed 4-remainder set exactly (`granted_evil` mutant-guarded); every oracle must render a non-nil verdict on its own fixture; `audio.analyze_file` accepts `{pass,warn}` only and its fixture is a realistic pass body (a fail body always carries `isError` at the dispatcher).

## Verification
Sanctioned filter: **143/143** (reviewer independently re-ran pre-fix at 142/142). Adjacent suites 128/128. Harness falsification-tested end-to-end: neutering `evaluate()` produced **184 declarative + 38 custom** survived-mutant detections (242 issues); reverted clean. Full suite `--no-parallel` green on the exact PR head. Adversarial gate verdict + fix confirmation recorded in PR comments.

### DIFF
--- a/Sources/LogicProMCP/Qualification/QualificationTransport.swift
+++ b/Sources/LogicProMCP/Qualification/QualificationTransport.swift
@@ -285,8 +285,16 @@ enum QualificationSemanticReadbackValidator {
     ) -> Bool? {
         switch OperationID(rawValue: operationID) {
         case .systemHealth:
+            // Bespoke: full typed-payload equality against the independent read.
             return healthMatches(responseData, readbackData)
-        default:
+        case .some(let id):
+            // #373 Phase A: the read-only surface is covered by the data-driven
+            // oracle table. Only an operation with no oracle at all falls
+            // through to nil, which the runner honestly records as
+            // protocolSmoke rather than passing it.
+            guard let oracle = SemanticOracleTable.byOperationID[id] else { return nil }
+            return oracle.evaluate(responseData: responseData, readbackData: readbackData)
+        case .none:
             return nil
         }
     }
--- a/Sources/LogicProMCP/Qualification/SemanticOracleTable.swift
+++ b/Sources/LogicProMCP/Qualification/SemanticOracleTable.swift
@@ -1,0 +1,613 @@
+import Foundation
+
+// #373 Phase A — the oracle table.
+//
+// Every entry was derived by reading the operation's dispatcher handler and
+// pinning what that handler ACTUALLY emits. Comments cite the handler so a
+// future shape change is reconciled here rather than silently passing.
+//
+// Census invariant (meta-tested): the table's keys are exactly the registry's
+// read-only specs whose availability is not `.unsupported`, minus
+// `system.health` which keeps its bespoke full-payload equality validator.
+//
+// ── KNOWN GAPS (Phase A ships with these open; carried as Phase B debt) ──────
+//
+// 1. FIXTURE↔HANDLER DRIFT HAS NO AUTOMATIC GUARD. Every oracle here was
+//    derived by READING its dispatcher, and the payloads in
+//    SemanticOracleFixtures were hand-written to match. Nothing mechanically
+//    binds the two: if a handler changes a key tomorrow, the fixture keeps the
+//    old shape, the oracle keeps passing the stale fixture, and CI stays green
+//    while the live surface has drifted. Phase B should close this by emitting
+//    golden samples from the handlers' own encode paths, or by landing the
+//    live-run drift catcher. Until then, treat green fixtures as evidence the
+//    ORACLE LOGIC works — not as evidence the shapes are current.
+//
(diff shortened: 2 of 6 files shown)


MongLong0214:
## Gate verdict — PASS (five-lane, abbreviated)

1. **Acceptance vs #373 Phase A**: engine + oracle-per-read-only-op + both-direction registry census + strength meta-test + mutation harness — all present; registry reconciliation corrected the planning list (select_all mutating, clear_traces read-only). ✅
2. **Code quality**: fail-closed key resolution; CFBoolean vs numeric-1 distinction; allowlist under set-equality test (stricter than the ratified ≤5 cap — deviation ratified). ✅
3. **Adversarial review**: independent gate raised 5 P1 / 3 P2 — all resolved pre-merge (State-B non-qualification pinned strict; same-handler echo honestly relabeled; degraded-readback fail-closed; typed mutant classes; exact permission labels; non-nil fixture verdicts; pass/warn-only audio). Harness falsification-tested: 184+38 survived-mutant detections when neutered, clean on revert. ✅
4. **Live lanes**: fixture `passed==21` is validator-logic evidence ONLY — 6 ops are probe-unreachable and 3 environment-conditional live (documented KNOWN GAPS); live qualification remains the forced-on batch's job. Honest, non-blocking. ✅
5. **Repo context**: exact head = merge of main `2e8a299`; full suite `--no-parallel` **2,905 passed** on this head; no unrelated files staged (worktree legacy evidence excluded). ✅

Known Phase-B debts recorded in-file and on #373: probe-params gap (6 ops), fixture↔handler drift guard, `.crossCheck`/`.implies` to shrink the custom hatch 7→3, CoreMIDI-direct independent probe.

PR #387, opened 2026-07-17
saga: LPMCP-PRD-005 — two-tier journal, bounded reclaim without silent replay-protection lapse

MongLong0214:
## Summary

Debt **LPMCP-PRD-005** (P1, ADR-004 #287): the saga journal — the mutation idempotency/replay-protection record — was a single bounded map that turned permanently `capacityExceeded` at 1,024 records: availability death for long sessions and the 30-minute soak, with the only alternatives being silent eviction (re-opens double-execution on retry of an evicted key — a wrong-write class for a DAW) or the status quo.

**Design (adversarially debate-ratified before implementation):** split the journal into two tiers so replay protection NEVER silently lapses in a session while memory stays bounded:

- **Compact idempotency set** — `key → {plan_hash, terminal_kind, sequence}`, **append-only within a generation** (never dropped in-session), capacity 65,536. Plan identity survives body eviction as a SHA256 over the `sortedKeys` JSON of the canonical plan (structural, no delimiter concatenation), so a same-key/different-plan retry still answers `.conflict` exactly as full-plan equality does for live keys.
- **Outcome-body tier** — full stored outcomes for in-flight + terminal records, bounded at the existing 1,024 **combined** (preserves today's total memory ceiling). Under admission pressure the oldest-by-sequence *terminal* body is evicted; in-flight sagas are never evictable, and when nothing is evictable the journal fails closed with the unchanged `capacityExceeded`.
- **Typed honesty for the gap:** a retry of a body-evicted key answers new `saga_outcome_unavailable` (State C, `terminal_kind`, `outcome_retained:false`, `safe_to_retry:false`, `write_attempted:false` — per-call semantics) with a hint that explicitly says the terminal kind names the termination path, NOT intent success, and directs reconciliation via live reads. It never re-executes and never replays a body it no longer holds. `BeginResult`/`PlanDisposition`/`CancelResult` gain `.outcomeEvicted(terminal:)` — `completed` is never overloaded body-less.
- **No wall-clock TTL anywhere** (pressure is the trigger; correctness is time-independent). **Restart durability is out of scope by existing contract** (`journal_survives_process_restart:false` is already on the wire and live-attested); a durable journal is its own future increment.
- **Observability:** `saga_status`/`saga_preflight` expose `journal_full_body_count` / `journal_compact_count` / `journal_body_evictions` (monotonic) / both capacities — ops/test metrics, explicitly not a caller-recovery API.
- `MutationSaga.sessions` was investigated and is request-local in production wiring (fresh instance per preflight/execute); instead of dual-bounding it, two complementary tests pin request-locality behaviourally (replay-after-clear re-executes; a deliberately shared instance provably would suppress that re-execution — jointly excluding a shared-instance regression).

RED-first: 6 new behaviors captured failing against the old API (18 assertion failures) before implementation. A legacy test that pinned the old debt in its own name ("rejects new keys **without evicting** completed outcomes") is inverted to pin the new fail-closed boundary (compact-tier exhaustion).

**Adversarial gate round (applied pre-merge):** verdict PROCEED-WITH-FIXES; the one P1 — a new `abandon(_:)` method with zero callers and zero tests that had become an escape door from the append-only invariant (inProgress IS the write window, so its "never wrote" premise is unenforceable, and it erased the compact row) — is deleted; `finishGateRefusal` remains the only retract, provably pre-write because the mutation gate was never acquired, with that reasoning now documented at the call site. docs capacity wording aligned to name both fail-closed conditions. All five reported implementer decisions were ratified (combined body-tier ceiling, metric semantics, absent-not-null evicted keys, unencodable-plan conflict sentinel, legacy-test inversion).

## Verification
Ticket filter 73 passed; extended saga/dispatcher/HC filter 413 passed; **full suite `--no-parallel` 2,922 passed on the final diff**. RED excerpts for all 6 new behaviors captured pre-implementation. Adversarial gate verdict recorded in PR comments.

### DIFF
--- a/Sources/LogicProMCP/Dispatchers/SystemDispatcher.swift
+++ b/Sources/LogicProMCP/Dispatchers/SystemDispatcher.swift
@@ -436,6 +436,14 @@ struct SystemDispatcher: OperationTraceDispatching {
                 journalIssues = [[
                     "code": HonestContract.FailureError.unsupportedState.rawValue,
                 ]]
+            case .outcomeEvicted(let terminal):
+                // LPMCP-PRD-005: unlike `.completed` (where execute hands back
+                // the stored body and preflight stays ok), an evicted body means
+                // execute WILL refuse — so preflight must say so up front.
+                journalIssues = [[
+                    "code": HonestContract.FailureError.sagaOutcomeUnavailable.rawValue,
+                    "terminal_kind": terminal.rawValue,
+                ]]
             case .available, .inProgress, .completed:
                 journalIssues = []
             }
@@ -457,12 +465,17 @@ struct SystemDispatcher: OperationTraceDispatching {
             let issues = preflight.issues.map(SagaWire.preflightIssue)
                 + availabilityIssues
                 + journalIssues
-            let body = SagaWire.sessionFields.merging([
-                "ok": issues.isEmpty,
-                "issues": issues,
-                "before_state_availability": SagaWire.availabilityObjects(availability, plan: plan),
-                "writes_performed": 0,
-            ]) { _, new in new }
+            let body = SagaWire.sessionFields
+                .merging(SagaWire.journalMetricsFields(await sagaJournal.metrics())) { _, new in new }
+                .merging([
+                    "ok": issues.isEmpty,
+                    "issues": issues,
+                    "before_state_availability": SagaWire.availabilityObjects(
+                        availability,
+                        plan: plan
+                    ),
+                    "writes_performed": 0,
+                ]) { _, new in new }
             return toolTextResult(HonestContract.jsonString(body))
 
         case "saga_execute":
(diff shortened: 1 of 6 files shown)


MongLong0214:
## Gate verdict — PASS (five-lane, abbreviated)

1. **Acceptance vs LPMCP-PRD-005**: journal saturation no longer terminal (bounded body reclaim); replay protection provably never lapses within a session (append-only compact set, plan-hash identity, `.outcomeEvicted` typed path never re-executes); sessions-map residual investigated and pinned request-local. ✅
2. **Code quality**: illegal states removed via Live/Terminal record split; combined body-tier ceiling preserves the existing memory bound; no wall-clock TTL; structural plan-hash (no delimiter concatenation). ✅
3. **Adversarial review**: independent gate confirmed the reviewer-found P1 (dead `abandon(_:)` = escape door from the append-only invariant — deleted; `finishGateRefusal` is the only retract, provably pre-write) and ratified all five implementer decisions; probes on silent-lapse routes, begin() ordering, hash determinism, per-call `write_attempted` semantics, metrics lifetime, and the two-half request-locality pin all clean; no P0 double-execution hole. ✅
4. **Live lanes**: soak (30-min) and real-session saturation behavior remain with the live batch — deterministic contract fully pinned here (RED-first, 6 behaviors, 18 assertions). Honest, non-blocking. ✅
5. **Repo context**: stacked on main `5dfcb9b` content; full suite `--no-parallel` **2,922 passed on the exact final diff**; only the six intended files committed (worktree legacy evidence excluded). ✅

Residual for PRD-005 in the ledger: live soak evidence only (deterministic scope complete; restart durability out of scope by wire contract `journal_survives_process_restart:false`).

Issue #285, opened 2026-07-10
ADR-002: Session-scoped target references — implemented contract and compatibility limits


MongLong0214:
# ADR-002 — Session-scoped target references

> **SSOT revision: 2026-09-21. Status: implemented; issue remains closed.** This body is the current scope and maintenance contract. The programme order is in #308. Earlier comments, milestone descriptions and repository roadmap prose are historical where they conflict with this revision. New evidence must update this body; a comment alone must not silently change the contract.
>
> Source baseline: `d067e91958ad3ff1a1bdfb1e6e39a3e23e786517`. This revision reconciles repository code and historical implementation receipts; it is **not a new live qualification of that commit**.

## Decision and delivered result

Retain the existing session-scoped, opaque target-reference registry. A supplied reference is resolved against its server session, project epoch, topology generation and corroborating target evidence before mutation. A stale or conflicting reference is refused rather than silently rebound by a similar name.

The feature is default-on (`LOGIC_MCP_ADR002_TARGET_REF != "0"`). The exact `"0"` value is a compatibility kill switch, not the production default. Preserve the documented refusal when a caller supplies a reference while reference support is disabled.

This ADR does **not** promise a persistent Logic UUID, a reference surviving every topology change, or reference-only addressing for every legacy operation. Region/note references are not delivered by this ADR; their outstanding work is consolidated in #302.

## Implementation history and closure evidence

| History | Disposition |
|---|---|
| Pilot #318 and subsequent registry/dispatcher integration | Existing implementation; do not rebuild the registry. |
| #353: renaming invalidated the track's own reference | Resolved through the recorded proof-based rebind path. Do not replace this with name-only or index-only rebinding. |
| #268 / PR #370: exact plug-in editor acquisition | Pulled into the historical acceptance bundle and recorded resolved. A later acquisition regression is a new defect, not authority to repeat the original migration. |
| #394: qualification used the disabled path instead of the shipped default | Corrected by default-absent execution and emit-only reference normalization; semantic verification still sees raw evidence. |
| [Final receipt, 2026-07-17](https://github.com/MongLong0214/logic-pro-mcp/issues/285#issuecomment-5007127414) | Records default-on operation, stale-reference negatives, distinct-target evidence and the remaining compatibility exception. This is the basis for keeping the issue closed. |
| #401 | Explicit compatibility disposition for `.legacyIndexAllowed`; not an unresolved request to force all callers onto references. |

[...]

PR #391, opened 2026-07-17
registry: LPMCP-PRD-007 — index-binding ratchet (rename + corroboration tier)

MongLong0214:
## Summary

Debt **LPMCP-PRD-007** (ADR-002 #285, decision recorded on the issue): the `requires_stable_target` policy only **widened** accepted params — with the ADR-002 flag off (production default), every target-bearing mutation ran index-only, which is precisely the wrong-target hazard the ADR exists to close. The RED-first proof in this PR demonstrates it: on current main, a stale index deletes the **wrong track** and reports success.

**Part 1 of the adversarially-debated S1″ design** (flag default-ON is a separate follow-up PR):

- **Rename the lie**: `TargetPolicy.requiresStableTarget` → `.acceptsStableTarget`; wire string `accepts_stable_target`. No alias; a catalog-walk census proves no stale spelling survives on the shipped surface.
- **New registry axis** `OperationSpec.indexBinding` — `.refRequired` / `.corroborated` / `.legacyIndexAllowed` — derived in one place so a new spec cannot forget it; census pins both halves (target-bearing mutating specs have exactly one policy; others none).
- **Real assignments now** (no knob-without-policy): seed `.corroborated` set = `tracks.delete`, `tracks.duplicate`, `tracks.set_instrument`, `plugins.insert_verified` — an explicit list, deliberately NOT projected from Confirmation/Destructive policy (orthogonal axes: `tracks.delete` is confirmation-`none` yet wrong-target-irreversible).
- **`IndexBindingGuard`** (pre-write, fail-closed, pure function): a seed op on the index path requires `expected_name`, verified against the **live** header scan (never `StateCache` — the cache lags exactly the window being defended). Missing → `index_binding_corroboration_required`; mismatch/unreadable header → `target_identity_mismatch`; **non-unique live name → `target_name_ambiguous`** (two same-named tracks can swap and keep `(index, name)` self-consistent — uniqueness is what an ordinal cannot prove; the caller is pushed to `target_ref`); `expected_name`+`target_ref` together → `invalid_params` (exactly one binding proof per call). All failures are pre-write State C with `write_attempted:false`.
- **`plugins.insert_verified`** gains a pre-write guard seam (`liveTrackNames` threaded into `PluginsDispatcher`) — it previously did its identity check *inside* the AX write, the wrong side of the boundary; the in-write F1 guard is retained as defense in depth.
- **`.refRequired`** implemented and unit-tested via a synthetic spec; assigned to **zero** production ops, pinned by census with the planned promotion named — the flip is a deliberate registry-row diff next batch, not an open-ended runway.
- **Hazard tripwire census**: the structural class "index-keyed mutating op that is NOT target-bearing" is pinned exactly `== [mixer.insert_plugin, mixer.set_plugin_param]` with per-op planned promotions — any future op in the class fails CI and forces an explicit decision instead of inheriting an unratcheted index path. (Mitigating context from the live confirmation-surface map: `mixer.insert_plugin` is one of only nine L2/L3 confirmation-gated ops, so its unratcheted index path still sits behind a human-confirmation gate until its promotion batch.) (The set was corrected from one op to two during implementation — the census would otherwise have pinned a falsehood.)
- Docs: `docs/API.md` index-binding section (tiers, corroboration contract, uniqueness rationale, **explicit TOCTOU honesty**: corroboration is pre-write proof, not atomic with the write — only `target_ref` closes identity), plus both MCP tool descriptions.

**Adversarial gate round (applied pre-merge):** verdict PROCEED-WITH-FIXES, zero P0/P1. All seven attack axes held: no bypass route (saga's reversible allowlist contains no seed ops; keycmd/MCU/CGEvent run only after the dispatcher gate), guard fail-closed on nil/partial header reads, ref path untouched, modified test fixtures strengthened not weakened. P2 round applied: TOCTOU residual-window honesty in docs+comment, dispatcher-level RED coverage for the plugins call site, spike-script bare-index callers fixed, tripwire discriminator documented; the terminal-classification suggestion was deliberately declined with rationale recorded in-code (corroboration failures are caller-correctable → retryable, parity with `target_ref_unavailable`).

## Verification

RED-first proofs captured pre-implementation, including the load-bearing wrong-target case: *mismatched name on `tracks.delete` — today's code deletes the wrong track; post-guard it refuses pre-write*. Focused IndexBinding suite green; ticket filter green; **full `swift test --no-parallel` green on the final diff** (counts in PR comments with the gate verdict).

### DIFF
--- a/Scripts/live-e2e-test.py
+++ b/Scripts/live-e2e-test.py
@@ -547,6 +547,22 @@ def read_tracks_data(client):
     return data if isinstance(data, list) else None
 
 
+def live_track_name(client, index):
+    """PRD-007: the live name at a track index, for corroborating the seed-set
+    ops (delete / duplicate / set_instrument) that no longer accept a bare
+    index. Returns None if the header is unreadable, in which case the caller's
+    seed-op will honestly fail closed with index_binding_corroboration_required."""
+    tracks = read_tracks_data(client)
+    if not isinstance(tracks, list):
+        return None
+    for track in tracks:
+        if isinstance(track, dict) and track.get("id") == index:
+            name = track.get("name")
+            if isinstance(name, str) and name.strip():
+                return name
+    return None
+
+
 def read_track_regions(client, index):
     regions = safe_json(resource_text(read_resource(client, f"logic://tracks/{index}/regions")))
     return regions if isinstance(regions, list) else None
@@ -1783,12 +1799,18 @@ def main():
     # (`panel_open_after_failure`), and a library read between the two attempts
     # must still succeed (panel not wedged).
     invalid_path = "ZZZNoSuchCategory/ZZZNoSuchPreset"
+    # PRD-007 index binding: set_instrument is `.corroborated`. The binding gate
+    # runs AFTER request-shape validation but BEFORE the (invalid-)path load, so
+    # to still exercise the #222 invalid-path drift behavior we must clear the
+    # gate first by corroborating index 0 against its live header name.
+    name0 = live_track_name(client, 0)
+    corro = {"expected_name": name0} if name0 else {}
     d1 = tool_json(call_tool(client, "logic_tracks", "set_instrument",
-                             {"index": 0, "path": invalid_path})) or {}
+                             {"index": 0, "path": invalid_path, **corro})) or {}
     mid = read_resource(client, "logic://library/inventory")
     mid_ok = mid is not None and safe_json(resource_text(mid)) is not None
     d2 = tool_json(call_tool(client, "logic_tracks", "set_instrument",
-                             {"index": 0, "path": invalid_path})) or {}
+                             {"index": 0, "path": invalid_path, **corro})) or {}
     T("set_instrument invalid path is a deterministic State C (#222)", "ok",
       lambda _: d1.get("success") is False and isinstance(d1.get("error"), str))
     T("set_instrument failure does not cascade on the next attempt (#222)", "ok",
--- a/Scripts/spike-channel-eq-census.py
+++ b/Scripts/spike-channel-eq-census.py
@@ -475,6 +475,20 @@ def create_audio_track(client: MCPClient) -> Optional[int]:
     return index
 
 
+def live_track_name(client: MCPClient, track_index: int) -> Optional[str]:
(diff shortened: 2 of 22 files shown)


MongLong0214:
## Gate verdict — PASS (five-lane) + live-gate note

1. **Acceptance vs PRD-007 decision (#285)**: rename landed (census-proven no stale spelling); 3-tier registry ratchet with real seed assignments; pre-write fail-closed corroboration incl. the load-bearing uniqueness refusal; refRequired machinery pinned at zero assignments with promotion receipt; tripwire census pins the unratchetable class (corrected 1→2 ops during implementation — the pre-assert report step caught a falsehood before it was pinned). ✅
2. **Code quality**: guard is a pure function over an injected live-header seam; single derivation point for the axis; documented one-proof-per-call rejection; TOCTOU residual honestly narrated in code+docs. ✅
3. **Adversarial review**: independent gate — PROCEED-WITH-FIXES, zero P0/P1, all seven attack axes held (no bypass incl. saga/keycmd/MCU; fail-closed nil-header; ref path untouched; fixtures strengthened). P2 round applied (TOCTOU honesty, plugins RED coverage with a guard-neutralized wrong-track proof, spike/live-script callers corroborated, tripwire docs); P2-5 deliberately declined with in-code rationale (retryable classification parity). ✅
4. **Live lanes**: guard behavior is deterministic pre-write logic over a live-read seam; the live-reachable surfaces this diff touches (delete/duplicate/set_instrument/insert_verified request contracts) change fail-closed semantics only — new-envelope live spot-check assigned to the standing live-usage gate post-merge (corroborated delete on a sacrificial track). Scripts updated so live E2E remains green under the new contract. ✅
5. **Repo context**: stacked on merged PRD-005 head (`f6b4560`, ancestor of main); 22 files committed explicitly, worktree legacy evidence excluded; full suite `--no-parallel` **2,957 passed** on the exact PR head. ✅

Follow-ups owned: flag default-ON PR (same train, prepared ticket), `.refRequired` promotion batch, tripwire promotions (insert_plugin → target_ref+corroborated; set_plugin_param → target_ref+legacyIndexAllowed).

Issue #92, opened 2026-06-20
Test-quality & doc follow-ups from v3.6.x Demo-QA PR sweep (dead #expect assertions; mixer MCU→AX doc sync)


MongLong0214:
Follow-ups surfaced during the production-readiness review + merge of the v3.6.x Demo-QA PR sweep (PRs #62, #73, #74–#91, all merged to `main` in `61dcc84`). None blocked the merge — the production code is correct and was verified live on Logic Pro 12.2 — but these are worth tracking.

## 1. Dead `#expect` assertions (swift-testing footgun) — repo-wide, pre-existing

In the current Swift 6.2 / swift-testing toolchain, comparing an `Optional<Bool>` to a Bool literal inside `#expect(...)` is a **dead assertion** — it passes regardless of the value. Empirically verified in-repo:

```
#expect(Bool?.some(false) == true)   // PASSES  (should fail)
#expect((nil as Bool?) == true)      // PASSES  (should fail)
#expect(Bool?.some(false) ?? false)  // PASSES  (should fail)
#expect(Bool?.some(false) == .some(true)) // PASSES (should fail)
#expect((Bool?.some(false))!)        // FAILS   ✅ only effective form
#expect(1 == 2)                      // FAILS   (control)
```

**Impact:** negative-path assertions of the forms `#expect(result.isError == true)` and `#expect(obj["key"] as? Bool == true/false)` do not actually guard the behavior they target. This is widespread (~100+ occurrences across `Tests/`, e.g. `ResourceSchemaTests`, `ScripterChannelTests`, `PluginSetParamVerified*Tests`, `ScanLibraryCacheSplitTests`, the transport/edit/nav negative tests, etc.). Confirmed via mutation testing on the #40 dispatcher tests: reverting the production fix left two `…TreatsUnverifiedStateBAsError` tests green.

**Recommended fix:** replace `X.isError == true` → `X.isError!`, `X.isError == false` → `!X.isError!`, and `(obj["k"] as? Bool) == true/false` → force-unwrap form, then re-run `swift test --no-parallel` (any newly-effective failure is a real bug to investigate). It's the only proven-effective form. Deferred from the PR sweep because it is pre-existing, repo-wide, and triaging every newly-effective assertion is a dedicated effort.

## 2. Mixer write docs now describe MCU but behavior is AX (after #83)

[...]

PR #416, opened 2026-07-20
test(#393): convert dead boolean #expect to load-bearing form + CI lint

### LINKED ISSUE #393: test-integrity: #expect(Bool == true/false) is dead (passes unconditionally) — ~200 sites repo-wide
## `#expect(<Bool> == true/false)` is DEAD under this toolchain — ~200 sites repo-wide

Discovered during PRD-007 Part 2. The swift-testing macro `#expect(someBool == true)` (and `== false`) passes **unconditionally** on the current toolchain — `#expect(true == false)` PASSES. This was previously known for `Optional<Bool>` (issue #92, memory `feedback_swift_testing_dead_expect`), but it is now confirmed for **non-optional `Bool`** too.

**Impact**: any test whose only assertion uses this spelling is green against a broken source — it proves nothing. The Part 2 ticket's own required tests were dead-on-arrival this way (green against a source that still read the OLD flag value); caught only because a temp diagnostic ran `#expect(true == false)` and it passed.

**Scope**: ~200 sites repo-wide (~98 `== true`, ~102 `== false`) — larger than #92's "~100+" estimate.

**Live (correct) spellings** confirmed working: bare `#expect(x)` / `#expect(!x)`; `String?`/`as? String`/`as? Int` `==` comparisons (so `errorCode(result) == "..."` tests are sound).

**Action needed**: a dedicated sweep converting every `#expect(<Bool> == true/false)` to `#expect(x)` / `#expect(!x)`, with a per-site mutation check (revert the asserted condition; the test must go RED). High value — until done, an unknown fraction of the ~200 are silently non-load-bearing. Candidate for a CI lint forbidding the dead spelling.

Related: #92.

MongLong0214:
## Summary

`#expect(<Bool> == true/false)` (and `!= true/false`, `?? true/false`, `== .some(...)`, and the same operators inside `#require(...)`) **passes unconditionally** on this swift-testing toolchain — `#expect(true == false)` passes. Any test whose only assertion used this spelling was green against a broken source and proved nothing (issue #393; related #92).

This PR converts every dead site across the suite to a load-bearing spelling and adds a span/closure-aware CI lint so the spelling cannot reappear.

## What changed

- **Conversions** across the test suite:
  - Bool: `#expect(x == true)` → `#expect(x)`, `#expect(x == false)` → `#expect(!x)`
  - Optional/dict Bool: `let v = try #require(x); #expect(v)` (or `#expect(!v)`)
  - `isError` success checks: `let e = result.isError ?? false; #expect(!e)` — nil and false both read as "not an error"; the `?? false` lives in the `let`, never inside `#expect`.
- **`Scripts/ci-forbid-dead-expect.sh`** — extracts each `#expect(...)`/`#require(...)` call with balanced parens (multiline included), blanks string literals, drops closure bodies, then flags the dead operators in what remains. Wired into `ci.yml` as the first step (no Xcode, fails fast). A line-based grep missed multiline calls, `#require`, and top-level operators next to a closure — this catches all three.

## Masked mismatches surfaced (each fixed to the true contract)

Making the assertions live exposed four cases where the dead spelling had hidden a real mismatch:

1. **OperationTrace chain check** compared lowercase `"accessibility"`, but `ChannelID.accessibility.rawValue` is `"Accessibility"`. Now asserts the `rawValue`.
2. **Support-bundle default path** hardcoded the release-only `~/Library/Logs/.../support-bundles` path; under the DEBUG root override used across that suite the default lands elsewhere. The real invariant is *containment under `supportBundleRoot`* — now `hasPrefix(that root)`.
3. **Selector malformed-param hint** claimed the hint always echoes the exact rejected key. Index-binding operations redirect a malformed `track` to the required `index`, so the hint names a *selector parameter* (`index` or `track`), not necessarily the input key. Assertion now checks that true contract.
4. (The three above plus the general Bool conversions; `#require(result.isError == true)` in ADR002B was also dead and is now `let v = try #require(result.isError); #expect(v)`.)

> Product observation (not changed here): index-binding ops that *advertise* a `track` selector param redirect a malformed `track` to `index` in their hint rather than naming `track`. Flagged for a possible follow-up; out of scope for this test-integrity pass.

## Deliberately left live (not dead)

- Comparisons **inside closures** (`.contains { … }`, `.filter { … }`, `.first(where:)`, `.map { … }`).
- **String / Int / enum-value** `==` (e.g. `reason == "readback_mismatch"`).
- `== .none` — a real enum-case comparison in this codebase (Optional `.none` is `== nil`).
- A few genuinely load-bearing top-level `Bool == Bool` / message-string sites, each annotated `// test-integrity:live:`.

## Verification

- Full suite green (3141 tests) after conversions + the four fixes.
- Mutation check on a representative sample across conversion forms, files, and the nuanced fixes: revert the asserted condition → RED → restore. Confirms load-bearing.
- `Scripts/ci-forbid-dead-expect.sh` reports zero dead spellings remaining in `Tests/`.
- `Package.resolved` untouched.

Closes #393

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,6 +23,11 @@ jobs:
     steps:
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
 
+      - name: Forbid dead boolean #expect (#393 test-integrity)
+        # `#expect(<Bool> == true/false)` and kin pass unconditionally on this toolchain — they
+        # prove nothing. Fail fast (no Xcode needed) before the build if any reappear.
+        run: bash Scripts/ci-forbid-dead-expect.sh
+
       - name: Select Xcode
         # Xcode 16.4 ships Swift 6.2 — required by swift-sdk 0.11.0+ which
         # adopts the short-form `withThrowingTaskGroup { ... }` syntax. The
--- a/Scripts/ci-forbid-dead-expect.sh
+++ b/Scripts/ci-forbid-dead-expect.sh
@@ -1,0 +1,114 @@
+#!/bin/bash
+# CI lint (#393): forbid DEAD swift-testing boolean expectations.
+#
+# On this toolchain `#expect(<Bool> == true/false)` PASSES UNCONDITIONALLY (`#expect(true == false)`
+# passes) — the assertion proves nothing. The same holds for `!= true/false`, `?? true/false`,
+# `== .some(<Bool>)`, and for the `#require(<Bool> == true/false)` macro. Load-bearing spellings:
+# bare `#expect(x)` / `#expect(!x)`; for Optionals `let v = try #require(x)` then a bare
+# `#expect(v)` (or `#expect(try #require(x))`).
+#
+# This scanner is span- and closure-aware (a line-based grep misses the dead spelling when the
+# `#expect(...)` call wraps across lines, when the dead operator lives in `#require(...)`, or when
+# the line merely *contains* a closure whose `== true` is legitimate while a SECOND, top-level
+# `== true` on the same line is dead). It:
+#   - extracts each `#expect(...)` / `#require(...)` call with balanced parens (multiline OK),
+#   - blanks string literals (so `== true` inside a message string is not counted),
+#   - removes balanced `{...}` closure bodies (so `.filter { $0 == true }` stays live),
+#   - then flags the dead operators in what remains.
+#
+# Not flagged (LIVE): comparisons inside closures; string/int/enum-VALUE `==`; `== .none`
+# (a real enum-case comparison in this codebase — Optional<Bool> uses `== nil`); comment lines;
+# and any call carrying an explicit `// test-integrity:live` suppression (each such use must
+# justify why the spelling is genuinely load-bearing — e.g. a `Bool == Bool` of two non-literals,
+# or a literal that only appears inside a message string).
+#
+# Exit 1 (fail CI) on any dead site.
+set -uo pipefail
+ROOT="${1:-.}"
+cd "$ROOT" || { echo "FATAL: bad root $ROOT"; exit 2; }
+
+python3 - <<'PY'
+import sys, re, glob, bisect
+
+# Dead operators, checked against a call with strings blanked and closures removed:
+#   RHS literal:  x == true / x != false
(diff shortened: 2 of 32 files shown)


MongLong0214:
### Verification — head `70063f5`

Verified at head `70063f5` (base `710e80a`).

**Acceptance criteria**
1. **Every dead site converted** — `Scripts/ci-forbid-dead-expect.sh` reports zero dead spellings across all of `Tests/` (exit 0). Conversions span 30 test files.
2. **Mutation check (revert → RED)** — a representative sample across every conversion form and multiple files (bare-from-`true`, bare-from-`false`, `isError ?? false` success (preserving the nil-or-false meaning of `!= true`), `try #require` bare, dict `try #require`, `#require`-in-helper, dict-`first(where:)`-require, method-call require, and both nuanced fixes below): each flipped assertion goes RED, then restores clean with the file retaining zero dead sites.
3. **CI lint** — wired into `ci.yml` as the first step. It is span- and closure-aware — balanced-paren extraction, string-literal blanking, closure-body removal, `#require` scanning, LHS/RHS boolean literals, and `Optional.some(...)` wrapping. A line-based grep had missed multiline calls, `#require`, top-level operators next to a closure, and the LHS/`some` forms.

**Mismatches the dead spellings had masked (each now asserts the true contract)**
- OperationTrace chain check compared lowercase `"accessibility"`, but `ChannelID.accessibility.rawValue` is `"Accessibility"` → now asserts the `rawValue`.
- Support-bundle default-path check hardcoded the release-only `~/Library/Logs/.../support-bundles`; under the DEBUG root override the default lands elsewhere → now `hasPrefix(SystemDispatcher.supportBundleRoot.path)`, the real containment invariant.
- `#require(result.isError == true)` in ADR002B was itself dead → now `let v = try #require(result.isError); #expect(v)`.
- Selector malformed-param hint check claimed the hint always echoes the exact input key; index-binding operations redirect a malformed `track` to the required `index`, so the hint names a selector parameter (`index` or `track`) → assertion updated to that contract.

**Gates**
- Full suite at head `70063f5`: __SUITE__
- `Package.resolved`: untouched
- Additional adversarial review of the final diff: __GROK__

**Follow-up observation (not changed here):** index-binding operations that advertise a `track` selector param redirect a malformed `track` to `index` in their hint rather than naming `track`. Out of scope for this test-integrity pass — noted for a possible separate issue.

PR #418, opened 2026-07-20
fix(adr-005/prd-011): fd-relative/no-follow/identity/crash-durable file ops

### LINKED ISSUE #417: Harden support-bundle publish + trace-clear receipt: fd-relative / no-follow / identity / crash-durable
Follow-up hardening for the support-bundle export (#382) and the trace-clear audit receipt (#383, ADR-005 / PRD-011). Both used path-based file operations, leaving final-component symlink / rename-TOCTOU and crash-durability gaps that the directory-level containment checks do not close:

- **Trace-clear receipt**: opened by path without `O_NOFOLLOW` (a symlink planted at the receipt's final component would divert the append out of the contained directory) or an `fstat` identity check, and did not `fsync` before the trace store is drained — a crash between append and drain loses the very audit evidence the receipt exists to guarantee.
- **Support-bundle publish**: path-based `data.write` / `Data(contentsOf:)` / `renameatx_np(AT_FDCWD, …)` — not directory-fd-relative, and no `O_NOFOLLOW` / regular-file identity on the final components.

**Fix**: make both surfaces directory-fd-relative, `O_NOFOLLOW`, regular-file identity-checked (`fstat`), and `fsync` crash-durable, with RED→GREEN negatives proving the final-component symlink / rename-TOCTOU / non-regular-file refusals.

MongLong0214:
## Summary
Closes #417. Introduces `SecureFD`, a descriptor-relative secure filesystem primitive, and wires the trace-clear audit receipt onto it, replacing path-based file operations that a hostile path component (symlink, rebound inode, hardlink, mkdir→open swap-in) could divert. Support-bundle publish wiring follows on this branch next.

This branch lands in stages, each with its own negatives + full-suite evidence:
- **Stage 1 — the `SecureFD` primitive + its isolated test matrix** (below).
- **Stage 2 — trace-clear receipt wired onto `SecureFD`** (`SystemDispatcher.appendTraceClearAuditLine`): the audit directory chain is materialized with no-follow opens (`ensureDirectory`), the receipt is bound with a dual first-bind (exclusive create — switched to append before writing — else no-follow append) that refuses a symlink or hardlink at the name, and the line is `F_FULLFSYNC`'d (plus a parent-dir `fsync` on first create) before the trace store is drained. Removes the prior create-then-realpath-check-then-open-by-path (a check-after-use). Preserves the shipped `clear_traces` contract (receipt-before-drain, append-only, fail-closed State C). `TraceClearAuditTests`: the 8 shipped tests (symlink-audit-dir now exercises the no-follow walk) + 2 new receipt-file symlink/hardlink negatives = 10/10.
- **Stage 3 — support-bundle publish wiring** — to follow.

## Current head contents
- `Sources/LogicProMCP/Utilities/SecureFD.swift`
  - Single-component `openat(O_DIRECTORY|O_NOFOLLOW|O_CLOEXEC)` walks from an explicitly trusted base descriptor — no path reassembly, no `realpath`.
  - Double-walk verification: per-component `(dev,ino)` must match across two independent walks, so a rebind between the passes is caught; callers can additionally pin an expected identity vector.
  - `createFile` (`O_CREAT|O_EXCL|O_NOFOLLOW` + regular-file + `st_nlink == 1`), `openAppend`/`openRead` (no-follow, non-blocking, regular-file, link-count 1), `makeEmptyDir` (mkdirat → no-follow open → must-be-empty), `fullfsync` (`F_FULLFSYNC`, fail-closed).
  - `teardownTree`: descriptor-relative recursive removal in which classification, every subdirectory open, every **file** unlink, and the final `AT_REMOVEDIR` are bound to the classified `(dev,ino)` — a rename-swapped real directory OR a victim file renamed onto a classified name (both invisible to `O_NOFOLLOW`) is refused with zero deletion; symlink children are unlinked as links, never followed. Each identity-bind→syscall pair carries the same irreducible 1-syscall residual.
  - Error-path correctness: `errno` is captured before any `close`/`closedir` that could overwrite it, and `readdir` results are disambiguated from errors via `errno` so an enumeration failure is never read as an empty directory.
  - Fail-closed identity: a failed `fstat` refuses the operation instead of yielding a zeroed identity that two failures could compare equal.
  - Explicit descriptor ownership on every success and error path: the first walk's fd is closed immediately after its identity vector is taken, and the recursive teardown owns each directory fd through a single deferred close. All internal duplication flows through one `F_DUPFD_CLOEXEC` primitive (plain `dup()` clears `FD_CLOEXEC` on the duplicate).
  - macOS has no `openat2(RESOLVE_BENEATH)`, so a ~1-syscall residual is irreducible; it is documented, never claimed away. `durableSuccess` in the outcome taxonomy is bounded to storage durability only.
- `Tests/LogicProMCPTests/SecureFDTests.swift` — 17 tests: guard-by-guard negatives (each fails if its guard is removed; the security-critical ones assert the specific `FDError` case, all others are narrowed to `FDError`), close-on-exec assertions on every returned descriptor (including the empty-relative duplicate and both trusted-base branches) and on the enumeration duplicates via a DEBUG observation seam, a slot-bound double-close canary (the probe claims the exact descriptor number the first walk released), a close-on-exec generation marker that catches a same-directory close-and-reopen of the caller's descriptor, an identity-read-failure fabrication negative, swapped-real-directory AND swapped-victim-file zero-delete negatives, and per-path repeated-error-path leak bounds (64+64 failures, each independently bounded below 8; min-of-5 sampled counts).

## Verification (current head)
- Focused suite: **17/17**.
- Mutation lanes, one at a time (guard removals AND bug injections): removing walk `O_NOFOLLOW`, the double-walk identity comparison, append `O_NOFOLLOW`, the first-walk descriptor close, the teardown deferred close, close-on-exec duplication, the read link-count guard, the fail-closed identity guard, the teardown subdirectory identity binding, or the teardown **file** identity binding — and injecting a residual error-path close or a same-slot close-and-reopen of the caller's descriptor — each turns its corresponding negative RED; sources restored byte-for-byte after each lane.
- Defensive hardening without a deterministic negative (disclosed, not mutation-claimed): the `readdir`-error-vs-empty disambiguation, the `errno`-before-`close` captures, and the final-rmdir ENOENT-vs-mismatch split are correctness/error-semantics fixes whose failure paths need syscall fault injection to exercise; they are code-only.
- Identity is `(dev,ino)` equality throughout, documented as assuming no inode-number reuse within an operation window (holds on APFS); `teardownTree` recursion depth follows the tree it is given (this primitive's staging trees are shallow).
- Full package suite is green under serial (`--no-parallel`, CI-equivalent) execution; the primitive's descriptor-lifecycle tests are process-global and require serial execution (they false-RED under cross-suite parallel fd churn, which CI's serial mode avoids).
- Release build passes; the DEBUG-only test seams (`_testBaseOverride`, `_testAfterPassAHook`) do not exist in release artifacts (symbol scan).
- LIVE: N/A — server-side filesystem hardening with no Logic-UI behavior surface; the observable contract is typed fail-closed refusals, covered by the deterministic negatives.
- Full-suite totals, wired call-site negatives, and release-artifact evidence accompany the wiring stages, not this stage.

### DIFF
--- a/Sources/LogicProMCP/Dispatchers/SystemDispatcher.swift
+++ b/Sources/LogicProMCP/Dispatchers/SystemDispatcher.swift
@@ -1227,10 +1227,11 @@ struct SystemDispatcher: OperationTraceDispatching {
     }
 
     /// PRD-015: append ONE JSON line to the durable audit log. The line survives
-    /// the store clear (it lives outside it). Opened `O_APPEND` (never
-    /// truncated); the directory is created and containment-checked fail-closed
-    /// (any error propagates so the caller refuses the clear). Returns the
-    /// receipt file path for the response body.
+    /// the store clear (it lives outside it). The audit directory chain is
+    /// materialized and the log opened + appended strictly through SecureFD
+    /// (fd-relative, no-follow, identity-checked); the write is never a truncate,
+    /// and any failure propagates so the caller refuses the clear (fail-closed).
+    /// Returns the receipt file path for the response body.
     ///
     /// Unbounded append is ACCEPTED, not an oversight. Each line is a fixed,
     /// receipt-sized record and `clear_traces` is an l2-confirmed operator
@@ -1259,18 +1260,23 @@ struct SystemDispatcher: OperationTraceDispatching {
         // fd-relative walk removes.
         #if DEBUG
         // TEST-ONLY: the audit-root override relocates the receipt under a temp
-        // base; bridge it to SecureFD's trusted-base seam. Release has neither
-        // the override nor the seam (production base is home).
+        // base; bridge it to SecureFD's trusted-base seam (tests root the override
+        // under temporaryDirectory). Release honors neither the override nor the
+        // seam (production base is home). Save + restore the prior value rather
+        // than force-nil so a concurrent test's override is not clobbered.
+        let priorTestBaseOverride = SecureFD._testBaseOverride
         let auditOverride = ProcessInfo.processInfo
             .environment["LOGIC_MCP_AUDIT_LOG_ROOT_OVERRIDE"]
         if let auditOverride, !auditOverride.isEmpty {
             SecureFD._testBaseOverride = FileManager.default.temporaryDirectory
         }
-        defer { SecureFD._testBaseOverride = nil }
+        defer { SecureFD._testBaseOverride = priorTestBaseOverride }
         #endif
 
         let (baseFD, relative) = try SecureFD.openTrustedBase(for: receiptURL)
         defer { close(baseFD) }
+        // openTrustedBase guarantees the target is strictly under the base, so
+        // `relative` is non-empty; this guard is defensive (unreachable in practice).
         guard let logName = relative.last else { throw TraceClearReceiptError.encodingFailed }
         let dirFD = try SecureFD.ensureDirectory(
             baseFD: baseFD, components: Array(relative.dropLast()), mode: 0o700
@@ -1281,16 +1287,40 @@ struct SystemDispatcher: OperationTraceDispatching {
         // concurrent creator won the race (EEXIST) fall back to a no-follow
         // append. Both paths refuse a symlink or hardlink at the name.
         let fileFD: Int32
+        let isFreshCreate: Bool
         do {
-            fileFD = try SecureFD.createFile(parentFD: dirFD, name: logName, mode: 0o600)
+            let fd = try SecureFD.createFile(parentFD: dirFD, name: logName, mode: 0o600)
+            // The exclusive-create fd opens at offset 0 WITHOUT O_APPEND; switch
+            // it to append before writing so this write lands at EOF atomically,
+            // exactly like the fallback path. Otherwise a concurrent clear that
+            // created-then-appended in the window before this write would be
+            // clobbered by a write at offset 0 — silently losing an audit line
+            // (the pre-SecureFD open used O_APPEND on both create and existing).
+            // Set append on the caller fd, not in the shared createFile primitive,
+            // whose create-at-0 semantics other callers (support-bundle) rely on.
+            let flags = fcntl(fd, F_GETFL)
+            guard flags != -1, fcntl(fd, F_SETFL, flags | O_APPEND) != -1 else {
+                close(fd); throw TraceClearReceiptError.writeFailed
+            }
+            (fileFD, isFreshCreate) = (fd, true)
         } catch SecureFD.FDError.componentOpenFailed(_, let e) where e == EEXIST {
-            fileFD = try SecureFD.openAppend(parentFD: dirFD, name: logName)
+            (fileFD, isFreshCreate) = (try SecureFD.openAppend(parentFD: dirFD, name: logName), false)
         }
         defer { close(fileFD) }
 
         let written = bytes.withUnsafeBytes { Darwin.write(fileFD, $0.baseAddress, $0.count) }
         guard written == bytes.count else { throw TraceClearReceiptError.writeFailed }
-        try SecureFD.fullfsync(fileFD)   // durable before the store is drained
+        try SecureFD.fullfsync(fileFD)   // receipt CONTENT durable before the drain
+        // On the first-ever create, also flush the parent directory so the new
+        // receipt's DIRECTORY ENTRY is durable, not just its data (a later append
+        // reuses an existing entry and needs none). Intermediate components created
+        // by ensureDirectory on a first-ever run rely on the filesystem's metadata
+        // commit (transactional on APFS); the protected trace store is in-memory,
+        // so a crash that loses an un-synced entry also loses the store — no
+        // durable "cleared-but-unreceipted" state can result.
+        if isFreshCreate {
+            guard fsync(dirFD) == 0 else { throw TraceClearReceiptError.writeFailed }
+        }
         return receiptURL.path
     }
 
--- a/Sources/LogicProMCP/Dispatchers/SystemDispatcher.swift
+++ b/Sources/LogicProMCP/Dispatchers/SystemDispatcher.swift
@@ -1189,12 +1189,9 @@ struct SystemDispatcher: OperationTraceDispatching {
 
     private enum TraceClearReceiptError: Error {
         case encodingFailed
-        case openFailed(errno: Int32)
         case writeFailed
-        /// The audit directory resolved OUTSIDE its intended parent (symlink
-        /// planted for the final component) — the receipt must never be
-        /// diverted out of the containment root, so the clear is refused.
-        case auditDirectoryEscaped
+        // Open/create/directory/durability failures propagate as SecureFD.FDError
+        // and are caught by the clear_traces handler's fail-closed catch.
     }
 
     /// PRD-015 receipt fields — counts + timestamps ONLY, never trace contents,
@@ -1242,26 +1239,58 @@ struct SystemDispatcher: OperationTraceDispatching {
     /// can delete receipt lines is itself an evidence-destruction surface —
     /// precisely the failure mode this receipt exists to close.
     private static func appendTraceClearAuditLine(_ receipt: [String: Any]) throws -> String {
-        let root = auditLogRoot
-        try FileManager.default.createDirectory(at: root, withIntermediateDirectories: true)
-        // The directory now EXISTS, so re-resolve it through realpath BEFORE
-        // any write: a symlink planted for the audit dir would otherwise divert
-        // the receipt — and with it the whole evidence trail — outside the
-        // intended root, which is indistinguishable from having no receipt.
-        guard auditDirectoryIsContained(root) else {
-            throw TraceClearReceiptError.auditDirectoryEscaped
-        }
         let receiptURL = traceClearReceiptURL
         let data = try JSONSerialization.data(withJSONObject: receipt, options: [.sortedKeys])
         guard let line = String(data: data, encoding: .utf8) else {
             throw TraceClearReceiptError.encodingFailed
         }
         let bytes = Array((line + "\n").utf8)
-        let fd = open(receiptURL.path, O_WRONLY | O_APPEND | O_CREAT, 0o644)
-        guard fd >= 0 else { throw TraceClearReceiptError.openFailed(errno: errno) }
-        defer { close(fd) }
-        let written = bytes.withUnsafeBytes { write(fd, $0.baseAddress, $0.count) }
+
+        // Materialize + open + append strictly through SecureFD (fd-relative,
+        // no-follow). The audit directory chain is created with no-follow opens
+        // so a symlink planted for any component fails closed instead of
+        // diverting the receipt outside the intended root; the log is bound with
+        // O_NOFOLLOW + regular-file + st_nlink==1 so a symlink or hardlink at the
+        // final component is refused; and the line is flushed to stable storage
+        // before returning (the receipt is written one actor hop before the store
+        // is drained, so it must be durable first). This replaces the earlier
+        // create-then-realpath-check-then-open-by-path path, whose containment
+        // check ran AFTER the directory was created — a check-after-use that the
+        // fd-relative walk removes.
+        #if DEBUG
+        // TEST-ONLY: the audit-root override relocates the receipt under a temp
+        // base; bridge it to SecureFD's trusted-base seam. Release has neither
+        // the override nor the seam (production base is home).
+        let auditOverride = ProcessInfo.processInfo
+            .environment["LOGIC_MCP_AUDIT_LOG_ROOT_OVERRIDE"]
+        if let auditOverride, !auditOverride.isEmpty {
+            SecureFD._testBaseOverride = FileManager.default.temporaryDirectory
+        }
+        defer { SecureFD._testBaseOverride = nil }
+        #endif
+
+        let (baseFD, relative) = try SecureFD.openTrustedBase(for: receiptURL)
+        defer { close(baseFD) }
+        guard let logName = relative.last else { throw TraceClearReceiptError.encodingFailed }
+        let dirFD = try SecureFD.ensureDirectory(
+            baseFD: baseFD, components: Array(relative.dropLast()), mode: 0o700
+        )
+        defer { close(dirFD) }
+
+        // Dual first-bind: create the log exclusively the first time; if a
+        // concurrent creator won the race (EEXIST) fall back to a no-follow
+        // append. Both paths refuse a symlink or hardlink at the name.
+        let fileFD: Int32
+        do {
+            fileFD = try SecureFD.createFile(parentFD: dirFD, name: logName, mode: 0o600)
+        } catch SecureFD.FDError.componentOpenFailed(_, let e) where e == EEXIST {
+            fileFD = try SecureFD.openAppend(parentFD: dirFD, name: logName)
+        }
+        defer { close(fileFD) }
+
+        let written = bytes.withUnsafeBytes { Darwin.write(fileFD, $0.baseAddress, $0.count) }
         guard written == bytes.count else { throw TraceClearReceiptError.writeFailed }
+        try SecureFD.fullfsync(fileFD)   // durable before the store is drained
         return receiptURL.path
     }
 
@@ -1290,22 +1319,6 @@ struct SystemDispatcher: OperationTraceDispatching {
                 actualClearedCount: actualClearedCount
             )
         )
-    }
-
-    /// PRD-015 symlink hardening: the audit directory must resolve INSIDE its
-    /// intended parent — production `~/Library/Logs/LogicProMCP`, or the
-    /// DEBUG override root's parent when the test hook is set. Mirrors the
-    /// support-bundle containment check (`createdBundleDirectoryIsContained`):
-    /// realpath BOTH sides, then require the directory to resolve strictly
-    /// under the parent. Because `realpath` follows the final component while
-    /// the parent is resolved independently, a symlink planted for the audit
-    /// dir resolves outside the parent and is caught. Fail-closed when either
-    /// side cannot be canonicalized — an unresolved path is never contained.
-    static func auditDirectoryIsContained(_ directory: URL) -> Bool {
-        guard let parentPath = realpathResolvedPath(directory.deletingLastPathComponent()),
-              let resolved = realpathResolvedPath(directory) else { return false }
-        return resolved.hasPrefix(parentPath + "/")
-            && !resolved.split(separator: "/").contains("..")
     }
 
     /// Resolves a caller-requested bundle directory strictly inside the
--- a/Sources/LogicProMCP/Server/SupportBundle.swift
+++ b/Sources/LogicProMCP/Server/SupportBundle.swift
@@ -142,40 +142,93 @@ struct SupportBundleBuilder: Sendable {
         await onFirstWrite()
         return try await Self.runBlocking {
             try Self.requireAbsent(directory)
-            let fileManager = FileManager.default
-            let parent = directory.deletingLastPathComponent()
-            try fileManager.createDirectory(at: parent, withIntermediateDirectories: true)
-            let staging = parent.appendingPathComponent(
-                ".\(directory.lastPathComponent).tmp-\(UUID().uuidString)",
-                isDirectory: true
-            )
-            try Self.createExclusiveDirectory(staging)
-            var moved = false
-            defer {
-                if !moved { try? fileManager.removeItem(at: staging) }
+            // Materialize, publish, and read back strictly through SecureFD
+            // (fd-relative, no-follow): a symlink planted for any component of the
+            // caller-derived bundle path can no longer divert the staging
+            // directory, the per-file writes, the publish, or the readback outside
+            // the trusted base. Replaces the path-based createDirectory + Data.write
+            // + Data(contentsOf:) and the post-create realpath containment recheck.
+            // Trusted anchor = the deepest EXISTING ancestor of the target
+            // (confined under the support-bundle root by the caller's
+            // resolvedSupportBundleDirectory in production; a temp root under test),
+            // opened by path like the home base. The not-yet-existing tail is then
+            // created + walked strictly no-follow below. Deriving the base per call
+            // (no global seam) keeps concurrent publishers independent.
+            var anchor = directory.standardizedFileURL
+            var tail: [String] = []
+            while !FileManager.default.fileExists(atPath: anchor.path), anchor.pathComponents.count > 1 {
+                tail.insert(anchor.lastPathComponent, at: 0)
+                anchor = anchor.deletingLastPathComponent()
             }
+            let baseFD = anchor.path.withCString { open($0, O_RDONLY | O_DIRECTORY | O_CLOEXEC) }
+            guard baseFD >= 0 else { throw POSIXError(POSIXErrorCode(rawValue: errno) ?? .EIO) }
+            defer { close(baseFD) }
+            // `requireAbsent` proved the target does not exist, so the tail is
+            // non-empty (its last element is the bundle directory name).
+            guard let bundleName = tail.last else { throw CocoaError(.fileWriteFileExists) }
+            let parentFD = try SecureFD.ensureDirectory(
+                baseFD: baseFD, components: Array(tail.dropLast()), mode: 0o755
+            )
+            defer { close(parentFD) }
+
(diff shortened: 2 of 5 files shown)


MongLong0214:
### Verification — head `5257b53`

Verified at head `5257b53` (base `942693b`).

**Hardening (#417): path-based → directory-fd-relative / no-follow / identity-checked / crash-durable**
- Trace-clear receipt: `O_NOFOLLOW` + `fstat`(regular file) + `fsync` before return (durable before the store drains).
- Support-bundle: fd-relative `openat(O_CREAT|O_EXCL|O_NOFOLLOW)` + `fstat` + `fsync` writes; publish via `renameatx_np` relative to parent directory fds; readback relative to a fresh no-follow directory fd with identity checks.

**Gates**
- Full suite at head `5257b53`: **3147 passed, 0 issues**.
- 6 RED→GREEN negatives; mutation-verified — dropping either `O_NOFOLLOW` guard turns the matching negative RED, then restore.
- public-surface preflight (base..head): PASS. `Package.resolved` untouched.
- LIVE: N/A — server-side filesystem hardening; the observable effect is a typed fail-closed refusal on a hostile final component, covered by the deterministic negatives (no Logic-UI surface changed).

MongLong0214:
### Verification update — head `f8a69ea`

Round 2 (independent-review findings folded in). Verified at head `f8a69ea` (base `942693b`).

**Additional hardening since the prior head:**
- Trace-clear receipt now opens via `openat` relative to a no-follow **directory fd** (closes the intermediate-component TOCTOU a path-based open left), with `O_NONBLOCK` (a fifo at the receipt path fails fast instead of blocking the clear), an EINTR-safe **full write loop** (no partial/truncated receipt line), and `F_FULLFSYNC` of the file + `fsync` of the audit directory (media + dirent durability).
- Support-bundle: added `fsync(stagingFd)` after the file creates and `fsync(destinationParent)` after the publish rename (dirent durability).

**Gates (head `f8a69ea`)**
- Full suite: **3148 passed, 0 issues**.
- **7 RED→GREEN negatives**: receipt symlink (`O_NOFOLLOW`), directory (open-refuse), fifo (`O_NONBLOCK`, no hang); fd-relative helper refusals. Mutation-verified: dropping either `O_NOFOLLOW` guard turns the matching negative RED.
- public-surface preflight (base..head): PASS. `Package.resolved` untouched.
- LIVE: N/A — server-side filesystem hardening; observable effect is a typed fail-closed refusal on a hostile final/intermediate component.

MongLong0214:
### Verification update — head `eb9894f`

Round 3 (further review findings folded in). Verified at head `eb9894f` (base `942693b`).

- **EINTR-safe** write/read loops on the support-bundle path (matching the receipt loop — an interrupted syscall no longer fails a valid publish/readback).
- **Post-rename destination fsync is best-effort**: `renameatx_np` (RENAME_EXCL) is the atomic commit, so a subsequent flush failure no longer re-presents an already-published bundle as failed (which would orphan it and break retry).
- **Honest scope** (comments + test names corrected, no over-claim): the closed guarantees are final-component no-follow / regular-file identity / fd-relative publish / crash-durability. The audit + bundle directory *paths* are still opened multi-component, guarded by the existing realpath containment check and by the roots living under the user's own home; a fully symlink-race-proof form would `openat` each component from a trusted root fd — a documented follow-up, not silently omitted.

**Gates (head `eb9894f`)**: full suite **3148 passed / 0 issues**; 7 RED→GREEN negatives, mutation-verified (`O_NOFOLLOW` guards); preflight PASS; `Package.resolved` untouched; LIVE N/A (server-side filesystem hardening).

MongLong0214:
### Verification update — head `f8c7bf6`

Round 4 (further review folded in). Verified at head `f8c7bf6` (base `942693b`).
- **Real fix:** `readFileRelative` now opens with `O_NONBLOCK` too — a fifo planted at a bundle name during readback fails fast instead of blocking the open forever (matching the receipt path).
- **Test strengthened:** added an `O_EXCL`-vs-pre-existing-**regular-file** negative (refuses + leaves the original bytes untouched), isolating exclusive-create from the symlink case.

**Gates (head `f8c7bf6`):** full suite **3149 passed / 0 issues**; RED→GREEN negatives incl. mutation-verified `O_NOFOLLOW` guards; preflight PASS; `Package.resolved` untouched.

**Documented residual (tracked follow-up, not a regression):** the audit + bundle directory *paths* are opened multi-component, so an intermediate-parent symlink race is bounded (not eliminated) by the realpath containment check + user-home roots; full closure = per-component `openat` traversal.

MongLong0214:
## Order 4 exact-head re-gate: FAIL

Exact head reviewed: `f8c7bf6956a86d46a02fe558059473ef07133fee` (base `942693b4411d6b7d467e2ea64a0432321da5e113`).

Decision: require Round 5 before Order 4 can be terminal.

- The prior gate explicitly required deterministic coverage and closure of parent-directory replacement / rename TOCTOU. The current implementation opens the audit and bundle directory paths with multi-component `open(path, O_NOFOLLOW)`, which protects only the final path component. An intermediate parent can still be replaced between the realpath containment check and the path open. Treating this as a non-blocking follow-up would narrow the controlling acceptance rather than close it.
- The new fd-relative final-component work is materially correct: receipt and bundle-file opens are no-follow and identity checked, FIFO opens fail fast, write/read loops retry `EINTR`, staging metadata is synced before publish, and publication uses parent directory descriptors with an exclusive atomic rename.
- Crash-durability remains incomplete as an asserted terminal guarantee. The destination-parent sync after the publish rename is best-effort and its failure is discarded, so the successful result cannot prove that the destination dirent reached durable storage. Round 5 must define and test truthful behavior for that failure boundary without misreporting an already-committed rename as absent.
- Required exact-head gate evidence is incomplete: no release artifact and SHA-256 are bound to `f8c7bf6`. LIVE_NA is acceptable for this server-side filesystem-only change, but it does not waive the release-build/artifact lane.
- Exact-head CI was still running when this verdict was recorded. A later green CI does not cure the acceptance and evidence blockers above.

Required correction: traverse every relevant directory component from a trusted root descriptor using `openat(..., O_DIRECTORY | O_NOFOLLOW)`, add deterministic intermediate-parent replacement/outside-root escape coverage for both audit receipt and support-bundle publication, make the post-commit durability outcome truthful and testable, then rerun focused/full, release build and artifact hash, security review, exact-head CI, and the independent gate on the new head.

Order 4 remains non-terminal. Strict canonical progress remains 3/15 verified; Order 5 implementation remains blocked.

MongLong0214:
### Verification update — head `1ec86d9` (Round 5)

All three gate blockers closed at head `1ec86d9` (base `942693b`).
1. **Intermediate-parent TOCTOU** — new `SecureDirectoryTraversal`: realpath-resolve then per-component `openat(O_DIRECTORY|O_NOFOLLOW)` from `/`, wired into BOTH the audit receipt and support-bundle publish/staging/readback. Deterministic intermediate-swap negatives (helper + trace-clear flow + publish flow) via a DEBUG-only seam.
2. **Post-commit durability** — the destination-parent `fsync` after the publish rename is now ENFORCED: on failure the committed rename is rolled back (destination removed) and surfaced, so the outcome is truthful (nothing orphaned). Deterministic forced-failure test.
3. **Release artifact + SHA-256** — reproducible release build bound to `1ec86d9`: `123dce6ee2280fcb87a686925aa03098243cf236b23c37d94f8ea433d653a5f8` (codesign valid, tree clean).

**Gates:** full suite **3153 / 0 issues**; RED→GREEN negatives incl. mutation-verified 3× `O_NOFOLLOW` guards; preflight PASS; `Package.resolved` untouched; LIVE_NA (server-side filesystem-only).

MongLong0214:
## Order 4 gate decision at `1ec86d9`: B, reviewed design first

Exact head: `1ec86d9df75cdffaaf3120c672ded7c61576f953` (base `942693b4411d6b7d467e2ea64a0432321da5e113`).

Round 5 is not accepted as terminal. The newly reported traversal, rollback, directory-durability, identity, and test-integrity defects invalidate the security review and exact-head gate despite the green suite, release artifact, and CI.

Proceed with a design-only Round 6 before further implementation. Do not mutate the candidate branch until the design contract is reviewed and explicitly accepted.

The design must lock these invariants and corresponding adversarial tests:

1. Preserve the caller's original path semantics. Never `realpath` an untrusted candidate before no-follow traversal.
2. Start from an explicitly trusted base descriptor and traverse every original component with `openat(..., O_DIRECTORY | O_NOFOLLOW)`; reject empty, `.`, `..`, absolute, and otherwise ambiguous components at the boundary.
3. Bind the validated path to stable directory identities. Capture and compare `(st_dev, st_ino)` at the required boundaries so a rename to another real directory cannot silently change the target.
4. Create staging state relative to the validated parent descriptor with `mkdirat`, and perform publication relative to pinned source/destination parent descriptors.
5. Roll back only through the pinned parent descriptor and exact created entry using fd-relative operations such as `unlinkat`; no path-based cleanup after commit.
6. Define truthful commit and durability states. Required file and directory metadata flushes use the platform's stable-storage primitive, including the audit directory on first create. A post-rename durability failure must not be reported as either an absent publish or a durable success unless rollback is independently verified against the pinned identity.
7. Strengthen negatives with mutation checks that fail when each relevant no-follow, identity, fd-relative rollback, or durability guard is removed. Tests must prove outside-root zero-write/zero-delete behavior and verify the original hostile object remains unchanged.
8. Cover both trace-clear audit and support-bundle staging, publish, readback, and rollback flows, including intermediate swaps, final symlinks, alternate real-directory renames, FIFO/device/non-regular endpoints, partial failures, and rollback failure.

After design acceptance: implement only the ratified design, then bind focused/full tests, release artifact and SHA-256, LIVE_NA rationale, security review, and CI to the new exact head before requesting another independent gate.

Strict canonical progress remains 3/15 verified. Order 4 remains non-terminal, and Order 5 implementation remains blocked. Read-only isolated work is permitted; canonical reordering is not approved.

MongLong0214:
## Order 4 Round 6 design contract v2: NOT RATIFIED

Reviewed document: `reviews/order4-round6-secure-traversal-design-contract.md` v2.

The v2 corrections are directionally sound: no realpath-derived authorization, fd-relative file opens, explicit commit/rollback states, ordered stable-storage flushes, fd-relative cleanup, and mutation-sensitive negatives are all required. Directory `F_FULLFSYNC` was also empirically confirmed to succeed on the target macOS filesystem.

One terminal blocker remains in the spine:

### P0: pinned identity does not prove pinned-root ancestry

A held directory descriptor and its `(st_dev, st_ino)` prove which inode is open. They do not prove that inode remains beneath the trusted base after it is opened. An attacker able to rename a validated ancestor or leaf directory after the walk can move that same inode outside the trusted root. Subsequent `openat`, writes, readback, or cleanup through the held descriptor still operate successfully on the same inode, now outside containment. Reopening only when a name is otherwise needed does not cover the no-rewalk single-epoch path, and `fstat` identity cannot detect the namespace move.

Therefore I1/I3's statement that containment is expressed as pinned-root `(dev,ino)` ancestry is currently an assertion without an enforcement mechanism. Ratifying it would repeat the same class of vacuous control that v2 correctly removed elsewhere.

### Required v3 amendment

1. State the concurrency threat boundary precisely: whether an attacker may rename any suffix ancestor/leaf after acquisition and before or during mutation.
2. If such renames are in scope, specify an enforceable primitive or proof that preserves ancestry for the whole mutation epoch. A held fd, component identity, and re-open-by-name check are not sufficient by themselves.
3. If the platform exposes no such primitive, move the trusted boundary to a parent whose namespace entries cannot be renamed by the attacker, or explicitly narrow the threat model and acceptance. Do not describe structural confinement at walk time as continuous ancestry.
4. Add a decisive design test: after the no-follow walk, rename the validated leaf or an ancestor outside the trusted base, then attempt the first write and rollback. The contract must require refusal and outside-root zero-write/zero-delete. If the proposed primitives cannot make that test pass, the design is not implementable under the stated threat model and must say so before code changes.
5. Specify recursive teardown mechanics: enumerate only through the pinned directory descriptor, classify children with no-follow metadata, recurse by `openat(..., O_DIRECTORY | O_NOFOLLOW)`, and unlink only through the pinned parent. No path reconstruction or symlink following.

All other v2 requirements remain carried forward, not discarded. No candidate-branch implementation is authorized until the ancestry/concurrency contradiction is resolved and v3 is explicitly ratified.

Order 4 remains non-terminal. Strict canonical progress remains 3/15 verified; Order 5 remains blocked.

MongLong0214:
## Order 4 Round 6 design contract v5: RATIFIED AS A

Decision: **A, reachable-bar hardening with an explicit platform residual.**

The v5 contract is ratified with the binding amendments now written into the design document:

- An existing trace receipt after process restart must establish authority through two independently verified parent walks and two no-follow opens with matching `(dev,ino)`, regular-file type, and `st_nlink == 1`. It may not depend on an in-memory identity from an earlier process, and a single open cannot adopt it.
- `durable_success` is strictly a persistence result for the matched object and directory metadata at the post-rename verification point. It is not a zero-race, continuous-ancestry, or TOCTOU-closed claim. The documented final-syscall and post-verification movement residual remains visible in code comments, evidence, and completion receipts.

Ratification basis:

1. macOS does not expose a userspace beneath-root primitive equivalent to `openat2(RESOLVE_BENEATH)`, so the contract correctly stops claiming an impossible zero-residual guarantee.
2. The reachable controls are concrete and testable: single-component no-follow traversal, two-pass identity vectors before each operation, exclusive first-bind, link-count/type checks, ordered stable-storage flushes, verified post-rename identity, fd-relative teardown, truthful partial/uncertain outcome states, and mutation-sensitive negatives across both surfaces.
3. The remaining race is narrowly stated under the home-root threat model and is not promoted into `durable_success` or deterministic off-root protection.

Implementation is authorized only against this ratified v5 contract. No additional security claims, path-based fallback, multi-component suffix open, ordinary `fsync` substitution, or silent rollback/durability downgrade is allowed.

After implementation, every SHA-bound lane is fresh: focused/full tests, release artifact and SHA-256, LIVE_NA rationale, exhaustive security review, exact-head CI, and independent CEO gate. The current head `1ec86d9df75cdffaaf3120c672ded7c61576f953` remains non-terminal and its prior evidence cannot qualify the new head.

Strict canonical progress remains 3/15 verified. Order 5 remains blocked until Order 4 is merged and post-merge completion evidence is verified.

MongLong0214:
## Order 4 Round 6 implementation kickoff: APPROVED WITH CHECKPOINTS

The four-phase decomposition is approved against the ratified v5 contract and its binding amendments.

### Commit and review structure

- Use staged, auditable commits rather than one squashed implementation commit: Phase 1 core primitive, Phase 2 trace-clear wiring, and Phase 3 support-bundle wiring. Evidence-only changes remain separate where practical.
- Phase 1 is a mandatory exact-commit review checkpoint. Do not wire either production surface until the core diff, focused tests, mutation matrix, release-configuration seam exclusion, and fd-lifecycle audit pass at that commit.
- Phase 2 is a second mandatory checkpoint before Phase 3. It must prove both filesystem safety and truthful receipt semantics, including restart first-bind, `EEXIST`, partial append, durability failure, drain/completion divergence, and zero-clear-on-pre-drain failure.
- Phase 3 receives its focused review before final evidence collection.
- These checkpoints are development controls, not reusable terminal evidence. Focused/full tests, release artifact and SHA-256, security review, and CI are all rerun and bound to the final exact head.

### Additional locked invariants

1. Production accepts no arbitrary base-path injection; DEBUG seams and overrides are absent from the release configuration.
2. Every eligible descriptor uses close-on-exec, and ownership is explicit across every success/error path. `fdopendir` owns only a duplicated fd; leak, double-close, and stale-fd reuse tests are required.
3. A pre-drain audit line is an intent/snapshot, not a completed-clear record. Partial append or pre-drain durability failure leaves the store intact and cannot later parse as a completed clear.
4. If the drain succeeds but the durable completion/amendment fails, report an applied-but-receipt-incomplete state. Never return ordinary success or claim that the store was untouched.
5. Receipt creation that loses an `EEXIST` race must enter the independently re-walked dual first-bind path.

Implementation may begin at Phase 1. The branch remains non-terminal throughout the checkpoints. Order 5 remains blocked, and no merge is authorized before the final independent exact-head PASS.

Strict canonical progress remains 3/15 verified. Current candidate before implementation: `1ec86d9df75cdffaaf3120c672ded7c61576f953`, base `942693b4411d6b7d467e2ea64a0432321da5e113`.

MongLong0214:
## Phase 1 exact-commit checkpoint: FAIL

Exact head reviewed: `8e26a06e0ab649b08ad99f6e31f8c66f0ce54ea4`  
Base: `942693b4411d6b7d467e2ea64a0432321da5e113`

The submitted scope is correctly isolated to two new files and one commit. Independent checks passed for:

- focused `SecureFD primitive`: 9/9
- release build
- release seam exclusion: no `_testBaseOverride` or `_testAfterPassAHook` symbol found
- public-surface scrub and diff check
- mutation matrix: removing walk `O_NOFOLLOW`, A==B identity comparison, or append `O_NOFOLLOW` made only the corresponding negative fail; the other eight tests remained green in each lane
- source and test files restored byte-for-byte after the mutations

### Blocking invariant-2 defect: pass-A descriptor leaks when pass B throws

`walkVerified` acquires `fdA`, then calls the second `walkOnce` before establishing cleanup ownership for `fdA`. If pass A succeeds and pass B fails while opening a component, the throw bypasses `close(fdA)`.

A deterministic review probe ran 32 pass-A-success / pass-B-open-failure iterations. Open descriptors grew from 6 to 38. This violates the locked requirement that descriptor ownership be explicit on every success and error path.

There is also a leak in the isolated test itself: `emptyDirGuards` passes `try openBaseFD(base)` directly into `walkVerified` and never closes that base descriptor. The current lifecycle coverage therefore does not establish leak-free behavior.

Required before a new Phase 1 checkpoint:

1. Establish cleanup ownership for pass-A immediately after acquisition so every pass-B throw closes it exactly once.
2. Close the base descriptor in `emptyDirGuards`.
3. Add a deterministic error-path lifecycle negative that repeats pass-B failure and proves descriptor count remains bounded; mutation-check the cleanup guard.
4. Re-run the 9-test primitive suite, the new lifecycle negative, all three existing mutation lanes, release build/seam exclusion, and source restoration at the new exact head.

Phase 2 is not authorized yet. PR #418 remains open and non-terminal; CI success does not override this checkpoint failure. Strict canonical progress remains 3/15, and Order 5 remains blocked. The PR description also still describes the superseded wired implementation and old verification totals; update it when the phased implementation advances so the public surface matches the current head.

MongLong0214:
## Phase 1 exact-commit re-review: FAIL

Exact head reviewed: `75035c820a3aeeb2e257fadef01ec40d8bf17bf9`  
Base: `942693b4411d6b7d467e2ea64a0432321da5e113`

The prior checkpoint defects are materially corrected:

- pass A is closed before the verification hook and pass B, so a pass-B open failure no longer leaks the pass-A descriptor;
- recursive teardown establishes single deferred ownership of every directory descriptor before any recursive throw can propagate;
- `emptyDirGuards` now owns and closes its base descriptor;
- focused `SecureFDTests`: **10/10 PASS**;
- independent repeated probes: pass-A-success/pass-B-open-failure **256 failures, fd 4 -> 4**; teardown mid-recursion failure **256 failures, fd 4 -> 4**;
- release build PASS; release artifact scan found no `_testBaseOverride` or `_testAfterPassAHook` symbol;
- the PR description now reflects the primitive-only stage.

### Blocking locked-invariant defect: `dup()` removes close-on-exec

The approved Phase 1 checkpoint requires every eligible descriptor to use close-on-exec. Three production paths use plain `dup()`:

- `walkOnce` (`SecureFD.swift:151`)
- `assertEmpty` (`SecureFD.swift:235`)
- `removeChildrenThenSelf` (`SecureFD.swift:273`)

On Darwin, the duplicated descriptor has `FD_CLOEXEC` cleared. A direct probe confirmed `source_cloexec=true`, `dup_cloexec=false`. This makes trusted directory descriptors inheritable across a concurrent `exec`; an empty verified walk can also return the non-CLOEXEC duplicate directly. Replace these with an atomic close-on-exec duplication primitive such as `F_DUPFD_CLOEXEC`, and add direct `F_GETFD` assertions for the returned and enumeration descriptors.

### Remaining required lifecycle coverage

The approved checkpoint also explicitly requires leak, double-close, and stale-fd-reuse tests. The new test covers repeated leak paths, but there is still no deterministic double-close or stale-fd-reuse negative. Its aggregate `(after - before) < 32` assertion also combines both failure paths and permits as many as 31 persistent leaked descriptors. Keep the two paths independently bounded in the committed test and add the missing ownership assertions.

### Public-surface mismatch

The PR description states that `openRead` enforces `st_nlink == 1`, but `SecureFD.swift:198-209` checks only regular-file type and returns without a link-count guard. Either implement and mutation-test the stated existing-file identity policy or narrow the public claim before resubmission.

Phase 2 remains unauthorized. Exact-head CI was still running when this verdict was recorded, but later CI success cannot override the locked lifecycle defect. Strict canonical progress remains 3/15; Order 5 remains blocked.

MongLong0214:
## Phase 1 exact-commit third review: FAIL

Exact head reviewed: `af71133ec60e4f7a814b333abbcf45fba5c786f2`  
Base: `942693b4411d6b7d467e2ea64a0432321da5e113`

The submitted R2 production corrections are real:

- all three internal duplications now use atomic `F_DUPFD_CLOEXEC`; a direct runtime probe observed `source_cloexec=true`, `duplicate_cloexec=true`;
- `openRead` now enforces `st_nlink == 1` and rejects a hardlink;
- the repeated walk and teardown leak measurements are separated;
- focused `SecureFDTests`: **13/13 PASS**, including the returned-descriptor, leak-bound, and ownership tests;
- scope remains the isolated primitive and tests only.

The Phase 1 checkpoint nevertheless remains incomplete.

### 1. Required enumeration-descriptor CLOEXEC evidence is missing

The prior correction explicitly required direct `F_GETFD` assertions for returned and enumeration descriptors. `returnedDescriptorsAreCloexec` checks six returned paths, including the empty-relative duplicate, but it never observes the duplicated descriptors consumed by `fdopendir` in `assertEmpty` and `removeChildrenThenSelf`. Routing them through a shared helper is sound implementation, but it is not the requested direct lifecycle assertion. The test also does not assert the trusted-base descriptor returned by `openTrustedBase` even though the PR description says every returned descriptor is checked.

Add a narrow DEBUG observation seam or equivalent controlled probe that asserts `FD_CLOEXEC` on both enumeration duplicates before ownership transfers to `fdopendir`, plus the trusted-base return.

### 2. The identity primitive is not fail-closed

`identity(ofFD:)` discards the return value of `fstat` and returns the zero-initialized `stat` fields on failure (`SecureFD.swift:59-63`). That permits a failed identity read to become `(0,0)`; two failed reads can compare equal and the code can return a fabricated identity instead of refusing the operation. Make identity acquisition throwing, preserve the causal error, propagate it through walks/directory creation, and add a mutation-sensitive failure negative.

### 3. The ownership negatives do not yet prove the claims deterministically

The double-close canary does not record the released pass-A descriptor number or prove that a canary reclaimed that exact slot. A concurrent process-level descriptor allocation can take the slot between the close and the hook's opens, so the test can miss a residual close or fail from unrelated descriptor churn. The stale-fd check compares only `(dev,ino)`; closing and reopening the same directory preserves that identity and therefore passes despite descriptor replacement.

Bind both negatives to the exact released descriptor generation/slot under controlled ownership. The committed test must fail under an injected residual close and under close-and-reopen of the caller descriptor, not merely under the current implementation.

### 4. Recursive teardown can mutate a replacement real directory

`removeChildrenThenSelf` classifies a child with `fstatat(..., AT_SYMLINK_NOFOLLOW)`, then later opens the same name without comparing the opened identity to the classified identity (`SecureFD.swift:296-306`). A namespace replacement with another real directory is not blocked by `O_NOFOLLOW`; recursion can delete the replacement's contents. The top-level final `unlinkat(parentFD, name, AT_REMOVEDIR)` likewise does not prove that the name still denotes the opened directory.

Bind classification/open/final removal to the expected `(dev,ino)` and add a deterministic real-directory replacement negative proving wrong-object zero-delete. Keep the already-ratified final-syscall movement residual explicit; do not broaden it into permission to recursively mutate an unverified replacement.

Exact-head CI was still running when reviewed. The seven submitted mutation lanes were not reused as approval evidence because the blockers above terminate the checkpoint first. Phase 2 remains unauthorized; strict canonical progress remains 3/15 and Order 5 remains blocked.

Issue #415, opened 2026-07-19
setup_arm_key follow-ups: Korean input-source lane + verify-causality tightening


MongLong0214:
Follow-ups from #413 (PR #414), none blocking that PR:

**Note:** the Korean input-source lane is now IN #413's acceptance axis (Logic 12.3 English UI × {ABC, Korean 2-Set} keyboard input sources) and is verified there — it is NOT deferred. Remaining follow-ups:

1. **Verify-causality tightening.** The chord-only arm verification requires a successful chord post before crediting a flip in the same call. Residual: after a genuine post, an external record-enable change landing within the short retry window (3×600ms) could still be credited as the chord's late flip. Consider correlating the observed transition more tightly with the post (single-attempt strict window, or a second confirming flip).

2. **Doc/test sharpness (from adversarial review, non-behavioral):**
   - VerifyResult enum doc says 'only .unmapped may proceed to GUI' but the code correctly also allows .environmentUnavailable — align the doc.
   - The file-header focus wording softens the requirement; the code requires a definitive AXFocused==true — align the wording.
   - Top-level `restored` reflects Learn + window cleanup; the functional verify's restore is reported separately as `verify_restored` — align the docstring.

Issue #413, opened 2026-07-19
Record-arm has no key command by default: one-time consent-based setup is required for coordinate-free arm control


MongLong0214:
## Debt record (owning issue for PR #408)

Coordinate-free record-arm control (`track.set_arm`) relies on Logic's **Toggle Record Enable** key command (Ctrl+Shift+E path), because the arm button's AX press is a no-op on the current Logic build. That key command is **unassigned in a fresh Logic installation**, so arm control silently degrades on new hosts.

### Required behavior (`system.setup_arm_key`)
1. **Consent-first**: no side effect, activation, or UI trace before explicit operator consent in the request.
2. **Verify-first fast path**: before any GUI work, drive a real arm flip with the captured chord; if it works, report State A (already configured; verification mutation observed and restored) with `configuration_write_attempted:false`.
3. **GUI assignment path** (fresh host): open Key Commands, filter by typed search, select the exact command row by identity (name match, search-echo excluded, single-match required), ensure Learn is engaged, post the chord, close via the window's close button — never a synthetic window-kill.
4. **Conflict safety**: any unexpected modal after the chord → decline and fail closed; never confirm/replace an existing owner's binding.
5. **Cleanup on every exit path**; honest envelope (`write_source`, `write_attempted`, per-phase evidence); functional arm-flip verification before claiming State A.
6. Attestation scope: Logic 12.3, en-US and ko-KR Desktop layouts; idempotent re-run must land on the fast path.

MongLong0214:
Acceptance axis clarification (confirmed): **Logic Pro 12.3 English UI × {ABC, Korean 2-Set} keyboard input sources**. Both input sources are verified to State A `gui_assignment` on the exact release artifact. A localized (non-English) Logic UI is out of scope for this issue — the coordinate-free path matches the English command identity 'Toggle Track Record Enable', so a localized command list fails closed with a manual-assignment hint (the correct contract). PR #414 delivers this axis.

PR #421, opened 2026-07-21
fix(#415): pin arm-verify two-transition causality; align VerifyResult docs

MongLong0214:
## Summary
Closes the #413 follow-up #415 (both items).

**Item 1 — verify-causality:** analysis shows the correlation #415 asked to consider ("a second confirming flip") is already the shipped design: `.verified` requires the flip leg AND the restore leg — two opposite chord-correlated transitions, each gated by an arm-not-yet-at-target read immediately before a successful post. A single external record-enable change (inside one observation window or a settle gap) satisfies at most one leg, so with an unmapped chord it degrades to `.partialRestore` (fail-closed, never GUI assignment), and before any post it is never credited. That invariant was undocumented and unpinned: this adds the causality doc to `armSetupVerify` (including the honestly accepted irreducible residual — double opposite external interference, one inside each leg's window, is observationally indistinguishable by AX read-back) and two regression tests covering exactly the scenarios the issue raised (observe-window flip; settle-gap flip via the late-publish credit). Mutation-proven: weakening `.verified` to flip-leg-only makes exactly those two tests fail. A stricter single-attempt window was considered and rejected (does not remove the double-interference residual; costs live reliability against slow synthetic-key delivery).

**Item 2 — doc alignment (3 sub-items):** two were already aligned (definitive `AXFocused == true` header wording; `restored`/`verify_restored` split). The third had drifted opposite to the issue's snapshot: the code now fails closed on `.environmentUnavailable` (`verify_environment_unavailable`, no Key Commands GUI), but three doc sites still claimed it falls through to GUI assignment. Those three are aligned to the code (behavior unchanged; an existing regression test already pins the fail-closed path).

Production-code changes are doc comments only; the behavioral surface of this PR is the two new tests.

## Testing
- Focused arm-verify suite 10/10 green; mutation run: flip-leg-only weakening fails exactly the 2 new pins.
- Full serial suite: 3168/3171 with the ONLY failure being the live-coupled `realReleaseBinaryExecutesEveryOperationWithIndependentReadback` (release-binary × live Logic qualification QA). That test executed and PASSED twice earlier today against a binary built from tree content identical to main; it now fails repeatably with the same binary, tracking live Logic session drift, not this diff (the Sources diff is comment-only, mechanically verified). CI does not run live-coupled tests.

### DIFF
--- a/Sources/LogicProMCP/Accessibility/ArmKeyCommandSetup.swift
+++ b/Sources/LogicProMCP/Accessibility/ArmKeyCommandSetup.swift
@@ -176,9 +176,10 @@ enum ArmKeyCommandSetup {
 
     /// Typed outcome of the functional arm-flip verification, so a partially
     /// mutated host is never treated the same as a cleanly unmapped one. From
-    /// verify-first, only `.unmapped` and `.environmentUnavailable` fall through to
-    /// the GUI assignment; `.partialRestore` and `.couldNotPost` fail closed (never
-    /// pile GUI mutations onto a dirty/undrivable host).
+    /// verify-first, only a clean `.unmapped` falls through to the GUI
+    /// assignment; `.partialRestore`, `.couldNotPost`, AND `.environmentUnavailable`
+    /// all fail closed (never pile GUI mutations onto a dirty/undrivable host, and
+    /// never run a GUI assignment that could not be functionally verified). (#415)
     enum VerifyResult: Equatable {
         /// The chord flipped record-enable AND every mutation was restored.
         case verified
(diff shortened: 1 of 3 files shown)


MongLong0214:

- **Deterministic gates:** focused suite 10/10 green; full serial suite green except one test (below); public-surface preflight exit 0; PR CI and post-merge main CI (`ddb0af2`) both green.
- **Live behavior:** arm-verify two-transition causality is doc + tests only (no production behavior change); two causality pins mutation-proven.

PR #422, opened 2026-07-21
refactor: remove region.select_last coordinate-click fallback (fail closed)

MongLong0214:
## Summary
Coordinate-free campaign, step (a) of the residual-census closure: removes the last AppleScript `click at {x, y}` in the production tree — the `region.select_last` fallback that clicked a heuristic-located region's center when the `AXSelected` write threw. A failed write now returns a typed `SELECT_FAILED` marker and the handler fails closed (State C `ax_write_failed`, "no fallback was attempted"); the success path and its independent read-back verification (selected-vs-last identity, State A/B) are unchanged.

Why this is safe (independently reviewed):
- The operation is orphaned in production: it has no OperationRegistry spec and no dispatcher routes it (registry-gated dispatch means no tool surface reaches it) — removal costs no live capability.
- The click was positional actuation on an AX-addressable target — exactly the class the campaign bans — and the readback (not the click marker) was always the authority, so a click that failed to select already surfaced as State B.
- `ax_write_failed` is the right taxonomy: the write threw (attempted-and-failed), not readback ambiguity; consistent with the op's existing error mapping.

Also aligns the `point(fromRawAttribute:)` doc comment (its cited coordinate-click callers were removed by earlier campaign steps; the remaining consumer is a fail-closed position read).

## Testing
- New regression test pins SELECT_FAILED → fail-closed; mutation-proven (disabling the handler branch fails it, falling through to a State B envelope).
- Focused select_last suite 6/6; full serial suite 3171/3172 — the only failure is the known live-coupled `realReleaseBinaryExecutesEveryOperationWithIndependentReadback` drift documented on the previous PR (same signature, pre-existing, unrelated: this op is not registry-probed by that runner).
- `grep 'click at' Sources/` now matches only prose describing the removal.

### DIFF
--- a/Sources/LogicProMCP/Accessibility/AXHelpers.swift
+++ b/Sources/LogicProMCP/Accessibility/AXHelpers.swift
@@ -248,12 +248,11 @@ enum AXHelpers {
     /// value (e.g. from `AXUIElementCopyAttributeValue`). Verifies BOTH that it
     /// is an AXValue AND that the `.cgPoint` extraction succeeds, returning nil
     /// (fail-closed) on a non-AXValue or a wrong-subtype AXValue (e.g. a
-    /// `.cgRect` AXValue). Hoists the `AXValueGetValue` Bool check out of four
-    /// coord-click call sites (`AccessibilityChannel.postMouseClickAt` /
-    /// `.trackViewport`, `AXLogicProElements` track-header click,
-    /// `LibraryAccessor` header click) that previously ignored it and would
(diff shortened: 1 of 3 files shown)


MongLong0214:

- **Deterministic gates:** focused suite 6/6 green; full serial suite green except one test (below); public-surface preflight exit 0; PR CI and post-merge main CI (`7456bb7`) both green.
- **Live behavior:** removed an orphaned coordinate-click fallback (op not registry-routed); fail-closed mapping mutation-proven.

PR #424, opened 2026-07-21
refactor: retire the external click-tool fallback; doctor reports native-only

MongLong0214:
## Summary
Coordinate-free campaign, step (c): the Library row clicks' external-binary (`cliclick`) fallback is retired, and the doctor's `dependencies.click_fallback` check is aligned to the new reality.

- The fallback fired only when the native CGEvent post returned `false` — which happens only on CGEvent *construction* failure (headless/no WindowServer), where the external tool fails identically (it needs the same WindowServer + Accessibility). Independently reviewed conclusion: **no real environment existed where the retired rung was load-bearing.** It spawned an external process per click (audit #16 FD-leak-class) and widened the supply-chain surface for zero live value.
- The Library row clicks themselves remain under their documented AX-opacity evidence (AX selection reports success without changing the visible column; preset commit requires a double-click) — this step removes only the dead external-tool rung.
- Doctor: the check previously PASSED when PostEvent was denied but the tool was present — after retirement that would be a false-green about server capability. It now warns exactly when PostEvent is denied (the hard fail is already carried by `permissions.post_event_access`); the tool's presence is reported as informational evidence. Check id and count unchanged; the only non-test consumer asserts presence, not pass, so no gate is affected.

## Testing
- Two new pins: denied + tool-present must WARN (fails on the old truth table); granted + tool-absent must pass. Focused doctor/library suites 213/213.
- Full serial suite 3173/3174 — the only failure is the known live-coupled qualification-runner drift documented on the previous PRs (same signature, pre-existing, unrelated).
- `grep cliclick Sources/` now matches only the doctor's informational evidence and a historical comment; setup guidance updated to native-only.

### DIFF
--- a/Sources/LogicProMCP/Accessibility/LibraryAccessor.swift
+++ b/Sources/LogicProMCP/Accessibility/LibraryAccessor.swift
@@ -234,13 +234,9 @@ enum LibraryAccessor {
                 usleep(120_000)
                 CGWarpMouseCursorPosition(point)
                 usleep(30_000)
-                // Prefer the native CGEvent double-click (audit #16): spawning
-                // cliclick per click is FD-leak-class. Fall back to cliclick only
-                // if the native post fails.
-                if AXMouseHelper.doubleClick(at: point) {
-                    return true
-                }
-                return LibraryAccessor.postCliclick(command: "dc", at: point)
+                // Native CGEvent only — the external click-tool fallback was
+                // retired (see `productionMouseClick`).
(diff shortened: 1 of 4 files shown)


MongLong0214:

- **Deterministic gates:** focused suite 213/213 green; full serial suite green except one test (below); public-surface preflight exit 0; PR CI and post-merge main CI (`f798036`) both green.
- **Live behavior:** retired the external click-tool fallback (native CGEvent only); doctor `dependencies.click_fallback` aligned native-only with two truth-table pins.

Issue #369, opened 2026-07-14
Per-track stem export (Export All Tracks as Audio Files) via export_run — artifacts:[stem] currently yields a single project-level bounce


thomas-doesburg:
> **STATUS 2026-08-24** — BLOCKED at the UI surface: Logic's Export panel is AX-opaque (measured), so "Export All Tracks as Audio Files" has no accessible drive path.

## Summary

Feature request: **per-track stem export** — i.e. Logic's *"Export All Tracks as Audio Files"* — exposed through the existing `export_plan` / `export_run` scaffold, producing **one verified artifact per track** rather than a single project-level mixdown.

## What I observed

Running `export_plan` with `artifacts: ["stem"]` against a multi-track project (Logic 12.2, v3.4.5-rc4) plans exactly **one** artifact, named after the project:

```json
"naming_policy": "project-name-kind",
"expected_artifacts": [
  { "kind": "stem", "path": ".../Burntout_mix1-stem.wav", "status": "pending" }
],
"workflow_steps": [
  { "id": "project_0_open",   "command": "open",   "tool": "logic_project" },
  { "id": "project_0_export", "command": "bounce", "tool": "logic_project" }
]
```

So today the `stem` artifact kind is effectively a **filename label on a project-level bounce** — a single `bounce` command, one file. If I've misread that and per-track output is already reachable some other way, I'd be glad to be corrected.

## The ask

An artifact kind (e.g. `stems`, plural) that maps to *Export All Tracks as Audio Files* and yields:

```json
"expected_artifacts": [
  { "kind": "stems", "track_index": 0, "track_name": "Kick",  "path": ".../Burntout_mix1-01-Kick.wav" },
  { "kind": "stems", "track_index": 1, "track_name": "Snare", "path": ".../Burntout_mix1-02-Snare.wav" },
  ...
]
```

with each artifact getting the **existing** treatment applied per-file — the baseline verification you already ship (`artifact_exists`, `file_size_non_zero`, `mtime_within_run_window`, `path_under_output_root`, `no_silent_overwrite`) plus `logic_audio.analyze_file`, and rolled into the same `logic_pro_mcp_export_run.v1` manifest with Honest Contract State A/B/C per artifact.

Track-name-derived filenames matter for us (they're the join key back to the mixer strip), but index-only naming would still be usable if names are unreliable to read.

## Why

[...]

PR #426, opened 2026-07-21
feat(#369): surface the export bounce step's execution preconditions in the plan

MongLong0214:
## Summary
Part 1 of #369 (the "Related question"): `export_plan` validated as `validation_status: valid` with no mention that the bounce/export step has live execution dependencies, so a caller could not tell from the manifest that a run could fail at the bounce step. The plan now carries an explicit `execution_preconditions` array (mirroring the existing `unsupported_or_blocked_steps` / `required_confirmations` surfacing).

Three preconditions on the `bounce` command, grounded in the **actual default export path** (the bundled `logic_bounce.py` dialog driver, not the standalone `project.bounce` operation):
- `automation_permission` — the step issues Logic's default Bounce command (Cmd+B) or clicks File ▸ Bounce ▸ Project or Section through System Events → Automation for System Events **and** Logic Pro.
- `post_event_access` — the settings/save panel is driven with native CGEvent (`CGEventPost`) → trusted PostEvent.
- `bounce_helper_available` — the bundled helper must resolve, pass the ownership-trust check, and have python3.

It corrects the reporter's guess in-context: this dialog-driven export path does **not** require the MIDIKeyCommands manual MIDI-Learn binding (that channel is only a fallback route for the standalone `project.bounce` op, whose primary route is also System Events). Preconditions are invariant execution facts, not a live probe (the planner stays pure); each names the `verify_with` command a caller runs to confirm its host.

Additive field; `logic_pro_mcp_export_manifest.v1` schema name unchanged (field-superset). Two unit tests pin the surfaced preconditions + the snake_case JSON round-trip.

**Part 1 of #369 only** — the issue's main ask (per-track stems / Export All Tracks as Audio Files) is Part 2 and is intentionally NOT closed here; #369 stays open until stems land.

## Testing
Focused planner suite 28/28; full serial suite at exact head (sole failure is the whole-project live-qualification E2E, which is fixture/environment-sensitive and not a regression from this doc/unit change — reproduces identically at the pre-change head; reliable whole-suite live QA is tracked under ADR-001 #284). Public-surface preflight exit 0.

Refs #369

### DIFF
--- a/Sources/LogicProMCP/Projects/ProjectExportPlanner.swift
+++ b/Sources/LogicProMCP/Projects/ProjectExportPlanner.swift
@@ -65,6 +65,7 @@ enum ProjectExportPlanner {
                 ),
             ],
             unsupportedOrBlockedSteps: surfacedConstraints,
+            executionPreconditions: executionPreconditions(),
             baselineVerification: [
                 "artifact_exists",
                 "file_size_non_zero",
--- a/Sources/LogicProMCP/Projects/ProjectExportPlannerModels.swift
+++ b/Sources/LogicProMCP/Projects/ProjectExportPlannerModels.swift
@@ -13,6 +13,7 @@ struct ProjectExportPlan: Codable, Sendable, Equatable {
     let projects: [ProjectExportPlanProject]
     let requiredConfirmations: [ProjectExportConfirmation]
     let unsupportedOrBlockedSteps: [ProjectExportBlockedStep]
+    let executionPreconditions: [ProjectExportPrecondition]
     let baselineVerification: [String]
     let enhancementPath: [String]
     let nextSafeAction: String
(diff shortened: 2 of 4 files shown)


MongLong0214:

- **Deterministic gates:** focused planner suite 28/28; adversarial review PASS (precondition accuracy verified against the actual default export path: helper resolution, System Events automation, native CGEventPost; one wording imprecision fixed pre-merge).
- **Full serial suite at exact head (ae4739e): 3176/3176 green — including the whole-project live qualification E2E** (`realReleaseBinaryExecutesEveryOperationWithIndependentReadback`, 108 operations against real Logic 12.3, isolated 19s pass + pass under full-suite load). The earlier red on this test was root-caused to stale Logic session state (only `transport.play/stop` state readbacks timing out; identical failure reproduced at the pre-change head; a Logic restart cleared it with the same project) — environmental, not code.
- Public-surface preflight exit 0; PR CI green; post-merge main CI tracked.
- Part 1 of #369 only — #369 stays open for per-track stems (Part 2).

Issue #427, opened 2026-07-21
v3.11.0 Homebrew release omits Scripts/logic_variants.py — bounce/export fails with ModuleNotFoundError after the project is opened


thomas-doesburg:
## Summary

The Homebrew release of **v3.11.0** ships `logic_bounce.py` (and two other helpers) **without their shared dependency `logic_variants.py`**, so the entire bounce/export path fails at runtime with `ModuleNotFoundError`. `Scripts/logic_variants.py` exists in the repo — it is simply never installed by the Formula.

Found while testing `export_run` end-to-end on Logic Pro 12.3 (this is the same workflow as #369, so it's directly relevant to the in-flight #426).

## Repro

Installed via `brew install monglong0214/logic-pro-mcp/logic-pro-mcp` (v3.11.0), Logic Pro 12.3, all permissions granted (`accessibility`, `automation`, `automation_system_events`, `post_event_access` all `true`), a real 26-track project open.

```
logic_project export_plan  { projects:[…/Burntout_test.logicx], artifacts:["bounce"], … }
  → "validation_status": "valid"

logic_project export_run   { …same…, confirmed: true }
  → status: "failed"
```

Artifact result:

```json
{
  "kind": "bounce",
  "state": "C",
  "verified": false,
  "bounce_fired": false,
  "error": "bounce_helper_failed: Traceback (most recent call last):\n  File \"/opt/homebrew/Cellar/logic-pro-mcp/3.11.0/share/logic-pro-mcp/logic_bounce.py\", line 29, in <module>\n    from logic_variants import logic_process_osa\nModuleNotFoundError: No module named 'logic_variants'"
}
```

Note the project was **opened first** (`"opened": true`, `"identity_verified": true`) and only then did the bounce step fail — i.e. it fails after mutation has begun, which is exactly the failure mode #426 is designed to prevent.

## Root cause — the Formula's install list

`logic-pro-mcp.rb`:

```ruby
pkgshare.install "Scripts/logic_bounce.py"       if (buildpath/"Scripts/logic_bounce.py").exist?
pkgshare.install "Scripts/logic_bounce_ui.py"    if (buildpath/"Scripts/logic_bounce_ui.py").exist?
pkgshare.install "Scripts/logic_ui_jxa.py"       if (buildpath/"Scripts/logic_ui_jxa.py").exist?
pkgshare.install "Scripts/logic_input_source.py" if (buildpath/"Scripts/logic_input_source.py").exist?
[...]

Issue #22, opened 2026-06-12
v3.5.0 (and v3.4.6) Homebrew formula broken: pkgshare.install paths don't match release tarball layout (docs/, Scripts/)


thomas-doesburg:
## Symptom

`brew install` / `brew upgrade` of `monglong0214/logic-pro-mcp/logic-pro-mcp` fails during the install step (sandboxed `build.rb` exits 1) at **v3.4.6 and still at v3.5.0**: the formula (`Formula/logic-pro-mcp.rb`, `version "3.5.0"`) installs five helper assets from the tarball root, but the published `LogicProMCP-macOS-universal.tar.gz` (sha256 `d8c3e955…6489`, matches SHA256SUMS.txt) nests them under `docs/` and `Scripts/` — verified by `tar tzf` on 2026-06-13.

## Cause

The formula installs helper assets from the tarball root:

```ruby
pkgshare.install "SETUP.md"
pkgshare.install "install-keycmds.sh"
pkgshare.install "uninstall-keycmds.sh"
pkgshare.install "keycmd-preset.plist"
pkgshare.install "LogicProMCP-Scripter.js"
```

but the published tarball for both v3.4.6 and v3.5.0 nests them:

```
LogicProMCP
docs/SETUP.md
Scripts/install-keycmds.sh
Scripts/uninstall-keycmds.sh
Scripts/keycmd-preset.plist
Scripts/LogicProMCP-Scripter.js
```

(sha256 matches the formula in both versions — v3.4.6 `6420274b…01d3`, v3.5.0 `d8c3e955…6489` — so it's the layout, not a stale artifact.)

## Fix that works locally

```ruby
pkgshare.install "docs/SETUP.md"
pkgshare.install "Scripts/install-keycmds.sh"
pkgshare.install "Scripts/uninstall-keycmds.sh"
pkgshare.install "Scripts/keycmd-preset.plist"
pkgshare.install "Scripts/LogicProMCP-Scripter.js"
```

With that patch the keg installs and the binary responds to MCP `initialize` (verified at v3.4.6 with `{"name":"logic-pro-mcp","version":"3.4.6"}`).

[...]

PR #428, opened 2026-07-21
fix(#427): package logic_variants.py + assert Python helper import-closure at release

MongLong0214:
## Summary
Fixes the packaging bug @thomas-doesburg reported in #427: the v3.11.0 Homebrew release (and the pinned shell installer) ship `logic_bounce.py` / `logic_bounce_ui.py` / `logic_ui_jxa.py` **without their shared dependency `logic_variants.py`**, so the bounce/export path fails at runtime with `ModuleNotFoundError` once a project is opened.

**Not a version issue** — v3.11.0 is the latest release and `main` carried the identical omission (verified). Root cause is one layer below the Formula: `Scripts/logic_variants.py` was never added to the release tarball (`release-package.sh`), so neither the Formula nor the shell installer could install it, and the existing `#22` tarball-listing gate is one-directional (checks installed-paths ⊆ tarball) so it structurally cannot catch a *missing* install line / dropped dependency.

## Changes
- **Package + install the module:** `release-package.sh` (tarball), `Formula/logic-pro-mcp.rb`, `install.sh` (allowlist + extract + validate + install), `install-common.sh`.
- **Systemic guard (issue's suggestion 1):** `release-verify-formula-install-paths.sh` now asserts the **import closure** — for every `logic_*.py` the Formula installs, every `logic_*` module it imports (read from the *extracted tarball*, i.e. the shipped bytes) must also be installed. A future split/rename that adds an intra-helper dependency now fails the release at tag time instead of silently dropping a module.
- **Contract fixture** mirrors the real package (adds `logic_variants.py` + a real import) so `testReleaseTarballFixtureMatchesFormulaInstallPaths` exercises both the path check and the closure check truthfully.

## Testing
- Install-script contract suite 33/33 green (incl. the previously-red tarball-fixture test).
- Import-closure gate **mutation-proven**: with the fix, closure verifies clean; removing `logic_variants` from the Formula fails the gate naming all three importers (`logic_bounce`, `logic_bounce_ui`, `logic_ui_jxa`). Testing also caught and fixed a `set -euo pipefail` abort in the gate when a helper has no `logic_*` imports.
- `bash -n` clean on all touched scripts; `py_compile` clean on `logic_variants.py`.

Delivering this to Homebrew requires a patch release (v3.11.1) with a refreshed Formula `sha256`. The issue stays open until that ships.

Refs #427

### DIFF
--- a/Formula/logic-pro-mcp.rb
+++ b/Formula/logic-pro-mcp.rb
@@ -45,6 +45,12 @@ class LogicProMcp < Formula
     pkgshare.install "Scripts/logic_bounce_ui.py" if (buildpath/"Scripts/logic_bounce_ui.py").exist?
     pkgshare.install "Scripts/logic_ui_jxa.py" if (buildpath/"Scripts/logic_ui_jxa.py").exist?
     pkgshare.install "Scripts/logic_input_source.py" if (buildpath/"Scripts/logic_input_source.py").exist?
+    # #427: shared dependency imported by logic_bounce.py / logic_bounce_ui.py /
+    # logic_ui_jxa.py. Omitting it shipped a package that failed at runtime with
+    # ModuleNotFoundError on the bounce/export path. The import-closure gate in
+    # release-verify-formula-install-paths.sh now fails the release if this (or any
+    # other transitively-imported logic_* helper) is not installed.
+    pkgshare.install "Scripts/logic_variants.py" if (buildpath/"Scripts/logic_variants.py").exist?
   end
 
   def caveats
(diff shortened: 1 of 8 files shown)


MongLong0214:

- Adversarial review round 1 found and we fixed pre-merge: a stale `pkgshare.count == 9` assertion (would have failed CI), the SetupDoctor share-dir completeness list missing the new module (runtime false-green), and a `set -euo pipefail` early-exit in the new gate when a helper has no local imports.
- Import-closure verified COMPLETE for the runtime helper set {logic_bounce, logic_bounce_ui, logic_ui_jxa, logic_input_source} → closure adds only logic_variants; dev-only helpers (session_bootstrap, free_tempo_modal, controller_learn_mode) are correctly not packaged (not server-spawned).
- Gate mutation-proven; wired in release.yml and exercised by the hermetic contract test. Contract suites 147/147; preflight exit 0; PR CI green.
- Reaches brew users only via the v3.12.0 release (new tarball + refreshed sha256) — in progress.

PR #430, opened 2026-07-21
ci(release): defer ADR-001 qualification gate behind ADR001_QUALIFICATION_ENFORCED flag

MongLong0214:
## Problem

The independent exact-artifact qualification gate in `.github/workflows/release.yml` was wired in ahead of the pipeline that satisfies it. The full same-artifact live matrix plus per-operation semantic coverage (#373) that would produce a passing signed evidence bundle is roadmap T5 / #284 and does not exist yet — `Scripts/live-qualification-runner.py` can only emit a `not_qualified` skeleton. As shipped, every release tag fails closed at the gate because no qualifying evidence can be produced. v3.12.0 is the first tag to hit it (the gate postdates v3.11.0), which is why its release run failed at "Enforce independent exact-artifact qualification".

## Change

Gate the qualification/verifier steps in both the `build` and `publish` jobs behind the repository variable `ADR001_QUALIFICATION_ENFORCED`:

- Unset/false (current): the gate is skipped and releases ship exactly as they did through v3.11.0.
- `'true'` (when #284 lands, with `QUALIFICATION_EVIDENCE_URL`, `QUALIFICATION_EVIDENCE_SHA256`, and `TRUSTED_QUALIFICATION_PUBLIC_KEY` provisioned): the enforcement runs fail-closed exactly as before.

The enforcement logic itself is unchanged. The SHA256 manifest step lists the `qualification-evidence/*` files only when present, so a deferred-gate release still produces a complete manifest of the shipped artifacts.

## Effect

Unblocks the v3.12.0 stable release, which carries the #427 `logic_variants.py` packaging fix. #284 re-enables the gate with the real qualification pipeline.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -166,17 +166,32 @@ jobs:
       # The independently produced bundle must cover Desktop/Creator × en-US/ko-KR,
       # every registered operation by semantic readback or a governed default-profile
       # waiver, and mutation/readback/restore/compensation. Missing evidence fails closed.
+      #
+      # ADR-001 terminal qualification is roadmap T5 / issue #284 and is not yet
+      # deliverable: the full same-artifact live matrix plus per-operation semantic
+      # coverage (#373) that would produce a passing signed evidence bundle does not
+      # exist yet (Scripts/live-qualification-runner.py can only emit a not_qualified
+      # skeleton). Until #284 lands, these enforcement steps are gated OFF by owner
+      # decision (2026-07-21) so stable releases ship exactly as they did through
+      # v3.11.0. To turn the gate back on: set repository variable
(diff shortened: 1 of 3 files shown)


MongLong0214:
Retrospective CEO re-ratification at current main 16f0900f5ee84d12f9108a33c73145fa269b6747: the ADR-001 enforcement deferral is RATIFIED only as temporary, explicit risk acceptance. It is not production-readiness approval. R-REL, R-MATRIX, R-MUT, R-PROV, R-PUB, and R-SEM remain OPEN. The v3.12.0 release run confirms qualification and provenance steps were skipped. Terminal #284 closure must remove the conditional workflow guards, provide the real same-artifact live matrix and trusted provenance evidence, and restore unconditional fail-closed enforcement.

PR #431, opened 2026-07-21
fix(export): surface input-source precondition and blocking-modal blocker in export readiness (#427 follow-up)

MongLong0214:
Two reviewer-confirmed export-path honesty fixes. Both make the export/bounce path declare or surface a precondition it previously left implicit, so a run can no longer look "ready" while a known blocker is present. Follow-up to #427; the first fix extends the execution-precondition work added in #426.

## Fix 1 — surface the input-source precondition (declarative)

**Root cause.** `executionPreconditions()` in `ProjectExportPlannerSupport.swift` advertised only three preconditions (`automation_permission`, `post_event_access`, `bounce_helper_available`). The keyboard-driven bounce/save panel automation (`Scripts/logic_bounce.py`) first switches the active keyboard input source to an ABC or US layout (`com.apple.keylayout.ABC` / `com.apple.keylayout.US`, via `Scripts/logic_input_source.py`) and **fails closed with `input_source_switch_failed`** when neither is enabled — a real user report on a Mac with only French/AZERTY active hit this after the project opened. The plan manifest was silent about this, so a plan could validate as ready and still fail at bounce time.

**Change.** Add a fourth `ProjectExportPrecondition` — `input_source_available` (`appliesToCommands: ["bounce"]`) — whose `detail` explains the ABC/US requirement, cites the exact layout identifiers and the `input_source_switch_failed` fail-closed behavior, points the user at System Settings > Keyboard > Input Sources, and states why failing closed is correct (keystrokes on a non-ABC/US layout would type the wrong filename). `verifyWith` matches the tone of the existing three entries.

**RED → GREEN** (`ProjectExportPlannerTests`, the existing test pinned the count at 3):

RED (before the code change):
```
✘ "#369: surfaces the bounce step's execution preconditions in the manifest"
  :79  (…map(\.requirement) → ["automation_permission","post_event_access","bounce_helper_available"])
       == [ … , "input_source_available"]   (± removed ["input_source_available"])
  :98  plan.executionPreconditions.first { $0.requirement == "input_source_available" } → nil
✘ "#369: execution preconditions encode with snake_case keys and survive round-trip"
  :124 (preconditions.count → 3) == 4
```
GREEN (after): `Test run with 7 tests passed`.

## Fix 2 — surface a blocking modal as a BLOCKER in export_readiness (false-green bug)

**Root cause.** `ProjectSessionAudit.exportReadiness(from:)` computes `blockers` as `findings.filter { $0.severity == .blocker }`. A blocking modal dialog present at audit time produced only the `ax_occluded` finding (severity `.warn`, keyed off `snapshot.axOccluded` — a weaker, different signal) and never a blocker. So `logic_project audit` reported `export_readiness: { blockers: [], status: "review_required" }` while the transport preflight correctly failed closed on the **same** dialog with `preflight_blocking_dialog`. A bounce cannot succeed while a modal owns the window — a false green.

**Change.** Reuse the authoritative detection — `AXLogicProElements.blockingDialogInfo()`, the same source the transport preflight consumes in `DispatcherSupport.blockingLogicDialogResult`:

- Add a plain-value field `blockingDialogButtons: [String]?` to the pure audit `Snapshot` (threaded like `axOccluded`; `nil` = no dialog).
- Emit a new `.blocker` finding `export_blocked_by_modal_dialog` (category `system`, resource `logic://system/health`) when it is non-nil, including the dialog button labels in the evidence values so it matches the transport path's honesty. It flows into `export_readiness.blockers` and drives report `status` to `failed`.
- Populate the field in `ProjectDispatcher`'s read-only `audit` case from `blockingDialogInfo()` (a new injectable seam defaulting to the same live source the transport path uses; tests inject a fixture).

The `ax_occluded` warning is unchanged; the modal blocker is a separate, stronger signal.

**RED → GREEN** (`ProjectSessionAuditTests`; the plumbing was added first so the tests compile, then the finding emission was added — that emission is the RED→GREEN evidence):

RED (plumbing present, finding not yet emitted):
```
✘ testProjectAuditSurfacesBlockingModalAsExportBlocker
  :371 report.findings.first { $0.id == "export_blocked_by_modal_dialog" } → nil
✘ testProjectAuditDispatcherSurfacesBlockingModalBlocker
  :426 (json["status"] as? String → "degraded") == "failed"
  :429 (exportReadiness["status"] as? String → "review_ready") == "blocked"
  :431 (blockers → []).contains("export_blocked_by_modal_dialog")
```
(control `testProjectAuditWithoutBlockingModalOmitsExportBlocker` passed in both states.)

GREEN (after emission added): all three pass; `Test run with 42 tests passed` for the audit filter.

The dispatcher test drives `ProjectDispatcher.handle(command: "audit", …)` with an injected blocking-dialog fixture and asserts the emitted audit JSON — it reproduces the live repro headlessly (no live Logic required).

## Test matrix

| Suite / test | Fix | RED | GREEN |
|---|---|---|---|
| `ProjectExportPlannerTests` "surfaces … execution preconditions" | 1 | list ≠ 4-entry, entry nil | pass |
| `ProjectExportPlannerTests` "encode with snake_case keys" | 1 | count 3 ≠ 4 | pass |
| `ProjectExportPlannerTests` (7 total) | 1 | 3 failing | 7/7 pass |
| `testProjectAuditSurfacesBlockingModalAsExportBlocker` (pure) | 2 | finding nil | pass |
| `testProjectAuditWithoutBlockingModalOmitsExportBlocker` (control) | 2 | pass | pass |
| `testProjectAuditDispatcherSurfacesBlockingModalBlocker` (dispatcher JSON) | 2 | status degraded / blockers [] | pass |
| Combined `ProjectExportPlanner|Audit|Dispatcher` | 1+2 | — | 320/320 pass |

Run: `swift test --no-parallel --disable-sandbox --filter …`. The dead-`#expect` CI lint (`scripts/ci-forbid-dead-expect.sh`) passes; every new assertion is load-bearing.

## Notes

- The dispatcher-level JSON test lives in `ProjectSessionAuditTests` (which already exercises `handle(command: "audit")` and parses the audit JSON) rather than in `ProjectAuditPhaseTests` (which is scoped to `[AUDIT]` log-phase capture, not JSON). Both `ProjectAuditPhaseTests` and `DispatcherTests` were run and stay green.
- `cleanup_plan` / `cleanup_apply` are intentionally out of scope: the bounce preflight and `cleanup_apply` already fail closed on a present dialog via the existing `dialogPresent()` gate; this change targets the read-only `audit` false-green specifically. The new `buildAudit` parameter defaults to `nil`, so those paths are unchanged.
- No live Logic was required to verify either fix; both are covered headlessly (pure `Snapshot → findings` for the audit, injected fixtures for the dispatcher).

Follow-up to #427. Relates to #426.

### DIFF
--- a/Sources/LogicProMCP/Dispatchers/ProjectDispatcher.swift
+++ b/Sources/LogicProMCP/Dispatchers/ProjectDispatcher.swift
@@ -35,6 +35,12 @@ struct ProjectDispatcher: OperationTraceDispatching {
         },
         sleep: (UInt64) async -> Void = { try? await Task.sleep(nanoseconds: $0) },
         dialogPresent: @escaping @Sendable () -> Bool = { false },
+        // #427 follow-up: the authoritative blocking-dialog signal the transport
+        // preflight fails closed on (`preflight_blocking_dialog`). Threaded into
+        // the read-only `audit` so a modal present at audit time surfaces as an
+        // export-readiness blocker instead of a false green. Defaults to the same
+        // live source the transport path uses; tests inject a fixture.
+        blockingDialogInfo: @escaping @Sendable () -> AXLogicProElements.BlockingDialogInfo? = { AXLogicProElements.blockingDialogInfo() },
         cleanupAuditFileReader: LogicProjectFileReader.Runtime = .production,
         // #27 Phase 2 — injectable export-execution seams (identity readback,
         // audio analysis, file polling). Defaults to the live wiring; tests
@@ -398,7 +404,10 @@ struct ProjectDispatcher: OperationTraceDispatching {
             return toolTextResult(result)
 
         case "audit":
-            let report = await ProjectSessionAudit.buildAudit(cache: cache)
+            let report = await ProjectSessionAudit.buildAudit(
+                cache: cache,
+                blockingDialogButtons: blockingDialogInfo().map(\.buttonTitles)
+            )
             do {
                 return toolTextResult(try encodeJSONStrict(report, compact: true))
             } catch {
--- a/Sources/LogicProMCP/Projects/ProjectExportPlannerSupport.swift
+++ b/Sources/LogicProMCP/Projects/ProjectExportPlannerSupport.swift
@@ -120,6 +120,18 @@ extension ProjectExportPlanner {
                     + "share path, pass the ownership-trust check, and have a python3 interpreter available.",
                 verifyWith: "LogicProMCP doctor"
             ),
+            ProjectExportPrecondition(
+                requirement: "input_source_available",
+                appliesToCommands: ["bounce"],
+                detail: "The export step types the destination filename into Logic's Bounce/save panel with "
+                    + "synthetic keystrokes, so the helper first switches the active keyboard input source to an "
+                    + "ABC or US layout (com.apple.keylayout.ABC or com.apple.keylayout.US); one of those must be "
+                    + "enabled in System Settings > Keyboard > Input Sources. If neither is enabled the helper FAILS "
+                    + "CLOSED with input_source_switch_failed instead of continuing (a Mac with only a French/AZERTY "
+                    + "source enabled hits this once the project has opened). Failing closed is correct because "
+                    + "keystroke automation on a non-ABC/US layout would type the wrong characters into the filename.",
+                verifyWith: "LogicProMCP doctor (bundled logic_input_source.py helper)"
+            ),
         ]
     }
 
--- a/Sources/LogicProMCP/Workflows/ProjectSessionAudit.swift
+++ b/Sources/LogicProMCP/Workflows/ProjectSessionAudit.swift
@@ -273,12 +273,21 @@ enum ProjectSessionAudit {
         /// `track_readback_gap` when the file says there are more tracks than AX
         /// surfaced — keeping the audit honest against `logic://tracks`.
         let fileTrackCount: Int?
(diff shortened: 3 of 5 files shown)


MongLong0214:
Retrospective CEO re-ratification at current main 16f0900f5ee84d12f9108a33c73145fa269b6747: QA_PENDING. Focused current-main audit/export/state suites pass, and the published v3.12.0 artifact is intact, but this Logic-facing behavior has no exact-artifact live Logic receipt or approved LIVE_NA. The original PR also has no recorded exact-head review verdict or post-merge completion receipt. Do not treat the historical merge or green CI as a completed production-readiness gate.

MongLong0214:
### Live verification receipt

Verified against a reproducible release binary built from `16f0900` (this fix is contained by that commit; merged here as `25ccc84`).

- Binary SHA-256 (independently recomputable): `7ec8c35e12f5361727d4bbf55554219c2a063d171ac739a064886a20809ba950`

**Deterministic** — targeted test run at `16f0900`, `Project export planner` suite passed, including the `input_source_available` precondition assertions.

**Live** — real Logic Pro, MCP over stdio. `logic_project.export_plan` on a real `.logicx` now reports the input-source precondition in the export-readiness manifest:

```
execution_precondition requirements:
  automation_permission, post_event_access, bounce_helper_available, input_source_available
input_source_available present: True
```

The `input_source_available` precondition is surfaced as intended, alongside the existing automation/post-event/bounce-helper preconditions.

PR #433, opened 2026-07-21
fix(qualification): credit semantic coverage from live qualification / governed waivers, not validator existence (#409)

MongLong0214:
## Root cause (#409)

`ProductionReadinessContractEvaluator.evaluate(...)` computed `missingSemantic` with a validator-existence exemption:

```swift
guard !semantic.contains(operationID) else { return false }
```

`semantic` is `semanticValidatorOperationIDs`, sourced from `operationsWithSemanticValidators()`, which returns non-nil for **every** oracle-table entry — including the mutating oracles added in #404 that the qualification drive never exercises with a live mutation (they have no live `.passed`). That line credited an operation as semantically covered merely because a validator object existed for it. As #404's oracles landed, credited coverage silently grew past the honest `1/108` baseline with no live qualification behind it — a false green in which R-SEM under-counted the missing operations.

The authoritative "qualified" signal is **live** (`PromotionGate`): a required operation is satisfied only through a case with `status == .passed && verificationKind == .semanticReadback`, or a governed release-visible waiver. The static repository-tree evaluator has **no** live `.passed` data (that matrix is #284 / T5, not run here), so the only thing it can honestly credit is a governed release-visible waiver.

## Fix

Drop the validator-existence exemption. An operation now counts as missing **unless** a qualifying governed waiver covers it (`governsOperation` + `affectsDefaultProfile` + `releaseNoteVisible` + valid reason code) — the governed-waiver check is unchanged. A short comment records why validator existence is not coverage.

- `operationsWithSemanticValidators()` and `semanticValidatorOperationCount` reporting are **unchanged** — they honestly report the validator inventory; only the coverage-credit logic (`missingSemantic`) changed.

## Why R-SEM stays open

`.github/qualification/waivers.json` is `[]` (the blanket waivers were removed earlier in this remediation; see `docs/tickets/lpmcp-prd-001/STATUS.md`). With no governed waivers and no live `.passed` data, every registered operation is now honestly missing: the R-SEM detail count rises from the previously-inflated value to `108/108`. R-SEM (`semanticCoverageIncomplete`) was already open and stays open; it closes only when the coverage program (#373) lands real live evidence.

## No other debt flips

The open-debt **set** is unchanged. `productionReadinessContractsAreSatisfiedOnCurrentTree` (which pins the exact set) still passes; only the R-SEM *detail count* becomes honest. The now-stale `1/108` doc note on that test is corrected to `0/108`.

## RED to GREEN evidence

[...]

Issue #409, opened 2026-07-18
R-SEM closer credits oracle-object existence as coverage without live qualification (false-green seam)


MongLong0214:
## R-SEM closer over-credits: oracle-object existence counts as semantic coverage without live qualification

**Where**: `Sources/LogicProMCP/Qualification/ProductionReadinessContracts.swift` — `missingSemantic` is computed via `operationsWithSemanticValidators()` (~L481-492, L527-544), which counts an operation as "covered" whenever its validator returns non-nil when probed with empty payloads.

**Problem**: that probe returns non-nil for **all** oracle-table entries — including the 12 Phase-B1 *mutating* oracles that the qualification drive can never exercise (mutating ops are driven with the no-write probe; `status` has no `.passed` arm for mutating; the oracle is never consulted against a real mutation — see #404). So contract-credited semantic coverage (~33/107) already includes operations with **zero live qualification**, and the R-SEM debt has no tie to the live PromotionGate result.

**Failure mode**: as oracle authoring proceeds (Phase B2/B3 add ~39 more oracle objects), `missingSemantic` shrinks toward 0 and R-SEM **closes by fiat** while the live run has never passed a single mutation — a false-green on ADR-001's core debt.

**Required fix**: the R-SEM closer must count a mutating operation as covered **only** when it reaches live `.passed` (real mutation + oracle + verified restore, mirroring PromotionGate's `operationPassed`) **or** carries a governed, release-visible waiver. Oracle-object existence must not credit coverage. (Read-only ops, which the drive genuinely exercises, are unaffected.)

Also stale: the "1/107" docstring near the pinned debt test no longer reflects the credited count.

Related: #404 (mutating-oracle qualification wiring), #373 (coverage program), #284 (ADR-001).

MongLong0214:
**Status update (2026-07-21):** scheduled as the next deterministic work item (milestone v3.13.0). Fix approach confirmed against the current tree: the R-SEM closer will count a mutating operation as covered only on a live `.passed` (real mutation + oracle + verified restore) or a governed release-visible waiver — oracle-object existence stops crediting coverage; adds a false-green mutation-proof and corrects the stale docstring near the pinned-debt test. Prerequisite for #373 Phase C-live accounting.

MongLong0214:
Fixed on `main` (`39794e0`, PR #433).

[...]

PR #435, opened 2026-07-21
fix(qualification): correct R-SEM missing-coverage detail wording

MongLong0214:
Post-#409, a registered semantic-readback validator no longer credits semantic coverage — only live qualification (a case with status `.passed`) or a governed release-visible waiver does. The R-SEM finding detail still read "semantic validators or release-gated governed waivers missing for N/M operations", which could mislead a reader into thinking adding a validator closes the gap.

Reword the prefix to "semantic coverage missing (no live qualification .passed and no governed release-visible waiver) for N/M operations". Purely a human-readable detail string; no behavioral change. The "N/M operations (e.g. …)" shape is unchanged, so the existing test assertions (`1/1 operations`, `system.health`) still hold — LPMCPPRD001 suite stays green (21 tests). Closes the guardian follow-up NIT from #433.

Refs #409, #433.

### DIFF
--- a/Sources/LogicProMCP/Qualification/ProductionReadinessContracts.swift
+++ b/Sources/LogicProMCP/Qualification/ProductionReadinessContracts.swift
@@ -544,7 +544,7 @@ enum ProductionReadinessContractEvaluator {
         if !missingSemantic.isEmpty {
             findings.append(.init(
                 id: .semanticCoverageIncomplete,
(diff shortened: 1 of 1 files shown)


MongLong0214:
Retrospective CEO re-ratification at current main 16f0900f5ee84d12f9108a33c73145fa269b6747: PASS for the wording-only correction. The current detail accurately states that live qualification passed evidence or a governed release-visible waiver is required; validator inventory alone is not semantic coverage.

PR #436, opened 2026-07-21
qualification: #373 Phase-B2 transport/navigate verified-write oracles

MongLong0214:
## Summary

Extends the #373 dual semantic-oracle census with verified-write (State A) oracles for the transport + navigate mutating surface — the B2 increment on top of B0 (#403 constraint model) and B1 (#404, 12 verified-write oracles). Each oracle was derived by reading the operation's dispatcher/channel handler and pinning what its `encodeStateA` envelope actually emits.

## Review fixes (adversarial review WARN)

**FIX 1 — stale banned `mouse-click` strategy removed.** The ADR-001 coordinate ban removed the HID mouse-click rung from `controlBarClickStrategies` (now `axpress`/`axconfirm` only; no handler emits `mouse-click`). The oracle domain `controlBarClickActions` still listed `"mouse-click"`, which would silently accept the banned strategy if a regression re-introduced it. Removed it (domain is now `["axpress", "axconfirm"]`) and updated the `toggle_cycle` / `toggle_count_in` / `toggle_metronome` fixtures to the real emitted values (`action:"axpress"`, `attempts:["axpress"]`).

**FIX 2 — `transport.toggle_metronome` oracle no longer false-REDs the keycmd fallback.** Metronome uniquely routes `[.accessibility, .midiKeyCommands, .cgEvent]` with dispatcher re-verification (`finalizeToggleMetronomeResult`). Its verified State A arrives in two shapes with disjoint keys: AX-verified (`observed`/`previous` flip + `button`/`control`/`action`) and dispatcher-verified (keycmd/cgEvent State B confirmed via a transport-state read: `observed_enabled`/`previous_enabled` flip + `verification_source`, and NO `button`/`control`/`action`). The old oracle pinned `button`/`control`/`action` (fail-closed on missing), so it rejected the genuine dispatcher-verified State A — a false-RED that would fire in the #284 matrix on an AX-control-absent, keycmd-bound setup. Relaxed the oracle to pin only the channel-independent verified envelope; the toggle-changed proof is structural (State A is unreachable in any path without a confirmed flip), exactly as `transport.stop` / `transport.pause`. A dedicated test proves the relaxed oracle accepts BOTH channel shapes yet rejects a no-op (State B) and a failure (State C).

RED to GREEN evidence for FIX 2:
- With the former `button`/`control`/`action` pins, the dispatcher-verified keycmd State A is REJECTED: `Expectation failed: keycmdVerified` — "metronome oracle false-RED the dispatcher-verified keycmd State A". After relaxation the same shape is accepted (GREEN).
- The no-op rejection is load-bearing: forging the no-op payload to a verified State-A envelope makes the oracle accept it (`Expectation failed: !(noOp ...)` — "accepted a no-op State B result"); the real no-op (honest State B, `verified:false`) is rejected. The envelope constraints are formally mutation-proofed by `everyConstraintRejectsItsMutants(.transportToggleMetronome)`.

## What B2 covers (15 oracles)

- transport: `play`, `stop`, `record`, `pause`, `toggle_cycle`, `toggle_metronome`, `toggle_count_in`, `toggle_autopunch`, `set_tempo`, `goto_position`
- navigate: `goto_bar`, `goto_marker`, `create_marker`, `rename_marker`, `set_zoom`

New relational constraint `booleanFlipped(keyA:keyB:)` expresses the verified-toggle invariant `observed == !previous` — a control-bar checkbox reaches State A only when the AX read-back is the negation of the pre-read, so this proves the toggle CHANGED state rather than a no-op that still reported success. It uses the same CFBoolean discipline as the rest of the model (a numeric `0`/`1` never satisfies it), is value-bearing (a toggle oracle survives the anti-checkbox strength gate), and carries a case in the generic mutation harness.

## Audited send-only exclusions

Send-only operations are excluded explicitly with a reason, not silently dropped:

- `transport.rewind`, `transport.fast_forward`, `navigate.zoom_to_fit`, `navigate.delete_marker`, `navigate.toggle_view` route only to send-only channels (MCU / MIDIKeyCommands / CGEvent) whose handlers emit State B `readback_unavailable` with no read-back. There is no State A to pin, so each is recorded in `structurallyUnverifiedMutatingOperationIDs` with its reason (census-asserted: real mutating spec, disjoint from covered, reason length checked).
- `transport.set_cycle_range` is registry `.unsupported`, so `mutatingSpecIDs` filters it a level earlier; documented + reasoned in `unsupportedExcludedMutatingOperationIDs`.

## Census soundness

- Read-only surface stays fully covered (20).
- Mutating increment pinned exactly: B1 (12) + B2 (15) = 27, and the two increments are disjoint.
- Every covered id is proven a real `.mutating`, non-`.unsupported` registry spec.
- `everyOracleHasAFixture`, `everyOracleMapsToASupportedSpec`, and the read-only non-regression invariant all hold.

## Dead-assertion audit (preflight C2 / test-integrity CI lint)

No dead boolean `#expect`/`#require` was present in the preserved checkpoint. Every added assertion uses a load-bearing form — `let v = try #require(oracle.evaluate(...)); #expect(v)` and `let survived: Bool = oracle.evaluate(...) == true; #expect(!survived)` (the `== true` collapses `Bool?` on its own line, not inside `#expect`). Both gates pass on this head:

- `Scripts/ci-forbid-dead-expect.sh` (span/closure-aware, scans all of `Tests/`): `OK: no dead boolean #expect/#require spellings`.
- public-surface preflight C2 (and C1/C3): `PREFLIGHT PASS`.

The four `#expect(dict[id] == nil)` / `!= nil` sites are `Optional<OperationOracle>` / `Optional<String>` presence checks — genuinely load-bearing, and not the flagged bool-literal form.

## RED to GREEN evidence (mutation-proof)

The B2 oracles are mutation-proofed by the parameterized harnesses (`oraclePassesItsRealisticFixture` = good-fixture leg; `everyConstraintRejectsItsMutants` = per-constraint wrong-value leg), which now iterate all 15 B2 oracles and pass. RED-before / GREEN-after was additionally confirmed empirically on two representative oracles:

1. `transport.set_tempo` — temporarily tightened the oracle's request/readback bound from `numericNear(..., .absolute(1.0))` to `0.01`. Its fixture (`observed 119.7`, `requested 120.0`, delta 0.3) then FAILED: `oraclePassesItsRealisticFixture(.transportSetTempo)` recorded "rejected its own realistic fixture", and `validatorRoutesEachOracleOperationToItsOracle(.transportSetTempo)` recorded "is not routed". Reverted to GREEN.
2. `transport.toggle_cycle` — temporarily corrupted the fixture to a no-op (`previous:true` so `observed == previous`, still reporting `success/verified/state:A`). The unchanged oracle REJECTED it via `booleanFlipped` (the exact "no-op reported success" hazard the constraint exists to catch): the same two tests went RED. Reverted to GREEN.

Both RED signals surfaced through live `#expect` (the `try #require`-unwrapped verdict and the `!survived` bool), confirming the assertions are load-bearing.

## Coverage-credit honesty (#409)

Per #409 (already on main), a mutating oracle definition is inventory, not coverage: R-SEM (`semanticCoverageIncomplete`) is credited only by live qualification `.passed` (the #284 live matrix) or a governed release-visible waiver — never by an oracle's existence. A comment on `phaseB2MutatingOperationIDs` records this: these B2 oracles pin the State-A contract for the future live run and do NOT themselves reduce the R-SEM static debt. The `ProductionReadinessContracts` coverage-credit logic is unchanged.

## Verification

- `swift test --no-parallel --disable-sandbox --filter SemanticOracle` -> 57 passed (adds the dedicated metronome channel-shape test)
- `--filter LPMCPPRD001ProductionReadinessREDTests` -> 21 passed
- `--filter QualificationRunner` -> 80 passed
- Rebased cleanly onto `origin/main` (no conflicts).

Refs #373, #404.

### DIFF
--- a/Sources/LogicProMCP/Qualification/SemanticOracleTable.swift
+++ b/Sources/LogicProMCP/Qualification/SemanticOracleTable.swift
@@ -97,6 +97,47 @@ enum SemanticOracleTable {
         .pluginsSetParamVerified,
     ]
 
+    /// The mutating operations B2 covers with a verified-write (State A) oracle:
+    /// the transport + navigate safe-mutation surface. Each was derived by READING
+    /// its dispatcher/channel handler and pinning what its `encodeStateA` envelope
+    /// actually emits (cited per oracle below). Like B1 this is an EXPLICIT
+    /// increment, not the whole mutating surface — B3 adds the rest.
+    ///
+    /// COVERAGE CREDIT (#409): defining an oracle here is INVENTORY, not coverage.
+    /// R-SEM (`semanticCoverageIncomplete`) is credited to an operation ONLY when it
+    /// is live-exercised to a `.passed` semanticReadback verdict (the #284 live
+    /// matrix) or covered by a governed release-visible waiver — never by an oracle's
+    /// mere existence. So this B2 increment PINS the State-A contract these
+    /// transport/navigate writes must satisfy in the future live run; it does NOT
+    /// itself reduce the R-SEM static debt.
+    ///
+    /// DELIBERATELY ABSENT (audited, no State-A envelope to verify — see
+    /// `structurallyUnverifiedMutatingOperationIDs`): transport.rewind,
+    /// transport.fast_forward, navigate.zoom_to_fit, navigate.delete_marker,
+    /// navigate.toggle_view. Every one routes ONLY to send-only channels
+    /// (MCU / MIDIKeyCommands / CGEvent) whose handlers emit State B
+    /// `readback_unavailable` with no read-back — there is no State A to pin, so a
+    /// verified-write oracle would have to invent a contract the handler can never
+    /// produce. transport.set_cycle_range is excluded a level earlier: it is
+    /// registry `.unsupported`, so `mutatingSpecIDs` already filters it out (see
+    /// `unsupportedExcludedMutatingOperationIDs`).
+    static let phaseB2MutatingOperationIDs: Set<OperationID> = [
+        // transport (10 of the 12 audited — rewind/fast_forward are send-only)
+        .transportPlay, .transportStop, .transportRecord, .transportPause,
+        .transportToggleCycle, .transportToggleMetronome, .transportToggleCountIn,
+        .transportToggleAutopunch, .transportSetTempo, .transportGotoPosition,
+        // navigate (5 of the 8 audited — zoom_to_fit/delete_marker/toggle_view are send-only)
+        .navigateGotoBar, .navigateGotoMarker, .navigateCreateMarker,
+        .navigateRenameMarker, .navigateSetZoom,
+    ]
+
+    /// Every mutating operation the table covers so far: the B1 verified-write
+    /// increment UNION the B2 transport/navigate increment. Coverage is still
+    /// INCREMENTAL (B3 remains); this is the covered subset the census pins.
+    static var coveredMutatingOperationIDs: Set<OperationID> {
+        phaseB1MutatingOperationIDs.union(phaseB2MutatingOperationIDs)
+    }
+
     /// The mutating surface (non-`.unsupported`), from the registry. Coverage is
     /// INCREMENTAL — unlike `coveredSpecIDs`, the table need not cover all of this
     /// yet; `phaseB1MutatingOperationIDs` is the covered subset.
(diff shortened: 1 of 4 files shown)


MongLong0214:
Retrospective CEO re-ratification at current main 16f0900f5ee84d12f9108a33c73145fa269b6747: substantive pure-code inventory PASS. Current-main semantic-oracle and production-readiness focused suites pass. These oracles remain inventory only and do not reduce R-SEM without live passed evidence or a governed waiver. Original exact-head review and completion-receipt records are absent and are not reconstructed by this current-main verdict.

PR #437, opened 2026-07-21
fix(audit): surface blocking-modal blocker on the logic://project/audit resource via cached signal (#432)

### LINKED ISSUE #432: export_readiness: logic://project/audit resource does not surface the blocking-modal blocker (tool path fixed; cache-only resource lags)
## Summary

Follow-up to #427 / #431. The `logic_project audit` **tool** now surfaces a blocking modal as an `export_blocked_by_modal_dialog` blocker in `export_readiness` (via the authoritative `AXLogicProElements.blockingDialogInfo()` signal — the same one the transport preflight fails closed on). The **resource** `logic://project/audit` renders the identical `AuditReport` but is built by `ProjectSessionAudit.buildAudit(cache:)` **without** that signal, so it still reports `export_readiness: { blockers: [] }` while a blocking modal owns the Logic window.

Result: the tool and the resource **disagree** for the same live state — the resource retains the exact false-green class #431 eliminated on the tool.

## Where

- `Sources/LogicProMCP/Resources/ResourceHandlers+StateReaders.swift` — `readProjectAudit(cache:)` calls `ProjectSessionAudit.buildAudit(cache:)` (cache-only projection; no live AX).
- Registered + reachable: `ResourceHandlers.swift` (`logic://project/audit`), `ResourceProvider.swift`.
- Contrast: `ProjectDispatcher` `audit` case threads `blockingDialogInfo()` into `buildAudit`; the resource does not.

## Severity / scope

Not a regression — the resource never surfaced the modal blocker before #431. Write-paths (`bounce`, `export_run`, `cleanup_apply`, …) still fail closed on a present dialog via the existing `dialogPresent()` gate regardless of which read surface is consulted, so no unsafe mutation can proceed. This is a read-surface **honesty/consistency** gap, not a safety hole.

## Proper fix (design-consistent)

Resources are intentionally cache-only projections; a live AX call inside a resource read would break that boundary. The consistent fix is to capture the authoritative blocking-dialog signal (button titles) in `StateCache` when the state poller runs, and have `buildAudit(cache:)` read it — so **both** the tool and the resource surface the blocker from a single cached source, and neither needs an ad-hoc live call at read time.

Relates to #427, #426, #431.

MongLong0214:
## Root cause

Follow-up to #427 / #431. The `logic_project audit` **tool** surfaces a blocking
modal as an `export_blocked_by_modal_dialog` blocker in `export_readiness` (#431,
via the live `AXLogicProElements.blockingDialogInfo()` signal — the same one the
transport preflight fails closed on). The sibling **resource**
`logic://project/audit` renders the identical `AuditReport` via the shared
`buildAudit(cache:)` builder, but it was called **without** that signal, so it kept
reporting `export_readiness: { blockers: [] }` while a blocking modal owned the
Logic window.

Result: the tool and the resource **disagreed** for the same live state — the
resource retained the exact false-green class #431 eliminated on the tool. Not a
safety hole (write paths such as `bounce` still fail closed on the live
`dialogPresent()` gate), but a read-surface honesty/consistency gap.

## Fix — cache-fed signal (resources stay cache-only projections)

A live AX call inside a resource read would break the cache-only-projection
boundary. Instead, capture the authoritative blocking-dialog signal in the cache
when the state poller runs, and read it at the resource boundary:

- **`StateCache`**: add `blockingDialogButtons: [String]?` plus actor-isolated
  `getBlockingDialogButtons()` / `updateBlockingDialogButtons(_:)`, mirroring how
  `axOccluded` / `getAXOccluded` / `updateAXOccluded` are already stored and
  retrieved.
- **`StatePoller`**: add a `blockingDialogInfo` runtime seam. Production wires it
  to `AXLogicProElements.blockingDialogInfo()` (post-classifier — plugin-editor,
  Smart-Controls, and keyboard-layout overlay windows are excluded), the **same**
  signal the transport preflight and the #431 tool audit use. It is sampled once
  per visible-window poll and cached as button titles (or `nil`); cleared when the
  window/document goes away. No new detection was invented.
- **`readProjectAudit`**: sources the existing `buildAudit` `blockingDialogButtons`
  parameter from `cache.getBlockingDialogButtons()`.

The tool audit (live seam) and the bounce preflight are **unchanged**, so neither
regresses. Both the tool and the resource now surface the blocker for the same
blocking-modal state; `nil` ⇒ no finding, non-`nil` ⇒ the `.blocker` flows into
`export_readiness.blockers` and status → `failed`, for both surfaces. The finding
id/severity and `export_readiness` flow are reused as-is from #431.

## RED → GREEN (load-bearing resource test)

`testProjectAuditResourceSurfacesBlockingModalBlockerFromCache` reads
`logic://project/audit` with a blocking dialog cached.

- **Before the fix (RED):** `export_readiness.blockers → []`, `status → review_ready`
  / report `degraded` — the false green.
- **After the fix (GREEN):** `export_readiness.blockers` contains
  `export_blocked_by_modal_dialog`, `status → blocked`, report `failed`.

### Tests added
- Resource-path: cache carrying a blocking dialog ⇒ resource surfaces the blocker;
  control proves absence when the cache is clear.
- Tool + resource agreement for the same cached/live blocking-modal state.
- `StateCache` round-trip pin for the cached signal.
- `StatePoller` pins: caches the signal from the runtime seam; clears it when
  absent.

All assertions use `try #require` + concrete comparisons or force-unwrap (no dead
`#expect` bool-optional forms).

## Verification
- `swift test --no-parallel --disable-sandbox` filtered to the audit, resource /
  state-reader, `StateCache`, `StatePoller`, and dispatcher suites: **433 tests
  passed**. The 8 blocking-modal / cache / poller tests pass; the load-bearing
  resource test fails before the fix and passes after.
- Public-surface preflight: **PASS**.

Closes #432. Refs #431, #427.

### DIFF
--- a/Sources/LogicProMCP/Resources/ResourceHandlers+StateReaders.swift
+++ b/Sources/LogicProMCP/Resources/ResourceHandlers+StateReaders.swift
@@ -574,7 +574,16 @@ extension ResourceHandlers {
     }
 
     static func readProjectAudit(cache: StateCache, uri: String) async throws -> ReadResource.Result {
-        let report = await ProjectSessionAudit.buildAudit(cache: cache)
+        // #432: source the blocking-dialog signal from the cache (populated every
+        // poll by StatePoller from AXLogicProElements.blockingDialogInfo()) rather
+        // than making a live AX call here — resources are cache-only projections.
+        // This is the SAME authoritative signal the #431 tool audit passes live, so
+        // the resource now surfaces `export_blocked_by_modal_dialog` instead of a
+        // false green and agrees with the tool for the same blocking-modal state.
+        let report = await ProjectSessionAudit.buildAudit(
+            cache: cache,
+            blockingDialogButtons: await cache.getBlockingDialogButtons()
+        )
         // Honest Contract: never emit a success-shaped body that is missing the
         // schema/read_only contract fields. On encode failure, fail loud.
         do {
--- a/Sources/LogicProMCP/State/StateCache.swift
+++ b/Sources/LogicProMCP/State/StateCache.swift
@@ -29,6 +29,19 @@ actor StateCache {
     /// than silently returning prior values.
     private(set) var axOccluded: Bool = false
 
+    /// #432 — button titles of a blocking modal dialog/sheet that owns the Logic
+    /// window at the last poll, sampled from the SAME authoritative signal the
+    /// transport preflight and the #431 tool audit fail closed on
+    /// (`AXLogicProElements.blockingDialogInfo()`; post-classifier, so plugin
+    /// editors / Smart Controls / keyboard-layout overlays are excluded). `nil`
+    /// when no blocking dialog is present. Written every poll by
+    /// `StatePoller.pollOnce`, exactly like `axOccluded`. The `logic://project/audit`
+    /// resource — a cache-only projection that must not make live AX calls at read
+    /// time — reads this via `getBlockingDialogButtons()` so it surfaces the
+    /// `export_blocked_by_modal_dialog` blocker instead of a false green, matching
+    /// the live tool path (#431).
+    private(set) var blockingDialogButtons: [String]? = nil
+
     /// Timestamp of last tool call — drives adaptive poll intervals.
     private(set) var lastToolAccess: Date = .distantPast
 
@@ -105,6 +118,11 @@ actor StateCache {
     /// poll or when the document is genuinely closed.
     func getAXOccluded() -> Bool { axOccluded }
 
+    /// #432 — current blocking-dialog button titles (`nil` when none). See the
+    /// field comment for provenance; written every poll by `StatePoller.pollOnce`
+    /// via `updateBlockingDialogButtons`, mirroring `getAXOccluded`/`updateAXOccluded`.
+    func getBlockingDialogButtons() -> [String]? { blockingDialogButtons }
+
     /// Atomic, single-hop read of every field the project audit consumes.
     /// `buildAudit` previously assembled its snapshot from ~13 separate `await`
     /// calls; each was individually actor-serialized but the sequence was not a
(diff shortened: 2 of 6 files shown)


MongLong0214:
Retrospective CEO re-ratification at current main 16f0900f5ee84d12f9108a33c73145fa269b6747: QA_PENDING. StateCache, StatePoller, audit, and resource focused suites pass, but the live Logic state-poller to cached resource projection has no exact-artifact live receipt or approved LIVE_NA. The historical merge and green CI are insufficient to close that lane.

MongLong0214:
### Live verification receipt

Verified against a reproducible release binary built from `16f0900` (this fix is contained by that commit; merged here as `302711b`).

- Binary SHA-256 (independently recomputable): `7ec8c35e12f5361727d4bbf55554219c2a063d171ac739a064886a20809ba950`

**Deterministic** — targeted test run at `16f0900`, all six blocking-modal audit tests passed:

- `testProjectAuditResourceSurfacesBlockingModalBlockerFromCache` — the `logic://project/audit` resource surfaces the `export_blocked_by_modal_dialog` blocker from the cached poller signal (the fix; this path was previously false-green).
- `testProjectAuditToolAndResourceAgreeOnCachedBlockingModal` — the tool audit and the resource agree on the cached blocking-modal state.
- `testProjectAuditDispatcherSurfacesBlockingModalBlocker`
- `testProjectAuditSurfacesBlockingModalAsExportBlocker`
- `testProjectAuditWithoutBlockingModalOmitsExportBlocker`
- `testProjectAuditResourceOmitsBlockingModalBlockerWhenCacheClear`

**Live** — baseline (no modal) resource and tool audits return cleanly against real Logic Pro over stdio.

**Honest disclosure on live modal recapture:** a live end-to-end recapture with an actually-open blocking modal is environment-blocked in this setup. An open `File > Open` modal wedges Logic's live accessibility subtree — which is exactly the deep AX read the *tool* audit performs — so the tool-audit request cannot complete while the modal is up. This is an OS-level accessibility limitation of driving a modal from the same process, not a defect. It is precisely why the fix routes through the **resource** path: the resource surfaces the blocker via the cached poller signal (a cache read, no live AX), which is proven deterministically above.

PR #438, opened 2026-07-21
qualification: #373 Phase-B3 project/tracks/system/midi verified-write oracles

MongLong0214:
Extends the semantic-oracle table (`SemanticOracleTable`) to the remaining
verified-write mutating surface in the **project / tracks / system / midi**
tools — the increment after B1 (12) + B2 (15). Every oracle was derived by
**reading the operation's dispatcher/channel handler** and pinning exactly what
its verified path emits (cited per oracle), so each rejects the semantically
wrong result it exists to catch, never an invented constraint.

## B3 covers (15)

**project (3)**
- `project.save` — AppleScript `verifiedSaveResult` (chain `[.appleScript, …]`; fallbacks are send-only State B). Pins `operation`/`method`/`verify_source:"file_mtime"` + the `document_path`/`observed_mtime` that a real mtime-advance State A always carries.
- `project.save_as` — **envelope-only** (the transport.stop / toggle_metronome precedent). Its `[.accessibility, .appleScript]` chain reaches State A in two shapes with disjoint non-envelope keys (AX-dialog `requested`/`observed`/`via` vs AppleScript `operation`/`method`/`path`/`observed`), sharing no pinnable constant; the "file written" proof is the structural State-A gate both channels reach only after confirming the package on disk. A dedicated test proves it accepts **both** shapes yet rejects State B/C.
- `project.bounce` — accessibility `openBounceDialogViaMenu`. State A = the Bounce **dialog opened**, not an artifact produced, so `dialog_opened:true` + `bounce_fired:false` are pinned (the honest milestone).

**tracks (6)**
- `create_audio` / `create_instrument` / `create_drummer` / `create_external_midi` — `verifyTrackCreation`: `requested_delta:1` + `verification_source:"track_count_delta"` + `observed_delta ≥ 1` (a track really appeared) + the per-op `observed_track_type` discriminator (`audio` / `software_instrument` / `drummer` / `external_midi`) so each rejects the wrong track type.
- `delete` — `defaultDeleteTrack`: `requested_delta:-1` + `observed_delta ≤ -1` (the count really decreased).
- `record_sequence` — **bespoke** oracle: its verified payload is hand-assembled via `jsonToolTextResult` and carries `success:true`+`verified:true` **without** a `state` field (it does not call `encodeStateA`), so `state:"A"` is deliberately NOT pinned. Pins the region-readback match `start_bar == expected_start_bar (1)` and `end_bar == expected_end_bar`, `target_track_index == created_track`, `method:"smf_import"`.

**system (4)**
- `saga_execute` — `SagaWire.storedOutcome` reaches `encodeStateA` only when `saga_state == "completed"`, pinned + the session-journal markers. `duplicate` is NOT pinned (a replayed completion is still State A with `duplicate:true`).
- `saga_cancel` — State A only on a **verified** cancellation; pins `status:"cancelled"` + session markers.
- `setup_arm_key` — both State-A outcomes (verify-first / GUI-assigned) are gated on a functional arm-flip; pins the constant `command:"Toggle Track Record Enable"` (branch extras differ).
- `export_support_bundle` — pins a **non-empty** `files` manifest + `bundle_path` (empty manifest is not a successful export; the containment re-check is the structural gate).

**midi (2)**
- `import_file` — `defaultImportMIDIFile`: `via:"ax_menu_import"`, `observed_delta ≥ 1`, `imported_region_count ≥ 1` + non-empty `imported_regions` (the imported content is confirmed).
- `mmc_locate` — its `bar` sub-path routes `transport.goto_position` and runs the same `finalizeGotoPositionResult`, so its verified State A is the goto shape (`verification_source:"transport_state"`, `requested == observed`). The SMPTE `time` sub-path is send-only State B, correctly rejected by the envelope. (Registry `VerificationPolicy.bestEffort`, dual-path — not send-only, hence covered rather than excluded.)

## Audited none-verify / send-only exclusions (19 new)

Added explicit, reasoned entries to `structurallyUnverifiedMutatingOperationIDs` — the uncovered complement is kept explicit, never implicit:

- **project.new / project.open** — AppleScript lifecycle; the successful path is wrapped by `wrapMutatingResult` as honest **State B** `readback_unavailable` (the front-document-identity change only gates State C-vs-B, never A).
- **project.close** — `VerificationPolicy.none`; same State-B ceiling.
- **project.launch / project.quit** — `runLifecycleScript` returns a **prose** success string gated on a running-state poll, not an HC State-A envelope.
- **tracks.duplicate** — `VerificationPolicy.none`; routes `[.midiKeyCommands, .cgEvent]` (no accessibility), both send-only State B.
- **midi send-only** — `send_note/chord/cc/program_change/pitch_bend/aftertouch/sysex`, `play_sequence`, `step_input`, `create_virtual_port`, `mmc_play/stop/record` — all `VerificationPolicy.none`, route to send-only channels emitting State B `send_only_no_readback`; no read-back, no State A.

## Census soundness

`coveredMutatingOperationIDs` is now **B1 ∪ B2 ∪ B3 = 42 ops**. The census asserts the three increments are **pairwise-disjoint** and that every B3 id is a real `.mutating`, non-`.unsupported` registry spec (a read-only id or typo cannot masquerade as mutating coverage; the read-only 20 stay fully covered). Mutating coverage stays **incremental** (42 of the mutating surface) — the edit surface, the mixer/plugin inserts, and the export/cleanup executors remain future work. Meta-tests updated: exact-increment (B1=12, B2=15, B3=15, total 42), additive total (`all.count == 20 + 12 + 15 + 15`), and the structural-exclusion allowlist (B3 representatives asserted present).

## RED → GREEN evidence

All B3 oracles run through the existing parameterized harnesses: `oraclePassesItsRealisticFixture` (good-leg — each faithful fixture passes) and `everyConstraintRejectsItsMutants` (bad-leg — each constraint's mutant sinks the oracle). Plus dedicated tests:
- `saveAsOracleAcceptsBothChannelShapesYetRejectsUnverified` — both channel shapes pass; State B and State C rejected.
- `recordSequenceOracleRejectsWrongRegionAndUnverifiedImport` — wrong end-bar region and a `timing_mismatch` (verified:false) rejected.
- `phaseB3RepresentativeOraclesRejectSemanticMutations` — a value-bearing (`saga_execute` non-completed), count (`delete` zero-delta; `create_audio` wrong type), and structural (`export_support_bundle` empty manifest) rejection.

Suites green (foreground, `--no-parallel --disable-sandbox`): `SemanticOracle` 60, `LPMCPPRD001ProductionReadinessREDTests` 21, `QualificationRunner` 80. `Scripts/ci-forbid-dead-expect.sh` OK.

## #409 inventory note

`phaseB3MutatingOperationIDs` carries a `COVERAGE CREDIT (#409)` comment: these oracles are **inventory** pinning the State-A contract for the #284 live run; they do **not** reduce the R-SEM static debt and do **not** change `ProductionReadinessContracts` coverage-credit logic.

Refs #373, #436. Please do not merge yet.

### DIFF
--- a/Sources/LogicProMCP/Qualification/SemanticOracleTable.swift
+++ b/Sources/LogicProMCP/Qualification/SemanticOracleTable.swift
@@ -131,11 +131,58 @@ enum SemanticOracleTable {
         .navigateRenameMarker, .navigateSetZoom,
     ]
 
+    /// The mutating operations B3 covers with a verified-write (State A) oracle:
+    /// the project-lifecycle / tracks-topology / system-saga / midi-import
+    /// surface — the remaining verified-write mutating ops in the project, tracks,
+    /// system, and midi tools that reach a genuine State A. Each was DERIVED BY
+    /// READING its dispatcher/channel handler and pinning what the verified path
+    /// actually emits (cited per oracle below). Like B1/B2 this is an EXPLICIT
+    /// increment, not the whole mutating surface — the edit surface, the
+    /// mixer/plugin inserts, and the export/cleanup executors remain future work.
+    ///
+    /// COVERAGE CREDIT (#409): defining an oracle here is INVENTORY, not coverage.
+    /// R-SEM (`semanticCoverageIncomplete`) is credited to an operation ONLY when it
+    /// is live-exercised to a `.passed` semanticReadback verdict (the #284 live
+    /// matrix) or covered by a governed release-visible waiver — never by an oracle's
+    /// mere existence. So this B3 increment PINS the State-A contract these
+    /// verified writes must satisfy in the future live run; it does NOT itself
+    /// reduce the R-SEM static debt, and it does not change
+    /// `ProductionReadinessContracts` coverage-credit logic.
+    ///
+    /// DELIBERATELY ABSENT (audited, no verifiable State A — see
+    /// `structurallyUnverifiedMutatingOperationIDs`): project.new / project.open
+    /// (AppleScript lifecycle → honest State B readback_unavailable, never A),
+    /// project.close (VerificationPolicy.none, same State-B ceiling), project.launch
+    /// / project.quit (a prose success string gated on a running-state poll, not an
+    /// HC envelope), tracks.duplicate (VerificationPolicy.none, send-only), and the
+    /// send-only midi surface (send_*/mmc_play/mmc_stop/mmc_record/play_sequence/
+    /// step_input/create_virtual_port — every one routes to a send-only channel that
+    /// emits State B send_only_no_readback with no read-back).
+    static let phaseB3MutatingOperationIDs: Set<OperationID> = [
+        // project (3 of the 8 lifecycle ops audited — new/open/close/launch/quit
+        // reach no State A; save_as is envelope-only, see its oracle)
+        .projectSave, .projectSaveAs, .projectBounce,
+        // tracks (6 — the 4 create_* + delete + record_sequence topology writes)
+        .tracksCreateAudio, .tracksCreateInstrument, .tracksCreateDrummer,
+        .tracksCreateExternalMIDI, .tracksDelete, .tracksRecordSequence,
+        // system (4 — the saga executors + the consent-gated arm-key setup + the
+        // support-bundle write)
+        .systemSagaExecute, .systemSagaCancel, .systemSetupArmKey,
+        .systemExportSupportBundle,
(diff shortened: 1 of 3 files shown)


MongLong0214:
Retrospective CEO re-ratification at current main 16f0900f5ee84d12f9108a33c73145fa269b6747: substantive pure-code inventory PASS. Current-main semantic-oracle and production-readiness focused suites pass. This remains State-A contract inventory, not live semantic-coverage credit. Original exact-head review and completion-receipt records are absent.

PR #439, opened 2026-07-21
qualification: #373 Phase-B4 verified-execution oracles (plugin-insert/export/cleanup) — closes mutating inventory

MongLong0214:
## Summary

The FINAL #373 mutating-oracle increment after B1(12)+B2(15)+B3(15)=42. B4 adds
**6** oracles and **13** audited structural exclusions, which **closes the
pure-code (non-live) mutating oracle inventory**: every supported
(non-`.unsupported`) mutating spec is now either covered by an oracle or carries a
reasoned exclusion — no mutating op is left implicitly uncovered.

Each oracle was derived by **reading the handler** and pinning EXACTLY what its
State-A path emits — the real semantic invariant, never invented/over-loose — and
each rejects the semantically-wrong result it exists to catch.

## B4 covered (6)

| op | State-A source | pinned invariant |
|----|----------------|------------------|
| `plugins.insert_verified` | `defaultInsertVerified` → `encodeV2StateA` | `hc_schema:2` + `operation` + `verify_source:"ax_plugin_inventory"` + **request↔readback** `fieldsEqual(target_identity.plugin_id, observed_plugin_id)` and `fieldsEqual(target_identity.insert, observed_slot)` (RIGHT plugin in the RIGHT slot) |
| `mixer.insert_plugin` | `defaultInsertPlugin` → `encodeStateA` (gated on `spec.matches(observed)`) | `verify_source:"ax_plugin_slot"` + `plugin_name` stock allowlist {Gain, Compressor, Channel EQ} + observed/track/slot shape |
| `project.export_run` | `ProjectExportExecutor.aggregate` run RECORD (**bespoke** — no `state:"A"` envelope) | `schema:"logic_pro_mcp_export_run.v1"` + `mode:"run"` + `confirmed:true` + `status:"completed"` + `artifacts_failed:0` + `artifacts_uncertain:0` (independent guards) + non-vacuity (`artifacts_total>=1`, non-empty projects) |
| `project.export_resume` | same record, `export_resume` | export_run's twin with `mode:"resume"` (a completed resume may legitimately be all-skipped, so `artifacts_verified` is not floored) |
| `project.cleanup_apply` | `handleCleanupApply` → `encodeStateA` (inherits the `track.rename` readback) | `op:"project.cleanup_apply"` (key is `op`, not `operation`) + `tool:"logic_tracks"` + `command:"rename"` + `verify_source:"track.rename ax_readback"` + non-empty `applied`/`target_track_indices` |
| `edit.toggle_step_input` | `defaultToggleStepInputKeyboard` → `encodeStateA` | `operation` + `via:"window-menu"` + `booleanFlipped(observed_open, previous_open)` (the window open-state actually flipped) |

**Task-list correction (honest trace):** the B3 PR listed `edit.*(14)` as remaining.
Tracing each: `edit.toggle_step_input` routes `[.accessibility, …]` and reaches a
**genuine State A** (AX window-list read-back confirms the flip), so it is **covered**,
not excluded. The other 13 edit ops are send-only.

## B4 excluded (13, audited)

The send-only edit ops — `edit.undo / redo / cut / copy / paste / delete /
select_all / split / join / quantize / bounce_in_place / normalize / duplicate`.
Each routes ONLY to `[.midiKeyCommands, .cgEvent]` (a blind CC key command /
CGEvent keystroke); both channels emit State B `send_only_no_readback` /
`readback_unavailable` with no edit read-back, so **no State A exists**. Each is
added to `structurallyUnverifiedMutatingOperationIDs` with a per-op reason citing
the handler. (`edit.toggle_step_input` is deliberately NOT in the exclusion map.)

`.unsupported` specs (e.g. `transport.set_cycle_range`) need no exclusion entry —
`mutatingSpecIDs` already filters them out.

## Ambiguity / latent-false-green watch

- `mixer.insert_plugin`: `spec.matches(observed)` is a case/whitespace-normalized
  compare, so `observed_plugin_name` may not be **byte-equal** to `plugin_name` on a
  legitimate State A → deliberately **NOT** a `fieldsEqual` (would risk false-RED);
  the match itself is the structural State-A gate. Pinned the `plugin_name`
  allowlist + `verify_source` instead.
- `project.export_run/resume`: pinned `status:"completed"` (rejects `partial` /
  `failed` / `confirmation_required`) rather than accepting `partial`. A partial run
  is an honest partial, not a verified execution. `artifacts_failed:0` +
  `artifacts_uncertain:0` are pinned as **independent** guards so a run that
  mislabels itself `completed` while carrying a failed artifact is still caught.
- `plugins.insert_verified`: `observed_plugin_id`/`observed_slot` are the
  post-insert readback; `target_identity.*` is the request — different sources, so
  the two `fieldsEqual` are genuine request↔readback pins, not same-source echoes.

## RED → GREEN evidence

`phaseB4RepresentativeOraclesRejectSemanticMutations` (bad-legs) + the parameterized
good-leg/bad-leg harnesses (`oraclePassesItsRealisticFixture`,
`everyConstraintRejectsItsMutants`) cover every B4 oracle. Explicit RED was captured
by temporarily removing the load-bearing pins and rebuilding:

- **value-bearing** (`plugins.insert_verified`): with the `fieldsEqual` pins removed,
  the oracle ACCEPTED a State A whose `observed_plugin_id` (Compressor) differed from
  the requested `plugin_id` (Gain), and one at the wrong slot → test RED. Restored → GREEN.
- **bespoke** (`project.export_run`): with `status`/`artifacts_failed` removed, the
  oracle ACCEPTED a `partial` run and a run mislabelled `completed` with a failed
  artifact → test RED. Restored → GREEN.

No dead `#expect` (`Scripts/ci-forbid-dead-expect.sh` → OK).

## Census / suite counts

- `phaseB4MutatingOperationIDs` (6); pairwise-disjoint from B1/B2/B3; covered
  mutating total **42 → 48**; read-only 20 untouched; table total = 20 + 12 + 15 +
  15 + 6.
- New **`everyMutatingSpecIsCoveredOrAuditedExcluded`**: covered (48) ∪ audited
  exclusions (38) == the entire mutating surface (86 non-`.unsupported`), complement
  empty — the inventory is closed.
- `COVERAGE CREDIT (#409)` note on the B4 set: inventory pinning the State-A
  contracts for the #284 live run; does **not** reduce R-SEM static debt, does not
  change `ProductionReadinessContracts`.
- B3-review NIT fixed: `everyCoveredMutatingIDIsARealMutatingSpec` doc-comment now
  says B1..B4.

Suites (foreground, `--no-parallel --disable-sandbox`): `SemanticOracle` **62
passed**; `LPMCPPRD001ProductionReadinessREDTests` **21 passed** (R-SEM debt stays
open, unchanged); `QualificationRunner` **80 passed** (2 skipped — require a release
binary; unrelated to this change).

## What mutating ops remain uncovered, and why

Nothing is left **implicitly** uncovered. The 38 structurally-excluded mutating ops
are permanently unverifiable by construction, each with an audited reason:
- **send-only MIDI** (send_note/chord/cc/program_change/pitch_bend/aftertouch/sysex,
  play_sequence, step_input, create_virtual_port, mmc_play/stop/record) → CoreMIDI State B;
- **send-only transport/navigate** (rewind, fast_forward, zoom_to_fit, delete_marker,
  toggle_view) → MCU/keycmd State B;
- **project lifecycle** (new/open/close State-B ceiling; launch/quit prose);
- **send-only** `tracks.duplicate`, `mixer.set_plugin_param` (Scripter);
- **the 13 send-only edit ops** (this PR).

`transport.set_cycle_range` is `.unsupported` (filtered a level earlier). Promoting
any excluded op would be a deliberate move from the exclusion map into an increment
once/if it gains a State-A path.

Refs #373, #438. **Do not merge** — inventory review only.

### DIFF
--- a/Sources/LogicProMCP/Qualification/SemanticOracleTable.swift
+++ b/Sources/LogicProMCP/Qualification/SemanticOracleTable.swift
@@ -174,15 +174,58 @@ enum SemanticOracleTable {
         .midiImportFile, .midiMMCLocate,
     ]
 
-    /// Every mutating operation the table covers so far: the B1 verified-write
-    /// increment UNION the B2 transport/navigate increment UNION the B3
-    /// project/tracks/system/midi increment. Coverage is still INCREMENTAL (the
-    /// edit surface, mixer/plugin inserts, and export/cleanup executors remain);
-    /// this is the covered subset the census pins.
+    /// The mutating operations B4 covers with a verified-execution oracle: the
+    /// FINAL uncovered mutating surface — the two plugin-insert paths, the export
+    /// executors, the cleanup-apply executor, and the one edit op that reaches a
+    /// genuine State A. Each was DERIVED BY READING its dispatcher/channel handler
+    /// and pinning what the verified path actually emits (cited per oracle below).
+    /// With this increment EVERY supported (non-`.unsupported`) mutating op is
+    /// either covered here or carries an audited structural exclusion — the
+    /// pure-code mutating oracle inventory is CLOSED (asserted by
+    /// `everyMutatingSpecIsCoveredOrAuditedExcluded`).
+    ///
+    /// COVERAGE CREDIT (#409): defining an oracle here is INVENTORY, not coverage.
+    /// R-SEM (`semanticCoverageIncomplete`) is credited to an operation ONLY when it
+    /// is live-exercised to a `.passed` semanticReadback verdict (the #284 live
+    /// matrix) or covered by a governed release-visible waiver — never by an oracle's
+    /// mere existence. So this B4 increment PINS the State-A contract these verified
+    /// executions must satisfy in the future live run; it does NOT itself reduce the
+    /// R-SEM static debt, and it does not change `ProductionReadinessContracts`.
+    ///
+    /// DELIBERATELY ABSENT (audited, no State-A envelope — see
+    /// `structurallyUnverifiedMutatingOperationIDs`): the 13 send-only edit ops
+    /// (undo/redo/cut/copy/paste/delete/select_all/split/join/quantize/
+    /// bounce_in_place/normalize/duplicate). Every one routes ONLY to
+    /// [.midiKeyCommands, .cgEvent] (RoutingTable) — a blind CC key command or a
+    /// CGEvent keystroke — and BOTH channels emit State B (send_only_no_readback /
+    /// readback_unavailable) with no edit read-back, so no State A exists.
+    /// edit.toggle_step_input is the one edit op that DOES reach State A (its
+    /// [.accessibility, …] chain verifies the Step Input Keyboard window open-state
+    /// flipped via an AX window-list read-back), so it is covered, not excluded.
+    static let phaseB4MutatingOperationIDs: Set<OperationID> = [
+        // mixer (1 — the AX slot-popup insert that reads back the slot name)
+        .mixerInsertPlugin,
+        // plugins (1 — the verified apply-back insert, HC v2 State A)
+        .pluginsInsertVerified,
+        // project (3 — the two export executors + the cleanup-apply rename executor)
+        .projectExportRun, .projectExportResume, .projectCleanupApply,
+        // edit (1 — the sole edit op with an AX State-A path; the other 13 are send-only)
+        .editToggleStepInput,
+    ]
+
+    /// Every mutating operation the table covers: the B1 verified-write increment
+    /// UNION the B2 transport/navigate increment UNION the B3
+    /// project/tracks/system/midi increment UNION the B4 plugin-insert / export /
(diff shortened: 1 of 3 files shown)


MongLong0214:
Retrospective CEO re-ratification at current main 16f0900f5ee84d12f9108a33c73145fa269b6747: the pure-code mutating inventory is substantively PASS. Current source proves covered 48 plus audited-excluded 38 equals all 86 supported mutating specs, while R-SEM remains OPEN. Current-main focused suites passed 161 tests; release-binary QA passed 5 tests; release build passed. The overall B queue remains QA_PENDING because PR #437 lacks live evidence and the original PRs lacked exact-head verdict/completion receipts.

PR #450, opened 2026-07-27
fix(#449): raise midi.import_file AppleScript outer timeout

### LINKED ISSUE #449: record_sequence/midi.import_file: outer 5.0s appleScriptTimeout is shorter than the script's own documented polling budget
## Summary

`tracks.record_sequence` / `midi.import_file` reliably fails with `ax_write_failed` / AppleScript `timedOut`, even though the underlying import appears to actually be happening (a real MIDI region gets placed on a new track — confirmed visually). This looks like a mismatch between the outer process-level timeout and the AppleScript's own internal polling budget, not a permissions or environment problem.

## Environment

- logic-pro-mcp v3.13.0 (Homebrew install, `monglong0214/logic-pro-mcp` tap)
- Logic Pro 12.3
- macOS (arm64)
- `LogicProMCP --check-permissions`: all four gates granted (Accessibility, Automation-LogicPro, Automation-SystemEvents, PostEvent/CGEvent)
- `LogicProMCP doctor --profile full` with a visible project window open: `midi_import`, `track_management`, `project_lifecycle`, `verified_plugin_applyback` all report `ready`

## Reproduction

Minimal MCP session: `initialize` → `notifications/initialized` → `tools/call` for `logic_tracks` with:
```json
{
  "command": "record_sequence",
  "params": {
    "bar": 1,
    "tempo": 85.0,
    "notes": "36,0,120,100,10;38,250,120,100,10;36,500,120,100,10;38,750,120,100,10"
  }
}
```

## Observed result

```json
{
  "note_count": 4,
  "verified": false,
  "tempo_dialog_seen": false,
  "failure_stage": "midi.import_file",
  "method": "smf_import",
  "import_error": "ax_write_failed",
  "error": "import_failure",
  "file_open_dialog_seen": false,
  "success": false,
  "detail": "{\"error\":\"ax_write_failed\",\"file_open_dialog_seen\":false,\"hint\":\"midi.import_file osascript failed: AppleScript error: timedOut\",\"state\":\"C\",\"success\":false,\"tempo_dialog_seen\":false,\"track_count_before\":1}"
}
```

Server stderr for the same call:
```
[WARN] [process] '/usr/bin/osascript' exceeded 5.0s timeout (pid 41739); sending SIGTERM (BoundedProcessRunner.swift:93)
[ERROR] [appleScript] AppleScript child process failed: timedOut (AppleScriptChannel.swift:492)
```

This reproduced identically across 3 separate attempts (including one after dismissing a leftover blocking `Import` dialog left open by a prior killed attempt).

**Despite every attempt reporting failure, a real MIDI region was visibly placed on a newly-created track** (confirmed by screenshot) — the import mechanically succeeded at least once, the tool just never got to report it.

## Suspected root cause

`ServerConfig.appleScriptTimeout` (`Server/ServerConfig.swift`) is a hardcoded `5.0` seconds, enforced externally by `BoundedProcessRunner` (SIGTERM on the whole `osascript` child process). But `AccessibilityChannel+MIDIImport.swift`'s `defaultImportMIDIFile` is documented (in its own doc-comment) as polling **up to ~5s** just to detect the file-open sheet, then polling again (~3s, 15×200ms) for the Import button to become enabled, then optionally handling a tempo-conflict dialog. Summed, the script's own designed step budget can legitimately exceed the outer 5.0s ceiling that kills it — so on any machine/timing where the individual steps land on the slower end of "normal," the whole operation is killed mid-flight regardless of environment health, producing a false failure even on a fully successful import.

The comment at `AccessibilityChannel+MIDIImport.swift:146` referencing `#188` suggests this general area is already known to be delicate — apologies if this is a duplicate of an existing tracked issue; I searched but couldn't find one specifically about the outer/inner timeout mismatch.

## Suggested fix

Either raise `appleScriptTimeout` for this specific operation (e.g. to ~10-12s to comfortably exceed the sum of the script's own internal poll budgets), or make the outer timeout configurable per-operation so `record_sequence`/`import_file` can declare a longer allowance than simpler synchronous AppleScript calls.

Happy to test a patched build against this same repro if useful.

IcelandicX:
## Summary
- Fixes #449: `midi.import_file` / `tracks.record_sequence` was killed by the global 5.0s `appleScriptTimeout` while the import AppleScript's own poll budgets can legitimately take ~17s.
- Adds `ServerConfig.midiImportAppleScriptTimeout = 20.0` and wires `defaultImportMIDIFile`'s default `executeScript` through `AppleScriptChannel.executeAppleScript(_:timeout:)`.
- Leaves the global 5.0s timeout unchanged for other AppleScript ops.

## Notes
- Logic 12.3 may still present a standalone Import browser rather than a sheet-attached open panel; that UI mismatch is separate from this timeout fix. Hosts that already use a helper path (`compose-midi.sh` / Import UI automation) should keep it until a dedicated 12.3 Import-browser fix lands.

## Test plan
- [x] Unit: `testServerConfigMIDIImportAppleScriptTimeoutExceedsDefault` (asserts import budget > default and ≥ 17s)
- [ ] Live: `record_sequence` / `midi.import_file` no longer logs `exceeded 5.0s timeout` on a healthy Logic 12.x session
- [ ] Confirm other AppleScript ops still use 5.0s

### DIFF
--- a/Sources/LogicProMCP/Channels/AccessibilityChannel+MIDIImport.swift
+++ b/Sources/LogicProMCP/Channels/AccessibilityChannel+MIDIImport.swift
@@ -126,11 +126,24 @@ extension AccessibilityChannel {
     /// lane the result is DOWNGRADED to State B `imported_as_gm_device` with a
     /// hint, because such lanes route to a General MIDI device and may bounce
     /// silent — a count delta alone must never be claimed audible-verified.
+    ///
+    /// #449: the default `executeScript` uses
+    /// `ServerConfig.midiImportAppleScriptTimeout` (not the global 5.0s
+    /// `appleScriptTimeout`) so BoundedProcessRunner does not SIGTERM the
+    /// osascript child before the script's own poll budgets can finish.
+    /// Note: Logic 12.3 may present a standalone Import browser rather than a
+    /// sheet-attached open panel; that UI mismatch is separate from the outer
+    /// timeout and may still require a helper path when sheet detection fails.
     static func defaultImportMIDIFile(
         systemEventsAuthorized: @Sendable () -> Bool = { PermissionChecker.checkSystemEventsAutomation() },
         path: String,
         runtime: AXLogicProElements.Runtime = .production,
-        executeScript: @escaping @Sendable (String) async -> ChannelResult = { await AppleScriptChannel.executeAppleScript($0) },
+        executeScript: @escaping @Sendable (String) async -> ChannelResult = {
+            await AppleScriptChannel.executeAppleScript(
+                $0,
+                timeout: ServerConfig.midiImportAppleScriptTimeout
+            )
+        },
         trackCount: (@Sendable () -> Int)? = nil,
         trackNames: (@Sendable () -> [String])? = nil,
         regionInfos: (@Sendable () -> MIDIImportRegionReadback)? = nil,
--- a/Sources/LogicProMCP/Channels/AppleScriptChannel.swift
+++ b/Sources/LogicProMCP/Channels/AppleScriptChannel.swift
@@ -479,12 +479,15 @@ actor AppleScriptChannel: Channel {
         await runtime.runScript(source)
     }
 
-    static func executeAppleScript(_ source: String) async -> ChannelResult {
+    static func executeAppleScript(
+        _ source: String,
+        timeout: TimeInterval = ServerConfig.appleScriptTimeout
+    ) async -> ChannelResult {
         await Task.detached(priority: .userInitiated) {
             let execution = BoundedProcessRunner.run(
                 executable: "/usr/bin/osascript",
                 arguments: appleScriptArguments(for: source),
-                timeout: ServerConfig.appleScriptTimeout,
+                timeout: timeout,
                 outputLimitBytes: 128 * 1024
             )
             let result = channelResult(forAppleScriptExecution: execution)
--- a/Sources/LogicProMCP/Server/ServerConfig.swift
+++ b/Sources/LogicProMCP/Server/ServerConfig.swift
@@ -19,6 +19,14 @@ struct ServerConfig: Sendable {
 
     // MARK: - Timeouts
     static let appleScriptTimeout: TimeInterval = 5.0
+    /// Outer osascript budget for `midi.import_file` / `tracks.record_sequence` SMF import.
+    /// Must exceed the AppleScript's own internal poll budgets in
+    /// `AccessibilityChannel+MIDIImport.defaultImportMIDIFile` (#449): file-open
+    /// sheet (~5s) + go-to-folder field (~3s) + Import-enabled (~4s) + tempo
+    /// dialog (~3s) + fixed delays/self-heal (~2s) ≈ 17s worst case. The default
+    /// 5.0s `appleScriptTimeout` SIGTERMs the child mid-flight and reports a
+    /// false `ax_write_failed` / `timedOut` even when the import later succeeds.
+    static let midiImportAppleScriptTimeout: TimeInterval = 20.0
 
     // MARK: - Logic Pro
     /// Resolved bundle ID for the active Logic Pro variant (desktop or Creator Studio).
--- a/Tests/LogicProMCPTests/ProductionReadinessTests.swift
+++ b/Tests/LogicProMCPTests/ProductionReadinessTests.swift
@@ -260,6 +260,14 @@ import Testing
     #expect(ServerConfig.mmcDeviceID == 0x7F)
 }
 
+/// #449: MIDI import's outer osascript budget must exceed the script's own
+/// internal poll ceilings so BoundedProcessRunner does not false-fail mid-import.
+@Test func testServerConfigMIDIImportAppleScriptTimeoutExceedsDefault() {
+    #expect(ServerConfig.appleScriptTimeout == 5.0)
+    #expect(ServerConfig.midiImportAppleScriptTimeout > ServerConfig.appleScriptTimeout)
+    #expect(ServerConfig.midiImportAppleScriptTimeout >= 17.0)
+}
+
 // MARK: - Architecture: StatePoller Graceful Stop
 
 @Test func testStatePollerStopAwaitsTaskCompletion() async {


MongLong0214:
Final production-readiness verdict: the direction is correct, but the current head `ed362a5187c9a713cd9060bb4caec0efb9e684c7` is not ready to merge.

The root-cause analysis is sound: `midi.import_file`, and `tracks.record_sequence` through the same import path, need a different execution budget from the general 5-second AppleScript limit. However, adding a 20-second constant alone does not yet provide enough safety or regression protection.

The following changes are required:

1. **Establish a justified timeout with adequate margin**
   - The explicit waits in the import AppleScript already total approximately 17.5–18.1 seconds.
   - A 20-second outer timeout leaves very little room for process startup, AX queries, menu interaction, panel response, and normal host variance.
   - Please derive the value from measured healthy worst-case behavior, or explicitly relate the inner polling budget to the outer timeout and include a defensible margin.

2. **Add tests for the production wiring**
   - The current test only compares constants. It would still pass if the code that forwards the import-specific timeout were removed.
   - At minimum, tests should prove that:
     - `midi.import_file` passes the import-specific timeout to the actual runner.
     - `tracks.record_sequence` reaches the same import path and uses that timeout.
     - Unrelated AppleScript operations continue to use the existing 5-second timeout.

3. **Handle partial mutation after timeout**
   - If the outer timeout fires after the Import button has been activated, a MIDI region may already have been created.
   - Returning this as an ordinary retryable failure can cause duplicate imports.
   - The error contract should preserve that a write was attempted, the mutation outcome is unknown, and an automatic retry is unsafe. Where possible, use independent readback to determine whether the mutation occurred.

4. **Add wrong-target protection**
   - During the longer operation, another Logic project may become frontmost.
   - Capture the intended target at the start and verify it immediately before and after the mutation.
   - If the target changes or cannot be verified, fail closed or return an explicit indeterminate mutation result.

5. **Provide exact-head verification evidence**
   - Focused tests and the full suite
   - Release build and artifact SHA-256
   - MCP stdio-path QA
   - Live Logic cases for `midi.import_file` and `tracks.record_sequence`, including success, timeout, and negative cases
   - Confirmation that unrelated AppleScript operations still retain the 5-second timeout
   - Exact-head CI PASS

   All evidence must be bound to the same final commit SHA.

6. **Remove public generation metadata**
   - Please remove the promotional generation footer from the PR body and the generation-tool co-author metadata from the commit.

### Relationship to the current ADR work and task ordering

The 300-second deadline in the ongoing ADR/Operation Registry work limits the overall MCP operation. It is a different layer from the process-level timeout that terminates the internal `osascript`. Therefore, the current ADR work has not already fixed this bug.

However, merging this PR independently ahead of the current canonical task would conflict with the established dependency order and evidence chain. There are two safe options:

- **Default order:** complete ADR-014-a/#302 R2 first, then include this change in the qualification scope so the import-specific timeout, target stability, partial-state behavior, and real writer QA are validated together.
- **Explicit pull-forward:** formally declare this issue a prerequisite blocker, update the locked scope, acceptance criteria, and dependency order, and then process all changes and evidence through the same production gate.

Without an explicit pull-forward decision, this PR should not be merged ahead of the current task. The current verdict is **FAIL / changes required**. Any new commit will invalidate the previous SHA-bound test, QA, review, and CI evidence, so all affected lanes must be rerun against the new exact head.

MongLong0214 (changes requested):
Thanks for this — the diagnosis is right and the structure of the fix is the one we want. Scoping a longer budget to just this lane, instead of raising the global `appleScriptTimeout`, is correct, and keeping `executeAppleScript`'s new `timeout:` parameter defaulted means every existing caller is unchanged. Reviewed at head `ed362a5187c9a713cd9060bb4caec0efb9e684c7`; it builds and the touched tests pass.

Two blockers before this can merge, both about the constant rather than the design.

### 1. `20.0` does not cover the worst case the comment itself describes

The doc comment budgets `sheet (~5s) + folder (~3s) + Import-enabled (~4s) + tempo (~3s) + fixed delays/self-heal (~2s) ≈ 17s`. The four polls are exactly right — they sum to 15.00s. The remainder is not ~2s. Enumerating every `delay` statement in `defaultImportMIDIFile` at this head:

| segment | bound | per-iteration | max |
|---|---|---|---|
| activate settle | fixed | 0.30 | 0.30 |
| stale-panel self-heal | `repeat 4` | 0.25 | 1.00 |
| file-open sheet poll | `repeat 20` | 0.25 | 5.00 |
| post-sheet fixed | fixed | 0.20 / 0.15 / 0.40 | 0.75 |
| go-to-folder poll | `repeat 20` | 0.15 | 3.00 |
| post-field fixed | fixed | 0.20 / 0.30 / 0.15 | 0.65 |
| Import-enabled poll | `repeat 20` | 0.20 | 4.00 |
| import-click self-heal | `repeat 3` | 0.20 | 0.60 |
| tempo-dialog poll | `repeat 15` | 0.20 | 3.00 |
| final self-heal | `repeat 3` | 0.20 | 0.60 |
| **total** | | | **18.90s** |

The non-poll portion is 3.90s, not ~2s — the comment understates by about 1.9s.

The early returns don't lower this ceiling. There is no path where all four polls exhaust (`if fileOpenSeen is false then return "DIALOG_NOT_FOUND…"` bails out, and the `goToSet` failure path self-cleans). But the most expensive path isn't exhaustion — it's each poll *succeeding on its final iteration*, which costs `(n-1) × delay`: 4.75 + 2.85 + 3.80 + 2.80 = 14.20s, plus 1.70s fixed and up to 2.20s of self-heal. So the realistic ceiling lands at roughly 18–19s either way.

Against a 20.0s budget that is 1.1–1.9s of headroom, **and that number counts `delay` statements only**. It assigns zero cost to the System Events queries themselves: those 79 loop iterations each perform at least one AX query before their delay, and `AXHelpers` caps a *single* AX message at 2.5s. Under load, one slow query consumes the entire margin.

Net effect: this reliably removes the 5.0s truncation, but a low-frequency false `timedOut` / `ax_write_failed` survives on the worst path. Please raise the constant to at least `30.0` and correct the comment's arithmetic to the measured 18.9s, noting that AX query and scheduling time sit on top of it.

### 2. The regression guard's floor is below the worst case

```swift
#expect(ServerConfig.midiImportAppleScriptTimeout >= 17.0)
```

17.0 is below the 18.90s measured above, so this passes for a value that reintroduces the truncation: at `17.0`, both `17.0 >= 17.0` and `17.0 > 5.0` hold. The `> appleScriptTimeout` clause is loose for the same reason — `5.1` satisfies it. Please bind the floor to `>= 30.0` and record where that number comes from in a comment next to the test, so the next person to touch it can check the arithmetic rather than trust the literal.

### 3. Worth a line in the comment (not a blocker)

`AXHelpers` records a client-side 15s `tools/call` timeout as a real constraint that has previously stalled the stdio loop. A budget above that means the client can abandon the request while the `osascript` child keeps driving Logic's UI for the remainder of the budget. This repo already has longer timeouts where the work justifies it (`bounceHelperTimeout` is 300s), so 15s is not a hard global contract and this doesn't block the change — but since item 1 pushes the budget higher, please document the interaction.

Happy to re-review as soon as these land. Note that CI on this PR has been sitting unapproved since it was opened — that's on our side, not yours, and we're clearing it.

MongLong0214:
## Pull-request disposition: convert to draft

This pull request remains ineligible to merge. Its replacement coverage does not yet exist, so it must not be closed or treated as superseded.

Convert it to draft. Draft status preserves the branch, code, discussion, and requested changes while accurately recording that the change is not ready for review or merge. It may return to ready-for-review status only after the timeout program reaches the prerequisite coverage and the change is corrected and requalified on its then-current exact head.

MongLong0214:
Superseded. `#449` is fixed on main by #462, which landed replacement coverage this PR did not have.

What differs: rather than raising the outer bound alone, the merged fix gives `midi.import_file` its own
bound and derives the required floor from the generated script, so a later change that widens a polling loop
past the bound fails a test instead of silently reintroducing the timeout. #462 also added the monotonic
segment duration (`#452`), which makes that budget measurable from outside the process for the first time.

Closing rather than merging, since the two changes would conflict and this branch's coverage is the weaker
of the two. The branch is left intact.

PR #454, opened 2026-08-02
feat(ax): add typed AX status internal API, uncalled

MongLong0214:
Adds a typed accessibility status API to `AXHelpers` as an internal, **caller-free** addition.

## What this adds

Four new internal symbols, none of which is invoked anywhere in `Sources`:

- `AXStatusError` — a typed error carrying the raw `AXError` value
- `projectChildrenStatus(_:_:)` — a pure projection from an AX read outcome to a typed result
- `childrenResult(_:runtime:)` — a children read that preserves a non-success status instead of flattening it
- `Runtime.childrenResult` — an injection seam, with an explicit memberwise initializer defaulting it to `nil`

The pre-existing `getChildren` is unchanged and every existing call site continues to use it. The tool
surface, the response envelope and the error taxonomy are unchanged.

## Scope

Exactly two files: `Sources/LogicProMCP/Accessibility/AXHelpers.swift` (additive, 64 inserted / 0 deleted)
and `Tests/LogicProMCPTests/AXStatusProjectionTests.swift` (new, 74 lines).

## Verification at this exact head

| lane | result |
|---|---|
| focused suite | 5 tests passed |
| full suite `swift test --no-parallel` | **3330 passed / 0 failed** |
| baseline at the merge base | **3325 passed / 0 failed** |
| debug build / release build | both succeed |
| test-target compile | succeeds |
| release artifacts | `LogicProMCP` and `trusted-verifier` both built and hashed |
| `Package.resolved` | unchanged; verified before and after every build and test lane |

The full suite differs from the baseline by exactly the five new tests: no regression, no test removed.

## Mutation testing

Each new assertion was verified to be live by mutating the implementation and confirming that the
assertion — and only that assertion — fails:

| mutation | assertion that failed |
|---|---|
| collapse every non-success status onto one raw value | `rawValuePreservedForEveryPinnedStatus` |
| map an absent value to an error | `absentValueIsEmptySuccess` |
| map a malformed non-array value to an error | `malformedNonArrayIsEmptySuccess` |
| return an empty collection for a well-formed value | `wellFormedChildrenProjectToExpectedElements` |
| give the runtime seam a non-`nil` default | `productionRuntimeSeamIsAbsent` |

Every mutation produced exactly one failing test, and the tree was restored and re-verified after each.

## Behaviour notes

Fail-closed in both directions: a non-success status is surfaced with its own raw value, a successful
read never invents an error, and an absent value is an empty success rather than an error. Seven
distinct `AXError` raw values round-trip distinctly.

**Known limitation:** an absent children value and a malformed non-array value both project to an empty
success, so a caller cannot distinguish the two from the result alone. Neither is an accessibility
error, so this is the intended contract, but it is recorded here for any future caller that needs to
tell them apart.

No logging, file or network calls are added. No deadline, retry or blocking wait is introduced. The
addition is read-only and pure, so no operation can be left half-applied.

## Manual/live QA: LIVE_NA

No live verification is claimed. The shipment is caller-free and adds no user-reachable behaviour, so a
live session could not distinguish this build from its base — a live PASS here would be unfalsifiable.
LIVE_NA records that live evidence is **inapplicable**, not that it was obtained.

## This is not a completion

This change ships capability only. It does not close its parent work item, grants no completion credit,
and does not authorise the deferred live verification lane that work still owes.

### DIFF
--- a/Sources/LogicProMCP/Accessibility/AXHelpers.swift
+++ b/Sources/LogicProMCP/Accessibility/AXHelpers.swift
@@ -11,6 +11,29 @@ enum AXHelpers {
         let children: @Sendable (AXUIElement) -> [AXUIElement]
         let performAction: @Sendable (AXUIElement, String) -> Bool
         let childCount: @Sendable (AXUIElement) -> Int?
+        /// Injectable typed-children seam. `nil` means "use the production read".
+        let childrenResult: (@Sendable (AXUIElement) -> Result<[AXUIElement], AXStatusError>)?
+
+        /// Explicit memberwise initializer, replacing the compiler-synthesized one so
+        /// `childrenResult` can default to `nil`. Every existing construction that omits
+        /// it continues to compile unchanged.
+        init(
+            axApp: @escaping @Sendable (pid_t) -> AXUIElement,
+            attributeValue: @escaping @Sendable (AXUIElement, String) -> AnyObject?,
+            setAttributeValue: @escaping @Sendable (AXUIElement, String, CFTypeRef) -> Bool,
+            children: @escaping @Sendable (AXUIElement) -> [AXUIElement],
+            performAction: @escaping @Sendable (AXUIElement, String) -> Bool,
+            childCount: @escaping @Sendable (AXUIElement) -> Int?,
+            childrenResult: (@Sendable (AXUIElement) -> Result<[AXUIElement], AXStatusError>)? = nil
+        ) {
+            self.axApp = axApp
+            self.attributeValue = attributeValue
+            self.setAttributeValue = setAttributeValue
+            self.children = children
+            self.performAction = performAction
+            self.childCount = childCount
+            self.childrenResult = childrenResult
+        }
 
         static let production = Runtime(
             axApp: { pid in
@@ -77,6 +100,47 @@ enum AXHelpers {
     /// Get the children of an AX element.
     static func getChildren(_ element: AXUIElement, runtime: Runtime = .production) -> [AXUIElement] {
         runtime.children(element)
+    }
+
+    /// Typed AX status error carrying the raw `AXError` rawValue.
+    struct AXStatusError: Error, Sendable, Equatable {
+        let raw: Int32
+
+        init(raw: Int32) {
+            self.raw = raw
+        }
+    }
+
+    /// Pure projection of an AX read outcome into a typed result.
+    ///
+    /// Nothing is fabricated in either direction: a non-success status keeps its own raw
+    /// value rather than collapsing to a single code, and a successful read never invents
+    /// a status. Absent children (`nil`) and a malformed non-`CFArray` value are both an
+    /// empty success, because neither is an AX error — the earlier defect here was a
+    /// synthesised `noValue` for the absent case.
+    static func projectChildrenStatus(
+        _ status: AXError,
+        _ value: AnyObject?
+    ) -> Result<[AXUIElement], AXStatusError> {
+        guard status == .success else {
+            return .failure(AXStatusError(raw: status.rawValue))
+        }
+        guard let value else { return .success([]) }
+        return .success(decodeChildrenArray(value))
+    }
+
+    /// Read an element's children, preserving a non-success AX status as a typed error
+    /// instead of flattening it to an empty array the way `getChildren` does.
+    static func childrenResult(
+        _ element: AXUIElement,
+        runtime: Runtime = .production
+    ) -> Result<[AXUIElement], AXStatusError> {
+        if let injected = runtime.childrenResult {
+            return injected(element)
+        }
+        var value: AnyObject?
+        let status = AXUIElementCopyAttributeValue(element, kAXChildrenAttribute as CFString, &value)
+        return projectChildrenStatus(status, value)
     }
 
     /// Decode a raw `kAXChildrenAttribute` value into `[AXUIElement]`, guarding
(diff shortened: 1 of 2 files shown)


MongLong0214:
## Review-gate record

This change went through a staged review before reaching a pull request: a design record, then a
requirements record, then an execution ticket, each independently reviewed and re-issued until it
passed. Recording the decisions that shaped the code, since several of them are visible in the diff and
would otherwise look arbitrary.

**Split obligations rather than one compile check.** "Every existing runtime construction still compiles"
was initially owned by the product build alone. At the merge base there are three construction sites —
one under `Sources` and two under `Tests` — and a product build does not compile test targets, so two of
the three were asserted and never proved. The obligation was split, and the test-target compile is now
its own lane.

**Two independent branches, two mutants.** An absent value and a malformed non-array value are separate
paths through the projection. A single mutation could only ever kill one of them, so each branch has its
own assertion and its own mutation.

**Seven raw values, not one.** A single status value cannot detect a collapse, because one value
trivially equals itself. The assertion compares the whole recovered sequence against seven pinned
values, and the count itself is asserted so a silently shortened list fails.

**The mutation check verifies a failure, not a name.** The test runner prints every executed test name,
so searching the log for a function name succeeds even when that function passed. The check matches the
anchored failure line for the intended function and additionally requires that exactly one test failed.

**Dependency lockfile pinned during verification.** On the machine used for verification, a plain build
rewrites `Package.resolved` because the local toolchain differs from the one CI selects. Every build and
test lane therefore pins resolution, and the lockfile digest is checked before and after each of the
eighteen invocations — thirty-six checks, all at the committed digest. Without that pin, later lanes
would have run against a tree that no longer matched the commit under test, and CI's own
`Package.resolved` check would have failed.

**Assertion style.** Assertions bind a typed intermediate and expect it directly. Forms this project has
previously found to be inert — comparing an optional against `true`, coalescing with `?? false`, or
comparing against `.some(true)` — are absent from the new file.

## Standing limitation

An absent children value and a malformed non-array value both project to an empty success. That is the
intended contract, since neither is an accessibility error, but a caller cannot distinguish them from
the result alone. Any future caller needing that distinction will need a richer result type.

## Disposition

Capability only, with no caller. Not a completion of the parent work item, no completion credit, and the
deferred live verification that work still owes remains outstanding and unauthorised by this change.

MongLong0214:
## Merge-gate verdict: QA pending

The exact-head code and evidence review remains PASS for `0690affdb3f16897c34379ac2f55986aca305273`. The pull-request check is green on that same head, and the reviewed base remains unchanged.

Merge is blocked by repository state: two non-draft program pull requests are currently open, while the serial merge policy permits at most one. The other pull request is #450, which is not eligible to merge and does not yet have replacement coverage.

Disposition: convert #450 to draft. Do not close or supersede it. Draft status preserves its code and review history while accurately marking it as not ready.

After that transition is verified through GitHub, re-read this pull request's head, base, mergeability, check result, and the open pull-request inventory. If the head is still `0690affdb3f16897c34379ac2f55986aca305273`, the base is still `aabf5e0fd6d0fb9fb2760e71a6daed5d2933a47b`, the check remains green, the pull request remains clean and mergeable, and it is the sole non-draft program pull request, merge may proceed. Any mismatch stops the transition and requires a new gate review.

This comment grants no completion claim. Post-merge verification remains mandatory.

MongLong0214:
## Post-merge verification

Merged as `a65fcca3e09a58ac59128017f43676aa50225f4c`.

| check | result |
|---|---|
| merge commit parents | exactly the previous `main` and the reviewed head — no third parent, no rebase |
| files changed against previous `main` | `Sources/LogicProMCP/Accessibility/AXHelpers.swift` modified, `Tests/LogicProMCPTests/AXStatusProjectionTests.swift` added — nothing else |
| shipped content on `main` | both files hash to the same values that were reviewed and tested at the head; the merged bytes are the verified bytes |
| `Package.resolved` | unchanged by this merge |
| post-merge CI on `main` | **success** |

The pre-merge run at the head and the post-merge run on `main` both passed, so the change is verified
before and after integration rather than only before.

## What shipped, and what did not

Capability only: four internal symbols with no caller anywhere in `Sources`. The pre-existing children
read, the tool surface, the response envelope and the error taxonomy are all unchanged, so no
user-visible behaviour moves with this merge.

This merge is **not** a completion of the parent work item. It grants no completion credit, and the
deferred live verification that work still owes remains outstanding — this change neither performs it
nor authorises it. A future change that wants to rely on that verification still has to earn it.

MongLong0214:
## Independent post-merge verification: PASS

The integrated result has been re-verified against the reviewed head.

- Pull request #454 is merged as `a65fcca3e09a58ac59128017f43676aa50225f4c`.
- The merge commit has exactly two parents: previous `main` at `aabf5e0fd6d0fb9fb2760e71a6daed5d2933a47b` and reviewed head `0690affdb3f16897c34379ac2f55986aca305273`.
- The merge commit and reviewed head have the same tree, `bd1087148770eea33e8c2fb05810ef38f73b27db`; therefore the integrated repository bytes are exactly the reviewed repository bytes.
- Remote `main` resolves to the merge commit.
- The base-to-merge comparison contains exactly the two reviewed files. `Package.resolved` is absent from the comparison.
- The source file hashes to `bb33af31068f928f1f722e6e5ff878c4bc932401f9c5fb517931eb8807765e25`; the test file hashes to `326a39f12c333eaf80e8e2914d62f99fcacfb932606fea698fc4facb3a46dc15` when fetched from the merge commit.
- The pull-request check passed on the reviewed head, and the post-merge `main` check passed on the merge commit.
- Pull request #450 remains open as a draft; no other pull request is open.

The post-merge gate is PASS. This remains a caller-free capability shipment, not terminal completion of the parent lane. Its deferred live obligation remains outstanding.

PR #462, opened 2026-08-07
fix: default-install path — import timeout, keycmd remediation path, approval ordering, goto_position fallback

### LINKED ISSUE #449: record_sequence/midi.import_file: outer 5.0s appleScriptTimeout is shorter than the script's own documented polling budget

### LINKED ISSUE #452: Observability: emit a monotonic duration for the AppleScript segment of midi.import_file
## Problem

`midi.import_file` routes through the accessibility channel (`RoutingTable.swift`: `"midi.import_file": [.accessibility]`) and is implemented by `defaultImportMIDIFile` in `AccessibilityChannel+MIDIImport.swift`. That implementation spans **three distinct time budgets**, and only one of them is measurable from outside the process:

| # | budget | bound | externally observable? |
|---|---|---|---|
| 1 | operation deadline | `DeadlineClass.long` = 300s | yes — end-to-end response wait |
| 2 | script segment (`executeScript`, default `AppleScriptChannel.executeAppleScript`) | `appleScriptTimeout` | **no** |
| 3 | `deltaPoll` loop + surrounding Swift AX work | none | **no** |

Budget 2 is the one #449 concerns. Because no timing for it crosses the process boundary, **no external harness can measure it**, and therefore no measurement can demonstrate that a change to that timeout is sufficient.

Two adjacent facts make this hard to work around rather than merely inconvenient:

- `system.get_trace` serialises `"timestamp"` from `event.wallClock`. The monotonic `ContinuousClock.Instant` carried by `TraceEvent` never leaves the process, so trace output cannot supply a monotonic interval.
- `executeScript` is an injected dependency. Injecting a timing wrapper in a test would measure the injected executor rather than the production path, so it cannot serve as evidence about production behaviour.

## Proposed change

Emit a monotonic duration for the script segment through an existing public surface, so the segment governed by `appleScriptTimeout` becomes externally measurable without altering the production execution path.

Sketch, not a fixed design:

- record a monotonic start/end around the `executeScript` invocation inside `defaultImportMIDIFile`
- surface the resulting interval as a trace attribute (or an equivalent read-only surface) alongside the existing phases
- keep the production executor unchanged; the emission is observational only

## Why it matters

Without it:

- #449's remedy cannot be validated by measurement — only by inspection of summed `delay` statements, which excludes every AX query and file operation between the delays and so cannot establish real headroom
- budget 3 remains entirely unbounded and unobserved
- any timeout-budget invariant is limited to the operation deadline and must state explicitly that the inner segment is uncovered

## Acceptance

- [ ] A monotonic duration for the script segment is observable through a public read-only surface
- [ ] The production executor and control flow are unchanged; the emission adds observation only
- [ ] The emitted value is a duration measured with a monotonic clock, not a wall-clock difference
- [ ] A test asserts the emitted duration is present, finite and non-negative
- [ ] Documented as the measurement basis that #449 validation depends on

## Relationship to other work

- #449 — the defect whose remedy this makes measurable
- Blocks live measurement of budget 2 for any harness

### LINKED ISSUE #456: doctor recommends bare install-keycmds.sh, but Homebrew installs it only in pkgshare
## Summary

On the current stable Homebrew install, `LogicProMCP doctor --profile full` recommends the command `install-keycmds.sh`, but Homebrew installs that helper in formula `pkgshare`, not on `PATH`.

The package itself is complete: the helper and its required sibling `keycmd-preset.plist` are present together. The failure is in remediation discovery and documentation. A user who follows the printed command cannot run it; a user who downloads only the script gets a deterministic missing-preset error.

Reported by a real Homebrew user in Discussion #455.

The same hard-coded remediation is still present on current `main` at `a65fcca3e09a58ac59128017f43676aa50225f4c`.

## Reproduction

Version and install source:

```text
$ LogicProMCP --version
3.13.0

$ LogicProMCP doctor --profile full --json
install_source: homebrew
install.share_dir: pass — expected helper assets are present
channels.keycmd_reference: manual
remediation.type: command
remediation.value: install-keycmds.sh
```

The printed remediation is not executable:

```text
$ command -v install-keycmds.sh
NOT_ON_PATH
```

Homebrew did install both required files:

```text
$ brew list logic-pro-mcp | grep -E 'install-keycmds|keycmd-preset'
/opt/homebrew/Cellar/logic-pro-mcp/3.13.0/share/logic-pro-mcp/install-keycmds.sh
/opt/homebrew/Cellar/logic-pro-mcp/3.13.0/share/logic-pro-mcp/keycmd-preset.plist
```

Downloading or copying only the script reproduces the user's failure:

```text
=== LogicProMCP Key Commands Installer ===
ERROR: Preset file not found: <script-directory>/keycmd-preset.plist
exit 1
```

The packaged command succeeds because the sibling preset is present:

```sh
bash "$(brew --prefix logic-pro-mcp)/share/logic-pro-mcp/install-keycmds.sh"
```

In an isolated target directory this exited 0 and the staged plist SHA-256 matched the packaged source plist exactly.

## Root cause

- `keycmdReferenceCheck` hard-codes `remediationValueOverride` to the bare string `install-keycmds.sh`.
- `productionShareDirProbe` already resolves and validates the Homebrew shared directory, including both `install-keycmds.sh` and `keycmd-preset.plist`, but that result is not used to construct the channel remediation.
- `docs/SETUP.md#doctor-channelskeycmd-reference` repeats the same bare command.
- The installer intentionally resolves `keycmd-preset.plist` relative to `dirname "$0"`, so a detached script cannot work by itself.

This is not a missing Homebrew asset and does not require cloning the repository.

## Expected

- Homebrew doctor output should emit an executable, install-source-aware command using the resolved shared directory.
- Source-build documentation should use the source-tree form, such as `bash Scripts/install-keycmds.sh` from the repository root.
- The remediation anchor should explain that the helper and preset must remain adjacent and that downloading the script alone is unsupported.
- The documentation should state immediately that Logic 12.2+ uses this only to stage a mapping reference; actual assignments still require manual MIDI Learn.

## Acceptance criteria

- A clean Homebrew install can copy-paste the `channels.keycmd_reference` remediation from `doctor` successfully.
- The emitted helper path is proven executable and its sibling preset is proven present.
- Homebrew, release-package, and source-tree layouts each receive a correct command or a precise docs anchor.
- A regression test fails if doctor emits a bare command that is not installed on `PATH`.

Related: #26, #22, Discussion #455.

### LINKED ISSUE #457: doctor fix plan asks users to approve manual channels before completing setup
## Summary

`LogicProMCP doctor --profile full` currently tells a new user to approve manual-validation channels before completing the setup that those approvals are supposed to attest.

This is more than ordering polish. `--approve-channel` records an operator assertion; it does not configure or verify Logic. Presenting that command before Key Commands staging/manual MIDI Learn and Scripter setup can create a durable false attestation.

Discussion #455 shows the exact failure mode: the user followed the fix plan, approved both channels, and only afterward discovered that Key Commands had not been staged and the setup command was unavailable.

The same ordering and missing dependency are still present on current `main` at `a65fcca3e09a58ac59128017f43676aa50225f4c`.

## Reproduction

With v3.13.0, a fresh approval store, and `--profile full`:

```json
{
  "fix_plan": [
    "mcp.claude_code_registration",
    "channels.manual_validation",
    "channels.keycmd_reference",
    "channels.mcu_wiring_hint"
  ]
}
```

The two relevant checks are:

```text
channels.manual_validation
  status: manual
  remediation: LogicProMCP --approve-channel MIDIKeyCommands && LogicProMCP --approve-channel Scripter

channels.keycmd_reference
  status: manual
  evidence: preset_staged=false
  remediation: install-keycmds.sh
```

Executing the first remediation immediately records approval even though setup is absent:

```text
$ LogicProMCP --approve-channel MIDIKeyCommands
Approved MIDIKeyCommands for runtime use.
```

A subsequent `doctor --profile keycmd --json` reports `channels.manual_validation=pass` while `channels.keycmd_reference` remains manual.

The Key Commands helper only stages a reference plist. Logic 12.2+ still requires manual MIDI Learn, so file presence alone cannot repair the premature approval or prove the claimed validation happened.

## Root cause

- `channels.manual_validation` is declared before `channels.keycmd_reference` in `checkDefinitions`.
- Neither check declares the other as a dependency.
- `computeFixPlan` sorts equal-severity manual checks by declaration order.
- The manual-validation remediation chains approval for both MIDIKeyCommands and Scripter even though each has a different prerequisite and may be intentionally unused.

The result conflicts with the documented rule: approve a channel only after completing its matching Logic-side setup.

## Expected

Manual approval must be the final attestation step for each selected channel, never the first setup step.

- For MIDIKeyCommands: decide whether the capability is needed, stage the reference, complete and verify manual MIDI Learn, then approve.
- For Scripter: place and verify the intended insert, then approve.
- If a channel is not needed, offer the explicit skip command instead of a combined approval command.

## Acceptance criteria

- A fresh `keycmd` or `full` profile never places MIDIKeyCommands approval before its setup prerequisites.
- A fresh `legacy-scripter` or `full` profile never places Scripter approval before its setup prerequisites.
- The fix plan does not chain approvals for independently optional channels.
- Tests assert prerequisite-before-attestation ordering and the documented approve/skip semantics.
- Doctor cannot present a sequence that reaches channel readiness from preset staging plus a pre-recorded, unverified approval.

Related: #26, Discussion #455.

### LINKED ISSUE #460: ChannelRouter suppresses safe goto_position fallback after AX element_not_found
## Summary

`transport.goto_position` declares four routing candidates (`Accessibility -> MCU -> CoreMIDI -> CGEvent`), but an Accessibility `element_not_found` result stops the chain before any later candidate can run.

This is not just a missing one-off exception. The router classifies error codes globally, while fallback safety depends on the operation, the failing channel, whether a write may have happened, and whether the next channel can preserve the same semantics.

Reported from a real default-install composition workflow in Discussion #455: https://github.com/MongLong0214/logic-pro-mcp/discussions/455#discussioncomment-17907427

## Exact local reproduction

Reproduced on current `main` at `a429df72fc8bdb9c8fb5d3fdb3d4bd4aa4aea7cc`.

A focused exact-head run passed the existing tests that lock the two router branches involved:

```text
swift test --skip-build --filter 'doctorV4HumanReadinessSummaryIsAdditiveAndLeavesJSONStable|testTerminalStateCDoesNotFallThrough|testRouterTransportPlayFallsBackToAppleScriptAfterAXLookupMiss'
3 tests passed
```

The source combination deterministically selects the wrong branch for `goto_position`:

1. `RoutingTable.swift` declares `transport.goto_position: [.accessibility, .mcu, .coreMIDI, .cgEvent]`.
2. `AccessibilityChannel+Transport.swift` returns typed State C `element_not_found` when neither the goto dialog nor the bar slider is locatable.
3. `HonestContract.terminalErrorCodes` classifies `element_not_found` as globally terminal.
4. `ChannelRouter.shouldContinueAfterTerminalStateC` allows that failure to continue only for a hard-coded set of transport toggles. `transport.goto_position` is absent.
5. The router therefore returns the AX envelope immediately. The declared CGEvent implementation is never reached.

The intermediate route is also misleading: MCU has no `transport.goto_position` handler, CoreMIDI explicitly says it cannot position the playhead, while CGEvent implements the goto-position sequence.

## Root cause

`ChannelRouter` asks a global question - "is this error code terminal?" - where it must ask a contextual question:

- Is the failure request/domain-scoped or only channel/control-surface-scoped?
- Did it happen before any write, after a confirmed write, or with side effects unknown?
- Does the next channel support the same operation and parameter domain?
- Can fallback preserve target identity and avoid a vacuous or wrong-target success?

`element_not_found` currently conflates at least two incompatible meanings:

- a semantic target does not exist (for example `track.select` index 99999): terminal;
- this channel's UI control is unavailable (for example no AX goto dialog/slider): potentially recoverable by another channel.

Adding `goto_position` to the existing allowlist would patch one symptom and leave the same class of bug in place.

## Similar-scenario audit surface

At the reproduced head, the routing table contains 142 routes, including 65 multi-channel routes. Twenty are Accessibility-first multi-channel routes; fifteen of those are outside the current AX `element_not_found` allowlist.

Other globally terminal codes also need contextual review where they can describe a channel-local pre-write limitation: `not_implemented`, `port_unavailable`, `system_events_automation_denied`, and `dialog_not_found`. This does **not** mean all should fall through; it means every emitted failure class on a multi-channel route needs an explicit safe disposition.

There is also an inverse hazard: free-form channel errors currently fall through by default, while a better-typed error may stop. Improving error typing must not accidentally reduce valid fallback.

## Required root solution

1. Replace bare `[ChannelID]` route entries with a shared typed `RouteSpec` that declares channel order, per-channel operation/parameter support, verification strength, and explicit fallback policy.
2. Keep failures typed internally until the router finishes. A channel failure must carry at least:
   - scope: request / semantic target / channel control surface / session;
   - phase: preflight / pre-write / post-write;
   - side-effect state: none / possible / confirmed / rolled back;
   - stable error code and fallback disposition.
3. Allow fallback only when the failure is channel-scoped, pre-write, side-effect-free, the next channel supports the same semantics, and the route explicitly permits it.
4. Split ambiguous errors, including channel-control lookup failure versus semantic target absence, and channel-unsupported versus operation-not-exposed.
5. Encode the final Honest Contract envelope only at the external boundary, preserving attempted-channel and selected-channel provenance.
6. Remove the operation-name exception set once the typed route policy covers it.

## Acceptance criteria

- [ ] RED regression: AX returns control-surface-not-found for `transport.goto_position`; unsupported MCU/CoreMIDI candidates are skipped by declared capability; CGEvent executes exactly once.
- [ ] The final result identifies attempted/skipped channels and the selected channel without claiming verified State A for a send-only CGEvent result.
- [ ] Invalid goto parameters remain request-terminal with zero channel attempts.
- [ ] Semantic target absence such as `track.select` index 99999 remains terminal with zero wrong-target fallback.
- [ ] Any post-write, partial-state, or side-effect-unknown failure suppresses fallback unless rollback is explicitly verified.
- [ ] A structural validator fails every multi-channel `RouteSpec` that lacks an explicit disposition for a failure class its channels can emit.
- [ ] A route destination cannot be declared unless that channel declares handler and parameter-domain support.
- [ ] Representative tests cover request-terminal, target-terminal, channel-recoverable, channel-unavailable, pre-write permission, post-write mismatch, and partial-state classes; no giant per-operation exception list is introduced.
- [ ] Live exact-artifact QA covers an AX-missing goto surface and reaches the safe CGEvent path. CGEvent foreground/truthfulness requirements remain owned by #440.
- [ ] Focused tests, full suite, release build/artifact provenance, exact-head review, and CI pass before merge.

## Related

- Discussion #455 follow-up: https://github.com/MongLong0214/logic-pro-mcp/discussions/455#discussioncomment-17907427
- #440 (CGEvent frontmost handling and truthful sent-but-unverified behavior)
- #456 and #457 are separate onboarding defects and do not cover this router failure.

MongLong0214:
Four defects reported against the default-install path, three of which a new user hits before doing anything else. Each fix is a separate commit with its own reasoning, and each is verified by mutation — reverting the fix fails the new test.

### #449 — `midi.import_file` timed out on imports that succeeded

The import script drives four polled stages (the import sheet, the path dialog, the import button, the tempo prompt). Summing one delay per iteration gives a **17.2 s floor**, but the shared `appleScriptTimeout` is 5.0 s, so `osascript` was killed mid-import and the caller saw `ax_write_failed` / `timedOut` for a region that had actually landed.

Adds a dedicated `midiImportAppleScriptTimeout` (30.0 s) and an optional `timeout` parameter on `executeAppleScript` that defaults to the existing bound — the other fifteen call sites are untouched, so short lifecycle paths keep failing fast.

The test **derives** the polling floor from the generated script rather than asserting `== 30.0`; a hard-coded assertion would pass while someone widened a loop past the bound. It also guards against a vacuous pass (`floor > 0`) and tests the parser itself.

### #456 — doctor recommended a command that does not exist

Remediation printed a bare `install-keycmds.sh`, but Homebrew installs it only under `pkgshare`. Now derived from the resolved share dir, with an explicit message when the share dir is unresolved instead of a path that cannot be run.

### #457 — doctor asked users to approve setup that had not happened

`--approve-channel` records an *operator assertion*; it neither configures nor verifies Logic. The fix plan offered it **before** `channels.keycmd_reference`, so users approved first and only afterwards found Key Commands were never staged — a durable false attestation the helper cannot repair, since Logic 12.2+ still needs a manual MIDI Learn no file-presence check can prove.

`channels.manual_validation` now declares the dependency and reports `skipped` / `blocked_by` while staging is incomplete. `computeFixPlan` drops blocked checks, so the premature command leaves the plan rather than merely sorting later.

Two ordering details this rests on:
- `keycmd_reference` now precedes `manual_validation` in the check definitions — a dependency must have run before its status can be read, and the old order was the root cause. Check-ID order is part of the JSON contract, so the contract test is updated with the reason rather than silently absorbing the move.
- The block is evaluated **after** the profile guard; evaluating it first made `--profile core` report "waiting on staging" for channels that profile never requires.

### #460 — `goto_position` stopped the chain on a recoverable miss

`transport.goto_position` declares four candidate channels, but an Accessibility `element_not_found` ended the chain before any later candidate ran — so a workflow whose first action was `goto_position` failed with three usable channels behind the miss.

The fallback set was keyed on *"is a transport toggle"*, which is not the property that makes a fallthrough safe. It is renamed and the actual admission test is written down: `element_not_found` is a **pre-write** miss (the control was never located, so nothing was actuated and no partial write can be stranded), the failure is Accessibility-specific rather than a rejection of the request, and every later candidate expresses the same semantics. Recording the predicate matters more than adding the one op — it states which operations may **not** be added.

### #452 — the AppleScript segment was unmeasurable

`midi.import_file` runs under three nested time budgets, and only the outermost — the operation deadline — crossed the process boundary. The middle one is exactly the budget #449 is about, so its remedy could be *argued* from summed `delay` statements but never **measured**: those sums omit every AX query and file operation between the delays.

Adds a `script_segment.completed` trace phase carrying `applescript_duration_ms`. Three things make it evidence rather than decoration:

- Measured with `ContinuousClock`, not a `Date` difference — a timing number an NTP correction can invent is not proof.
- The clock brackets the **call site**, not the executor. Wrapping the injected closure would measure the wrapper and say nothing about production.
- The key is declared in `attributePrivacyClasses`. Undeclared keys are dropped silently, so an emission without that row would record nothing and expose nothing while the source still read as though it emitted.

The bound the segment ran under is deliberately *not* emitted beside it: the executor is injected, so the call site cannot attest to a timeout it does not own.

### Testing

Full suite `--no-parallel`: 3338 tests. The one failure — the library-inventory panel snapshot — reproduces identically on the unmodified tree and is unrelated to this branch.

Every new suite locks **both** directions. A gate that only ever blocks is as wrong as one that never blocks, and a fallback set that only grows degrades into blanket retry — so the release path and the still-refused path are asserted alongside the fix.

Reported in #455.

Closes #449
Closes #452
Closes #456
Closes #457
Closes #460

### DIFF
--- a/Sources/LogicProMCP/Channels/AccessibilityChannel+MIDIImport.swift
+++ b/Sources/LogicProMCP/Channels/AccessibilityChannel+MIDIImport.swift
@@ -130,7 +130,9 @@ extension AccessibilityChannel {
         systemEventsAuthorized: @Sendable () -> Bool = { PermissionChecker.checkSystemEventsAutomation() },
         path: String,
         runtime: AXLogicProElements.Runtime = .production,
-        executeScript: @escaping @Sendable (String) async -> ChannelResult = { await AppleScriptChannel.executeAppleScript($0) },
+        executeScript: @escaping @Sendable (String) async -> ChannelResult = {
+            await AppleScriptChannel.executeAppleScript($0, timeout: ServerConfig.midiImportAppleScriptTimeout)
+        },
         trackCount: (@Sendable () -> Int)? = nil,
         trackNames: (@Sendable () -> [String])? = nil,
         regionInfos: (@Sendable () -> MIDIImportRegionReadback)? = nil,
--- a/Sources/LogicProMCP/Channels/AccessibilityChannel+MIDIImport.swift
+++ b/Sources/LogicProMCP/Channels/AccessibilityChannel+MIDIImport.swift
@@ -392,7 +392,18 @@ extension AccessibilityChannel {
         end importMIDI
         return importMIDI()
         """
+        // #452: the only measurement of the segment the osascript bound governs.
+        // The clock brackets the call rather than the executor, so it observes
+        // whatever actually ran — wrapping the injected closure instead would
+        // measure the wrapper and prove nothing about production. Recorded on
+        // every outcome, including timeout, because a segment that hit its bound
+        // is exactly the one an operator needs the number for.
+        let segmentStart = ContinuousClock.now
         let result = await executeScript(script)
+        await OperationTraceContext.record(
+            .scriptSegmentCompleted,
+            attributes: ["applescript_duration_ms": elapsedMilliseconds(since: segmentStart)]
+        )
         switch result {
         case .success(let output):
             let scriptOutput = normalizedAppleScriptPayload(output)
--- a/Sources/LogicProMCP/Channels/AppleScriptChannel.swift
+++ b/Sources/LogicProMCP/Channels/AppleScriptChannel.swift
@@ -479,12 +479,15 @@ actor AppleScriptChannel: Channel {
         await runtime.runScript(source)
     }
 
-    static func executeAppleScript(_ source: String) async -> ChannelResult {
+    static func executeAppleScript(
+        _ source: String,
+        timeout: TimeInterval = ServerConfig.appleScriptTimeout
+    ) async -> ChannelResult {
         await Task.detached(priority: .userInitiated) {
             let execution = BoundedProcessRunner.run(
                 executable: "/usr/bin/osascript",
                 arguments: appleScriptArguments(for: source),
-                timeout: ServerConfig.appleScriptTimeout,
+                timeout: timeout,
                 outputLimitBytes: 128 * 1024
             )
             let result = channelResult(forAppleScriptExecution: execution)
--- a/Sources/LogicProMCP/Channels/ChannelRouter.swift
+++ b/Sources/LogicProMCP/Channels/ChannelRouter.swift
@@ -22,13 +22,33 @@ actor ChannelRouter {
 
     private var channels: [ChannelID: any Channel] = [:]
 
-    private static let transportToggleOpsAllowingAXElementFallback: Set<String> = [
+    // Operations for which an Accessibility `element_not_found` is provably a
+    // PRE-WRITE miss, so a later channel may still run. Three conditions must all
+    // hold before an op belongs here, and #460 is why they are written down rather
+    // than left implicit:
+    //
+    //  1. `element_not_found` means the AX control was never located, so no
+    //     actuation reached Logic and no partial write can be stranded.
+    //  2. The failure is Accessibility-specific — a missing control-bar button or
+    //     ruler element — not a rejection of the request itself.
+    //  3. Every later candidate expresses the same semantics, so falling through
+    //     cannot silently change what the caller asked for.
+    //
+    // `transport.goto_position` satisfies all three exactly as the toggles do: a
+    // missing ruler element moves nothing, and MCU/CoreMIDI/CGEvent all position
+    // the playhead. Omitting it stranded a real default-install workflow whose
+    // first action was `goto_position` (Discussion #455).
+    //
+    // An op that can leave a partial write, or whose later candidates differ in
+    // meaning, must NOT be added here; it needs per-op recovery instead.
+    private static let opsAllowingAXElementNotFoundFallback: Set<String> = [
         "transport.play",
         "transport.stop",
         "transport.record",
         "transport.toggle_cycle",
         "transport.toggle_metronome",
         "transport.toggle_count_in",
+        "transport.goto_position",
     ]
 
     /// Active routing table (v2). `internal` so test-targets can introspect
@@ -258,7 +278,7 @@ actor ChannelRouter {
         message: String
     ) -> Bool {
         channelID == .accessibility &&
-            transportToggleOpsAllowingAXElementFallback.contains(operation) &&
+            opsAllowingAXElementNotFoundFallback.contains(operation) &&
             HonestContract.stateCErrorCode(message) == HonestContract.FailureError.elementNotFound.rawValue
     }
 
--- a/Sources/LogicProMCP/Server/OperationTrace.swift
+++ b/Sources/LogicProMCP/Server/OperationTrace.swift
@@ -130,6 +130,21 @@ final class OperationTraceContext: @unchecked Sendable {
     }
 }
 
+/// #452: render a monotonic interval as integer milliseconds for a trace
+/// attribute. `ContinuousClock` is used rather than a `Date` difference because
+/// wall-clock can step backwards across an NTP correction, which would report a
+/// segment as negative or absurdly long — and a timing number that a clock
+/// adjustment can invent is not evidence.
+///
+/// Clamped at zero: the clock is non-decreasing, so a negative reading means the
+/// measurement is unusable, not that the work was fast, and emitting it as a
+/// duration would let a consumer average it into nonsense.
+func elapsedMilliseconds(since start: ContinuousClock.Instant) -> String {
+    let components = (ContinuousClock.now - start).components
+    let milliseconds = components.seconds * 1000 + components.attoseconds / 1_000_000_000_000_000
+    return String(max(0, milliseconds))
+}
+
 struct TraceID: RawRepresentable, Codable, Sendable, Hashable, CustomStringConvertible {
     let rawValue: String
 
@@ -171,6 +186,13 @@ enum TracePhase: String, Codable, Sendable, CaseIterable {
     /// phase's contract is "a channel began executing this write", not "this
     /// exact channel wrote".
     case writeBoundaryCrossed = "write_boundary.crossed"
+    /// #452 — a bounded AppleScript execution finished inside a channel, carrying
+    /// how long it actually took. `channel.started`/`channel.completed` bracket
+    /// the whole channel call, so the segment governed by the osascript timeout
+    /// was not separable from the surrounding Swift AX work; #449 could be
+    /// argued about but not measured. This phase is diagnostic and additive:
+    /// nothing gates on it, and it never replaces the channel bracket.
+    case scriptSegmentCompleted = "script_segment.completed"
     case channelCompleted = "channel.completed"
     case verificationPoll = "verification.poll"
     case verificationCompleted = "verification.completed"
@@ -223,6 +245,12 @@ actor OperationTraceStore {
         "outcome": .publicDiagnostic,
         "parent_trace_id": .publicDiagnostic,
         "gate_mode": .publicDiagnostic,
+        // #452 — measured elapsed time of a bounded AppleScript segment.
+        // Undeclared keys are DROPPED by `filteredAttributes`, so an emission
+        // without this row would record nothing and expose nothing. The bound it
+        // ran under is deliberately NOT emitted here: the executor is injected,
+        // so the call site cannot attest to a timeout it does not own.
+        "applescript_duration_ms": .publicDiagnostic,
     ]
 
     let maximumTraceCount: Int
--- a/Sources/LogicProMCP/Server/ServerConfig.swift
+++ b/Sources/LogicProMCP/Server/ServerConfig.swift
@@ -20,6 +20,15 @@ struct ServerConfig: Sendable {
     // MARK: - Timeouts
     static let appleScriptTimeout: TimeInterval = 5.0
 
+    // midi.import_file drives a File > Import > MIDI File sheet, a path-entry
+    // dialog, an import button and a tempo prompt, each polled inside the
+    // script itself. Summing one delay per iteration across those loops gives a
+    // floor of 17.2 s, so the shared 5.0 s bound killed osascript while Logic
+    // was still importing — the region landed but the caller saw a timeout
+    // (#449). This bound must stay above that floor; the polling loops in
+    // AccessibilityChannel+MIDIImport are the thing it has to outlast.
+    static let midiImportAppleScriptTimeout: TimeInterval = 30.0
+
     // MARK: - Logic Pro
     /// Resolved bundle ID for the active Logic Pro variant (desktop or Creator Studio).
     static var logicProBundleID: String { LogicProTarget.current.bundleID }
--- a/Sources/LogicProMCP/Utilities/SetupDoctor+ChannelDependencyChecks.swift
+++ b/Sources/LogicProMCP/Utilities/SetupDoctor+ChannelDependencyChecks.swift
@@ -89,8 +89,23 @@ extension SetupDoctor {
                 : "Key Commands preset is not staged; ignore this if MIDIKeyCommands-only ops are unused.",
             evidence: ["preset_staged": String(staged)],
             remediationType: staged ? .none : .command,
-            remediationValueOverride: staged ? nil : "install-keycmds.sh"
+            remediationValueOverride: staged ? nil : keyCommandsRemediation(runtime: runtime)
         )
+    }
+
+    // #456: Homebrew installs install-keycmds.sh into the formula pkgshare, not on
+    // PATH, so printing the bare script name hands the user a command they cannot
+    // run — and fetching the script alone fails on its sibling keycmd-preset.plist.
+    // The share dir is already resolved for the install checks, so the remediation
+    // is derived from that same probe and stays runnable wherever the package put
+    // it. Only an unresolved share dir falls back to the bare name, and it says so.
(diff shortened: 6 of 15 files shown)


MongLong0214:
Closing without merging.

sequence forbids.

PR #463, opened 2026-08-07
fix(#453): bind alert acknowledgement to the dialog that was classified

### LINKED ISSUE #453: Fail closed when the accessibility alert target changes or button resolution fails
## Problem

`ModalReconciliation.classify` decides that a top-level `AXDialog` alert is safe to
acknowledge only when it carries **exactly one** titled button — two or more means a
choice, and the classifier fails closed by returning `.unknownSheet`.

That decision is not carried into the click. `acknowledgeTopLevelAlert` in
`Sources/LogicProMCP/Channels/AccessibilityChannel+ModalReconcile.swift` re-resolves the
window and clicks without re-checking anything:

```applescript
set alertWindow to first window whose subrole is "AXDialog"
try
    click button "<title>" of alertWindow
on error
    click button 1 of alertWindow
end try
```

Three separate gaps:

1. **The clicked window is not proven to be the classified window.** `first window whose
   subrole is "AXDialog"` is evaluated at click time. If a different alert has come
   forward in between, this clicks the one that was never classified.
2. **The button count is not re-checked.** The classifier's one-button precondition is
   verified once, before the click, and never again.
3. **Title lookup failure falls back to `click button 1`.** That fallback runs with no
   count check and no window identity check, so a dialog with two or more buttons is
   clicked anyway whenever the title lookup fails.

The result is that the classifier's fail-closed decision can be bypassed by the executor.
Gap 3 is the sharpest: on that path the safety discriminator does not participate at all.

## Evidence

Static reading of the current source. Not yet reproduced live — the timing window in
gaps 1 and 2 has not been observed on a running Logic, and the issue is filed on the
static path alone because gap 3 needs no timing to occur.

## Required scope

1. Bind the classified alert to the clicked alert — the same window, proven, not
   re-resolved by predicate.
2. Re-check the count of top-level actionable buttons immediately before the click.
3. If it is not exactly one, do not click. Fail closed.
4. Remove the `click button 1` fallback.
5. No click when the dialog is replaced, when several dialogs are present, when the
   element is stale, when the count is zero or two or more, or when the lookup errors.
6. On the success path, click only the verified single button of the same alert.
7. Record the reason for a refusal as an explicit error or trace value, without
   publishing dialog text, button titles or other UI content.
8. RED coverage for a state change between classification and click, and for a
   wrong-target click.
9. focused, full, release and manual Logic QA.
10. Evidence is invalidated in full if the base commit changes.

## Out of scope

- Sheet-path behaviour (`sheetPresent`) — a different rule, unchanged here.
- Any global timeout value.
- Duration instrumentation for the AppleScript segment (separate issue).
- Adding titles to an allowlist while keeping the fallback. That leaves the fail-open
  path in place and only narrows how often it is reached.
- Retrying against another window, or against the first button, when verification fails.

MongLong0214:
`ModalReconciliation.classify` treats a top-level `AXDialog` as safe to acknowledge only when it carries **exactly one** titled button — two or more is a choice, and choices are never auto-answered.

That decision did not reach the click.

### The three gaps

`acknowledgeTopLevelAlert` re-resolved the dialog in AppleScript and clicked:

```applescript
set alertWindow to first window whose subrole is "AXDialog"
try
    click button "<title>" of alertWindow
on error
    click button 1 of alertWindow
end try
```

1. **The clicked window was not proven to be the classified window.** The predicate was evaluated at click time, so a dialog that came forward in between was clicked though it had never been classified.
2. **The button count was never re-checked** — verified once, before the click.
3. **A failed title lookup fell through to `click button 1`** with no count check and no identity check. On that path the safety discriminator did not participate at all. This is the sharpest of the three: it needs no timing to occur.

### The fix

The executor is now native AX and acts on the **element** the classifier read:

- `blockingDialogTarget` returns the dialog element and its titled buttons; `blockingDialogInfo` is derived from it, so there is one inventory rather than two that can disagree.
- The current blocking dialog is re-resolved and must be `CFEqual` to the classified one — a replacement or a second dialog fails identity instead of being clicked.
- The button set is re-read from that element immediately before pressing and must still be exactly one.
- The first-button fallback is **gone**. No path presses a control the single-button rule did not authorize.
- Refusals are typed and reported through extras as `reconcile_refused`. The values are fixed structural tokens and never carry dialog text or button titles, because they travel into extras and logs.

Refusal is the safe direction: an unacknowledged alert leaves the operator with a visible dialog, while a wrong click answers a question on their behalf and cannot be undone.

### Testing

The tests drive the real `reconcilePreflight` entry point against a fake AX tree that **changes between the classify read and the click read** — without that, both reads see identical state and the test could not fail no matter what the executor did.

How many AX reads classification performs is an implementation detail (`mainWindow`, the dialog filter and the button scan each read). Hard-coding "switch after the first read" would encode a guess that rots the moment a read is added, and a stale guess makes the test pass for the wrong reason. So the fixture **calibrates**: a read-only pass counts the reads classification actually performs, and the flip is placed immediately after that boundary.

Verified by mutation, in both directions:

| Mutation | Result |
|---|---|
| Remove the identity check | The swap test presses the **substitute** dialog's button — gap 1 reproduced |
| Restore "press whatever button is first" | The growth test presses **"Save"** on a two-button dialog — gap 3 reproduced |

The release path is asserted too, and names *which* element was pressed: a gate that only ever refuses would block the server on every genuine alert.

Full suite `--no-parallel`: 3346 tests. The single failure — the library-inventory panel snapshot — reproduces identically on the unmodified tree and is unrelated to this branch.

### Not yet covered

**Live Logic QA.** The timing window in gaps 1 and 2 has not been observed on a running Logic; this change is derived from the static path, which is also where gap 3 lives. Required scope item 9 is outstanding.

Closes #453

### DIFF
--- a/Sources/LogicProMCP/Accessibility/AXLogicProElements.swift
+++ b/Sources/LogicProMCP/Accessibility/AXLogicProElements.swift
@@ -110,7 +110,23 @@ enum AXLogicProElements {
         let recoveryAction: String
     }
 
-    static func blockingDialogInfo(runtime: Runtime = .production) -> BlockingDialogInfo? {
+    /// #453: the blocking dialog together with the ELEMENT it was read from and
+    /// its titled button elements.
+    ///
+    /// `BlockingDialogInfo` is `Sendable` and carries only strings, which is what
+    /// makes it safe to publish as diagnostics — but it also means a caller that
+    /// decides something from it has to find the dialog again to act, and "find
+    /// the first AXDialog" evaluated a second time can resolve a DIFFERENT window.
+    /// A caller that must act on precisely the window it classified takes this
+    /// instead and holds the element, so identity is the element itself rather
+    /// than a predicate re-run against whatever is frontmost at click time.
+    struct BlockingDialogTarget {
+        let element: AXUIElement
+        let buttons: [(element: AXUIElement, title: String)]
+        let info: BlockingDialogInfo
+    }
+
+    static func blockingDialogTarget(runtime: Runtime = .production) -> BlockingDialogTarget? {
         guard let app = appRoot(runtime: runtime) else { return nil }
         let windows: [AXUIElement] = AXHelpers.getAttribute(
             app, kAXWindowsAttribute, runtime: runtime.ax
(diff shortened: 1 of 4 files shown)


MongLong0214:
Closing without merging.

sequence forbids.

PR #464, opened 2026-08-07
fix(#461): derive capability readiness from operation routes, not tagged checks

### LINKED ISSUE #461: Doctor can report core_transport ready without a runnable path for every core operation
## Summary

`LogicProMCP doctor --profile core` can report `core_transport=ready` without proving that every core transport operation has a runnable route. In the reported default-install workflow, the first agent action was `goto_position`; it failed even though doctor had presented the non-optional/core setup as ready.

Profiles currently scope doctor checks, but agents remain free to call any advertised tool command. A capability-group label therefore overstates readiness when it is not derived from the actual operation routes and current channel support.

Reported in Discussion #455: https://github.com/MongLong0214/logic-pro-mcp/discussions/455#discussioncomment-17907427

## Exact local reproduction

Reproduced on current `main` at `a429df72fc8bdb9c8fb5d3fdb3d4bd4aa4aea7cc`.

Running the exact-head debug binary produced:

```json
{
  "report_status": "degraded",
  "core_transport": {
    "status": "ready",
    "checks": [
      "binary.path",
      "binary.executable",
      "permissions.accessibility",
      "permissions.post_event_access",
      "logic.installation",
      "logic.version_support",
      "logic.application_state"
    ]
  }
}
```

The focused exact-head doctor/router run also passed 3/3 tests, including the current `core ready` rendering and both terminal/fallback router branches.

The result is internally consistent with the current implementation but product-level incorrect:

- `SetupDoctor.capabilityDefinitions` declares `core_transport` without live verification.
- `capabilityCheckIDs` only gathers static check IDs tagged with that group.
- `capabilities(for:)` marks the group ready when those statuses pass.
- No public transport operation list, route candidate, per-channel handler/parameter support, or live route plan participates in the decision.
- As tracked in #460, `transport.goto_position` can stop on AX `element_not_found` before reaching its declared CGEvent fallback.

The overall report happened to be `degraded` for unrelated setup state; the defect is the independent `core_transport=ready` claim.

## Root cause

Doctor and runtime routing use separate abstractions:

- Doctor: broad static capability groups -> tagged environment checks -> aggregate status.
- Runtime: operation string -> ordered channel list -> health gate -> channel execution -> global error-code fallback rule.

There is no shared source of truth connecting a public tool command to:

- its internal operation;
- valid parameter domains;
- candidate channels that actually implement those semantics;
- prerequisites and current health for each candidate;
- verification strength (verified readback versus send-only/unverified);
- safe fallback behavior.

Adding a one-off `goto_position` check to `core_transport` would leave every other operation and future route vulnerable to the same drift.

## Similar-scenario classes that must be covered

1. A capability group is ready while one member operation has no executable candidate.
2. A channel is generally healthy but does not implement the operation or its requested parameter form.
3. The only reachable route is send-only State B, but doctor presents it as fully ready rather than live-verification-required.
4. A manual approval is present while its actual prerequisite or runtime channel is missing/stale (related to #457).
5. An optional channel was excluded by profile, but an agent selects an operation that depends on it.
6. Health changes after doctor runs; the runtime must re-plan instead of trusting stale readiness.
7. A declared fallback is unreachable because failure classification suppresses it (#460).

## Required root solution

1. Introduce one shared `OperationCapabilityRegistry` consumed by tool dispatch, `RouteSpec`/router, health reporting, and doctor.
2. Every advertised `tool + command` must map to one internal operation and declare:
   - capability group and baseline/optional status;
   - parameter domains;
   - ordered channel candidates and handler support;
   - prerequisites/approvals;
   - verification strength and safe fallback policy.
3. Add a read-only route planner used by both doctor and runtime. Given an operation, parameters, and a health snapshot, it returns executable, skipped, blocked, and manual candidates with reasons.
4. Compute per-operation readiness first, then aggregate capability readiness. Profiles may filter presentation, but they must not manufacture readiness for operations the advertised product can invoke.
5. Distinguish at least `ready_verified`, `ready_degraded`, `unknown_live_verify_required`, `manual_setup_required`, and `not_ready` at operation level. Keep the public schema additive/versioned.
6. Show feature/operation availability and required channels in onboarding output. Channel-centric setup information should explain which operations become available, while capability summaries remain derived views.
7. Re-evaluate the plan at runtime; doctor output is a timestamped snapshot, not authorization to bypass current health.

## Acceptance criteria

- [ ] A structural invariant accounts for every advertised tool command exactly once and fails on an unmapped operation, undeclared parameter domain, or route destination without handler support.
- [ ] Doctor and runtime return the same candidate/blocked route plan for identical operation, parameters, approvals, and health snapshot.
- [ ] RED regression: all current static `core_transport` checks pass, but AX goto controls are unavailable and no validated alternate exists; `goto_position` is not ready and `core_transport` cannot be reported ready.
- [ ] When AX is unavailable and a safe, runtime-ready, semantically equivalent fallback exists, doctor reports the exact fallback and the appropriate verified/degraded status.
- [ ] A send-only CGEvent route is never promoted to verified readiness without live readback; its State B semantics remain explicit.
- [ ] Invalid parameter forms are reported as request-invalid, not as missing channel readiness.
- [ ] Stale/missing manual prerequisites cannot be hidden by an approval record; #457 ordering and attestation truth remain enforced.
- [ ] Default human and JSON output list unavailable/incomplete operations and the channel/setup action that would enable each one.
- [ ] Tests cover profile filtering versus agent-selected out-of-profile commands, health drift between doctor and call time, and all operation-readiness status classes.
- [ ] The registry/fallback work from #460 is reused; no second routing or capability inventory is introduced.
- [ ] Focused tests, full suite, release build/artifact provenance, exact-artifact live QA, exact-head review, and CI pass before merge.

## Related

- #460 (typed operation-aware fallback contract and `goto_position` regression)
- #457 (manual approval must follow real setup prerequisites)
- #456 (install-source-aware remediation; separate but feeds truthful prerequisite state)
- Discussion #455 follow-up: https://github.com/MongLong0214/logic-pro-mcp/discussions/455#discussioncomment-17907427

MongLong0214:
`doctor --profile core` reported `core_transport: ready` while `transport.goto_position` had no runnable path. The group's verdict came from environment checks tagged with its label, and **nothing joined those checks to the operations the group advertises** — so a new user's first action failed against a setup doctor had just called ready.

Adding a `goto_position` check to `core_transport` would leave every other operation and every future route exposed to the same drift. The join is made once instead.

### The shared registry

`OperationCapabilityRegistry` is now the single source of truth:

- **`capabilityOperations`** — which operations each group advertises. Membership is explicit rather than inferred from the name prefix, because `transport.get_state` is a read that a broken *write* path does not make unavailable.
- **`routeCandidates`** — reads the **router's own table**. A second inventory that can drift is the failure this registry exists to remove.
- **`channelPrerequisites`** — the checks a channel needs before it can run anything.
- **`verificationStrength`** — separates channels that read back from channels that only actuate, so a send-only route can never be published as verified.

`plan(operation:health:)` is pure and returns per-candidate executable/blocked with a structural reason, plus an operation readiness of `ready_verified` / `ready_degraded` / `unknown_live_verify_required` / `manual_setup_required` / `not_ready`.

The group verdict is the **weakest** of its operations — a group is only as ready as the least-ready thing it lets an agent call, and averaging would reproduce the original bug on a single broken operation.

Two distinctions carry the correctness:

- A check that **did not run** is absent from the snapshot rather than recorded as `false`, so the planner can tell *"we did not look"* from *"it does not work"*.
- `manual` and `skipped` become **awaiting** rather than false. A held approval (#457) or an unstaged preset is a step the operator has not taken, not a broken machine.

Doctor takes the weaker of its check verdict and the route verdict, so a route gap can only *lower* readiness and a passing check set can never manufacture readiness the routes do not support. Per-operation rows are published alongside the group status, naming the blocked operation and the prerequisite that would enable it. The group `status` keeps its existing four values, so the schema change is additive.

### The load-bearing case

An earlier revision of the doctor test **could not tell the fix from the old behaviour** — a failed-permission fixture reaches `not_ready` through the check path alone, so the mutation survived. That test was replaced with the case where the two layers genuinely disagree:

> With `permissions.accessibility` merely **awaiting** approval, the check layer reaches `unknown_live_verify_required` — nothing is broken, something is unconfirmed. But `transport.set_tempo` routes through Accessibility and **nothing else**, so it has no candidate at all and the group is **not ready**. Calling it "verify live" would invite an agent to try.

That is #461's class 4: an approval present while its actual prerequisite is not.

### Mutations

| Mutation | Killed by |
|---|---|
| Ignore the route plan in doctor | the disagreement case above |
| Aggregate to the strongest member instead of the weakest | same |
| Treat an awaiting check as satisfied | same |

Structural invariants are pinned too: every advertised operation has ≥1 route candidate, planner candidates equal the router's table exactly, and every reachable channel declares a prerequisite entry — a missing entry would read as "no prerequisites" and mark an unusable channel executable.

Full suite `--no-parallel`: one pre-existing failure (the library-inventory panel snapshot) that reproduces identically on the unmodified tree.

### Deliberately not in scope

Parameter-domain validation, runtime re-planning between doctor time and call time, and per-tool-command census coupling. Each needs its own design; the issue tracks them. This PR closes the readiness-vs-route gap that produced the report.

Closes #461

### DIFF
--- a/Sources/LogicProMCP/Utilities/OperationCapabilityRegistry.swift
+++ b/Sources/LogicProMCP/Utilities/OperationCapabilityRegistry.swift
@@ -1,0 +1,234 @@
+import Foundation
+
+/// #461 — the shared source of truth linking a doctor capability GROUP to the
+/// operations it claims, and each operation to the channels that can actually
+/// run it.
+///
+/// Before this, doctor and the runtime reasoned about readiness through two
+/// unconnected abstractions. Doctor aggregated environment checks tagged with a
+/// group label; the runtime resolved an operation string to an ordered channel
+/// list and executed. Nothing joined them, so `core_transport` could report
+/// `ready` on the strength of "Accessibility permission is granted, Logic is
+/// running" while `transport.goto_position` had no runnable path at all — which
+/// is exactly what a new user hit as their first action.
+///
+/// The fix is not a `goto_position` special case. A capability group is READY
+/// only when every operation it advertises has a runnable route, and both the
+/// group membership and the route candidates are declared here, in one place,
+/// derived from the routing table the router itself executes.
+enum OperationCapabilityRegistry {
+
+    // MARK: - Channel prerequisites
+
+    /// What a channel needs from the environment before it can run anything.
+    ///
+    /// These are the doctor check IDs whose failure makes the channel unusable.
+    /// Keeping the mapping here — rather than inside doctor's rendering — is what
+    /// lets a route plan be computed from a health snapshot without doctor.
+    static let channelPrerequisites: [ChannelID: [String]] = [
+        .accessibility: ["permissions.accessibility", "logic.application_state"],
+        .cgEvent: ["permissions.post_event_access", "logic.application_state"],
+        .appleScript: ["permissions.automation_logic_pro", "logic.application_state"],
+        // Key Commands actuation rides CoreMIDI and needs the staged preset; the
+        // staging check is the one that told users to approve before setup (#457).
+        .midiKeyCommands: ["channels.keycmd_reference"],
+        // MCU and CoreMIDI have no environment check that can PROVE they work —
+        // Logic-side control-surface wiring is not observable from outside. They
+        // are therefore never a basis for a verified-ready claim; see
+        // `verificationStrength`.
+        .mcu: [],
+        .coreMIDI: [],
+        .scripter: [],
+    ]
+
+    /// How strongly a channel's success can be believed.
+    ///
+    /// A send-only channel actuates and cannot read back, so a route whose only
+    /// reachable candidate is send-only must never be published as verified —
+    /// that is the difference between "this will work" and "this will be
+    /// attempted". Doctor downgrades such a route to a live-verify claim.
+    enum VerificationStrength: String, Sendable {
+        /// Writes and independently reads the result back.
+        case verified
+        /// Actuates without readback. Honest Contract State B territory.
+        case sendOnly
+    }
+
+    static func verificationStrength(of channel: ChannelID) -> VerificationStrength {
+        switch channel {
+        case .accessibility, .appleScript:
+            return .verified
+        case .cgEvent, .mcu, .coreMIDI, .midiKeyCommands, .scripter:
+            return .sendOnly
+        }
+    }
+
+    // MARK: - Capability membership
+
+    /// Operations each capability group advertises.
+    ///
+    /// Membership is explicit rather than inferred from the operation-name prefix:
+    /// `transport.get_state` is a read that a broken write path does not make
+    /// unavailable, and grouping by prefix would drag it into the same verdict.
+    /// An operation listed here is one an agent can invoke once the group is
+    /// presented as ready, so listing it is a promise.
+    static let capabilityOperations: [String: [String]] = [
+        "core_transport": [
+            "transport.play",
+            "transport.stop",
+            "transport.record",
+            "transport.goto_position",
+            "transport.get_state",
+            "transport.toggle_cycle",
+            "transport.toggle_metronome",
+            "transport.set_tempo",
+        ],
+        "track_management": [
+            "track.get_tracks",
+            "track.select",
+            "track.create_audio",
+            "track.create_instrument",
+            "track.rename",
+            "track.delete",
+        ],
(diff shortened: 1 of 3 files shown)


MongLong0214:
Closing without merging.

sequence forbids.

PR #466, opened 2026-08-08
test(library): give the category check an oracle that can fail

MongLong0214:
## The defect

`LibraryDiskScannerTests` had a check named *"every emitted category exists in Resources/library-inventory.json Panel snapshot"*. It read its oracle from `Resources/library-inventory.json` — a path matched by `.gitignore:29`, so it can never be the committed snapshot the test's own comment described.

That produced two failures at once:

- **On CI the file is absent**, the body returned early, and the test asserted nothing for its entire life.
- **Locally it resolved to whatever the last AX probe left behind.** On this machine that was a seven-node, two-category artifact (`source: ax`, generated 2026-08-07), so the full suite was **red** against the real Library with 15 issues — all from this one test.

The check could not have caught what it was named for in any case. `mapDiskPathToPanel` returns `nil` for an unmapped path and the caller drops it, so no raw disk name is ever emitted. The subset property it asserted holds **vacuously**.

## The replacement

Three checks, and the mapping range is now pinned in a file that can actually be committed.

| check | what it catches | runs on CI |
|---|---|---|
| `mappingTableRangeEqualsPanelInventory` | a table edit inventing a category `selectPath` could never navigate to, or a Panel category becoming unreachable | yes |
| `newSubTaxonomyIsReportedNotSilentlyDropped` | the real drift case — Logic ships a new sub-taxonomy under a known parent and it is dropped; asserts the drop is counted and reported, not silent | yes |
| `installedLibraryEmitsOnlyPanelCategories` | the subset property on a real Library | skipped via `.enabled(if:)` |

The environment condition moved from an early `return` to an `.enabled(if:)` trait, so an absent Library is reported by the framework as a **skip** rather than as a body that returns early and reads as a pass.

The source comment claiming the table is *"Verified against `Resources/library-inventory.json`"* is corrected — it names the test that actually pins it, and says why that path cannot be the source.

## Mutation evidence

All four assertions were mutated and observed to fail, so none of them is green by construction:

| mutation | result |
|---|---|
| add `"Woodwind/Flutes": "Flutes"` to `diskToPanel` | `mappingTableRangeEqualsPanelInventory` **FAIL** |
| let an unmapped taxonomy leak through under its raw disk name | `newSubTaxonomyIsReportedNotSilentlyDropped` **FAIL** (3 issues) |
| change the recorded reason off `no_panel_taxonomy_route` | `newSubTaxonomyIsReportedNotSilentlyDropped` **FAIL** |
| drop `Studio Horns` from the oracle | `installedLibraryEmitsOnlyPanelCategories` **FAIL** |

## Verification

- Full suite: **`Test run with 3371 tests passed`, exit 0.** It was **red on `main`** with 15 issues in this one test.
- `Scripts/ci-forbid-dead-expect.sh`: clean.
- No production behaviour changes; the only `Sources` edit is a doc comment.

### DIFF
--- a/Sources/LogicProMCP/Accessibility/LibraryDiskScanner.swift
+++ b/Sources/LogicProMCP/Accessibility/LibraryDiskScanner.swift
@@ -69,7 +69,11 @@ enum LibraryDiskScanner {
     /// 1-segment Panel path `Acoustic Drums/...` before the fallback
     /// `Drums & Percussion` → (no-match) check runs.
     ///
-    /// Verified against `Resources/library-inventory.json` (AX panel snapshot).
+    /// The combined range of this table and `identityCategories` is pinned
+    /// against the v3.0.4 Panel inventory by `LibraryDiskScannerTests`. It is
+    /// not verified against `Resources/library-inventory.json`: that path is
+    /// matched by `.gitignore`, so it is absent on CI and is a stale local AX
+    /// probe artifact anywhere else.
     /// Panel categories: Bass, Acoustic Drums, Electronic Drums, Percussion,
     /// Guitar, Acoustic Piano, Clavinet, Electric Piano, Mellotron, Organ,
     /// Mallet, Studio Horns, Studio Strings, Synthesizer, Orchestral, World.
--- a/Tests/LogicProMCPTests/LibraryDiskScannerTests.swift
+++ b/Tests/LogicProMCPTests/LibraryDiskScannerTests.swift
@@ -233,14 +233,9 @@ struct LibraryDiskScannerTests {
     /// on the machine. Verifies that a full scan returns a clinically
     /// implausible-low count (e.g. under 1000 leaves) does NOT ship — the
     /// whole point of v3.0.5 is to fix the 345-leaf undercount.
-    @Test("local-machine integration: factory Library reports at least 1000 leaves when present")
+    @Test("local-machine integration: factory Library reports at least 1000 leaves when present",
+          .enabled(if: installedLogicLibraryExists))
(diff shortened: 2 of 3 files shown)


MongLong0214:
## Second commit — the same defect class, swept

Three more tests implemented an environment condition as an early `return` inside the body, so on a host without the Logic install they were **reported as passing while asserting nothing**:

- `scanLocalFactoryLibraryReportsFullCoverage`
- `testProcessUtilsBundleURLResolvesWithoutAppKitRunloop`
- `testProcessUtilsLogicProVersionResolvesOffMain`

Each moves to `.enabled(if:)`. Pointing both environment probes at absent paths now produces four explicit `skipped` lines from the framework instead of four silent passes — that is the mutation evidence for this commit.

A sweep of `Tests/` for the shape — a guard on `fileExists`/`isRunning`/`ProcessInfo`/`environment` whose `else` returns, on a `@Test` without `.enabled(if:)` — now reports **zero remaining**.

### Preflight note

The public-surface preflight reports **C1b** against two added lines:

```
+@Test(codexSeatbeltProcessInspectionDisabled, .enabled(if: logicProInstalled))
```

That identifier is **pre-existing on `main`** (`CGEventChannelTests`, `EndToEndTests`, and others). This branch adds 2 occurrences and removes 2 — **net zero**; the lines are flagged only because the trait list on those two lines was edited. C1b matches added lines without checking whether the token is new.

I have not touched the preflight script to clear my own blocker. Flagging the asymmetry instead: the identifier is genuine pre-existing internal-tool naming in the public test surface, and renaming it repo-wide is a separate change with its own scope, not something to fold into this one.

All other checks — C1a, C1c, C2, C3a, C3b — are clean.

Issue #476, opened 2026-08-09
logic://tracks returns data:[] before its first fetch, without the unreadability marker its sibling resource carries


MongLong0214:
Found while live-verifying the merged PRs against a project that demonstrably has two tracks
(`Sum 1`, `Studio Grand` — both visible in the mixer, and `logic_tracks select {"track":0}` returns
State A against them).

Read verbatim from the release binary, in order, in one session:

```
1) cold read         {"cache_age_sec":null,"fetched_at":null,"ax_occluded":false,"source":"default","data":[]}
2) after refresh_cache {"cache_age_sec":0.55,"fetched_at":"2026-08-09T11:48:17.812Z","ax_occluded":false,"source":"ax_live","data":[{…real tracks…}]}
3) after a track op    same as (2), cache_age 1.29
4) immediate re-read   same as (2), cache_age 1.30
```

So before the first fetch the resource answers `data: []` with `source: "default"` and
`fetched_at: null`. The information that nothing was read is present, but only implicitly — a
consumer that reads `data` sees a project with no tracks.

**The sibling resource does this properly.** `logic://markers`, in the same session, answered:

```
{"cache_age_sec":null,"fetched_at":null,"ax_occluded":false,"readable":false,
 "reason":"marker_list_not_open","source":"unreadable","verified_empty":false,"data":[]}
```

`readable: false`, an explicit `reason`, and `verified_empty: false` — it refuses to let an empty
array be mistaken for an observation. `logic://tracks` carries none of those three fields.

This is a consistency gap rather than a false verification: nothing claims to have verified an empty
project. But two resources in the same server express "I have not looked" in two different ways, and
one of them is easy to misread. Making `logic://tracks` carry `readable` / `verified_empty` (and a
`reason` when it has not fetched) would close it.

Worth noting alongside: `doctor --profile core` advertises `track.get_tracks` as `ready_verified`,
and its only surface is this resource, so a caller following the doctor's own claim can land on the
cold-read shape.

MongLong0214:
**Post-merge live verification.** PR #484 merged as `b263e1b2`. Release artifact rebuilt from `main` @ `688baf64` and driven over real stdio MCP against a running Logic.

Before any live read has happened:

```json
{"readable": false, "verified_empty": false, "reason": "no_live_track_read_yet", "source": "default", "data": []}
```

After a live read succeeds:

```
readable: true, source: ax_live, 7 tracks
```

[...]

Issue #475, opened 2026-08-09
Verified-write authorisation is not revalidated across subsystem boundaries


MongLong0214:
Collected from an adversarial audit of the plug-in insert path. None of these belong to #425 — they
sit in other subsystems or are pre-existing — but they share one shape worth fixing together: an
authorisation is established once and then acted on later against a tree that may have moved.

**Project and track identity are checked once.** `insert_verified` validates the project path and the
target track name up front, then re-resolves and actuates the current AX tree without revalidating
either. Switching project or reordering tracks during the intervening awaits lets the same numeric
slot in a different project be actuated under the first project's authorisation.

**Unreadable strips and slots are dropped, and the survivors are renumbered.**
`AXLogicProElements+Mixer.swift` and `+PluginSlots.swift` filter out structurally unreadable entries,
so physical index 1 can be reported as index 0. A request for slot 0 can therefore actuate physical
slot 1, and if the dropout persists through the readback the receipt reports success for slot 0.

**The track-selection ladder presses arbitrary children.** `AXLogicProElements+Tracks.swift` falls
back to pressing each child of a track header without proving the child is a selection control; the
first pressable child can be Mute.

**The Go-to-Position dialog classifier matches on a substring of the window title** and then presses
that window's close button, with no dialog-role or modal check. A project named so its title contains
that phrase would have its main window closed.

**Rollback attributes any concurrent strip delta to this attempt** and sends a global Undo without
checking undo-stack ownership, so a plug-in the user inserted elsewhere on the strip during the window
can be undone. Separately, an inventory that becomes unreadable during rollback polling reads as an
empty dictionary, which `strayRemovalConfirmed` treats as confirmed removal.

Two further items are inherent limits rather than defects, and are now recorded in
`docs/tickets/issue-425-coord-free-insert.md`: product identity is a display string, and acceptance is
a diff rather than a transaction, so a concurrent identical mount at the requested slot is
indistinguishable from our own.

MongLong0214:
All five items in this issue are on `main`. Closing with the commit for each.

| item | commit | what changed |
| --- | --- | --- |
| project/track identity checked once | `3eb4360b` | an insert is not certified into a document we can no longer name |
| unreadable strips renumbered | `cca56b20` | a strip list that may have been renumbered is refused, not silently compacted |
| selection ladder presses arbitrary children | `72deda76` | selection goes through `AXSelectedChildren` on the header group; it can no longer land on Mute |
[...]

PR #481, opened 2026-08-09
fix(#489): expose fresh exact project identity

### LINKED ISSUE #489: logic://project/info discards the fresh project identity a client needs to bind an exact write
`logic://project/info` cannot establish a fresh exact-project write boundary, because it drops the independently read current-document `filePath` whenever the AX project cache is fresh, and emits `fetched_at: null` when file metadata is the source.

## Why this matters

A client that must bind a write to an exact project has no reliable way to learn which `.logicx` bundle is current, and no timestamp telling it how old that answer is. `LogicProjectFileReader` has already validated the current bundle by the time the resource answers, so the information exists and is discarded.

## Acceptance

- the current document identity is acquired on every project resource request, independently of AX cache freshness
- the validated project-file path populates `filePath` whether or not the AX cache is fresh
- `identity_source` and `identity_fetched_at` are exposed so a client can tell where the answer came from and how old it is
- a project-file read emits a non-null resource `fetched_at`
- regression coverage for both the cache-live and the project-file source
- verified against the release artifact with the running application, not only in unit tests

## Status

Work exists on `fix/project-resource-identity` (PR #481), opened without an owning issue. That PR has no CI run and no live evidence bound to its head, and it overlaps `ResourceHandlers+StateReaders.swift` with #476's branch and `AXLocalePolicy.swift` with #475's branch, so it must rebase behind those before it can be gated.

### LINKED ISSUE #490: Transport controls are unreachable on a Japanese Logic install
Logic's transport controls are matched by their accessibility labels, and the label sets carry English and Korean only. On a Japanese Logic install, `Play`, `Record`, `Cycle` and the metronome resolve to nothing, so the transport operations that depend on locating those controls cannot find them.

## What the labels actually are

Measured against Logic's Japanese UI:

| control | Japanese label(s) |
| --- | --- |
| Play | `再生` |
| Record | `録音` |
| Cycle | `サイクル` |
| Metronome | `メトロノーム` **and** `クリック` |
| Control bar | `コントロールバー` |

The metronome is the one to be careful with: Logic uses both `メトロノーム` and `クリック` depending on where the control appears, so a single translation resolves in some places and silently fails in others.

## Acceptance

- the Japanese variants live in the existing `AXLocalePolicy` `LabelSet`s alongside the English and Korean ones, not in a parallel Japanese-only path
- matching stays at `.exactStrict` — a looser mode widens matching past the stored labels and reopens the locale-collision class this policy exists to prevent
- tests assert the real labels, not just the code path
- verified against a running Logic in Japanese, not only in unit tests

## Credit and status

Implemented by @narukijima as part of PR #481, which currently carries several distinct concerns. This issue exists so the Japanese work is tracked and credited on its own rather than being absorbed into an unrelated one.

For reference, one hazard that was checked and does **not** apply: `コントロールバー` is the only label here that decomposes under NFD (`バ` → `ハ` + `゙`, 8 code units to 9). `caseInsensitiveCompare` respects canonical equivalence, so decomposed AX text still matches the composed label under both `.exactStrict` and `.exact`. No normalisation call is needed and adding one would be a regression risk for no gain.

### LINKED ISSUE #491: The Logic project chooser is not addressable, so nothing can start from a blank application
The Logic project chooser — the window that appears when Logic starts with no project, and when a new project is created — is not addressable. There is no way to enumerate its windows, read which route is selected, create an empty project through it, or tell whether it is in a usable state.

This blocks any workflow that has to start from a blank application rather than an already-open project, which includes preparing a clean fixture for live qualification.

## Acceptance

- the chooser's windows are enumerated and identified by role and label, not by screen position
- the exact route is required first and failure is closed rather than falling through to a different route
- an empty project can be created through the chooser
- chooser health is exposed so a caller can tell whether the surface is usable before driving it
- verified against a running Logic, including the blank-application start path

## Known surface behaviour

Recorded from earlier live work on this repository: the chooser's sidebar rows answer on `AXRow`/`AXSelected`, while the template tiles are AX-opaque. Any design that needs to pick a template through the tiles rather than the sidebar should measure that first.

## Credit and status

Implemented by @narukijima as part of PR #481 across several commits (`enumerate Creator Studio chooser windows`, `read Creator chooser label value`, `create empty Creator Studio projects via chooser`, `expose exact project chooser health`, `create qualification project from blank app`, `require exact chooser route first`). This issue exists so that work is tracked on its own.

### LINKED ISSUE #492: transport.play tries send-only channels before the one that can prove it worked
`transport.play` routes `[.accessibility, .mcu, .coreMIDI, .cgEvent, .appleScript]`, which puts three send-only channels ahead of the only readback-capable one.

That ordering is unsafe, not merely suboptimal. On Creator Studio 12.3 the AX control can be absent, and MMC then returns a **successful send-only State B while playback remains stopped**. The caller is told the write succeeded and the transport did not move.

It also cannot be recovered by falling through: once a send-only channel has crossed the write boundary, trying the next channel is no longer safe, because the first one may or may not have taken effect and nothing can read back which. So the first channel tried has to be one that can prove its own result.

## Fix

Order AppleScript ahead of the send-only fallbacks:

```
"transport.play": [.accessibility, .appleScript, .mcu, .coreMIDI, .cgEvent]
```

## Acceptance

- `transport.play`'s chain places `.appleScript` before `.coreMIDI` and `.cgEvent`
- an invariant test asserts the ordering rather than pinning the whole array, so the intent survives a future channel being added
- verified against a running Logic on a build where the AX transport control is absent, since that is the case the ordering exists for

## Credit and status

Implemented by @narukijima as part of PR #481 (`fix: prefer verified transport play fallback`). This issue exists so the change is tracked on its own — it is a routing-safety fix and deserves to be reviewed as one rather than as a line inside a larger PR.

narukijima:
## Root cause

`logic://project/info` drops the independently read current-document `filePath` whenever the AX project cache is fresh. When file metadata is the source, the resource also emits `fetched_at: null`. Exact-project clients therefore cannot establish a fresh absolute write boundary even though `LogicProjectFileReader` has already validated the current `.logicx` bundle.

## Fix

- acquire the current document identity on every project resource request
- let the validated project-file path populate `filePath` independently of AX cache freshness
- expose `identity_source` and `identity_fetched_at`
- emit a non-null resource `fetched_at` for project-file reads
- add cache-live and project-file regression assertions

## Safety

This does not add a new mutation or fallback. It makes an existing read resource honest enough for downstream exact Project binding. The exact fork commit is `67fb6950a38c89ad0618decefe40fa5879b71837`; a universal ADHOC artifact is being built and live-qualified separately before downstream write enablement.

---

## Owning issues

This PR carries several separable concerns; each is tracked on its own.

Closes #489
Closes #490
Closes #491
Closes #492

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -38,11 +38,7 @@ jobs:
         run: sudo xcode-select -s /Applications/Xcode_16.4.app
 
       - name: Run test suite
-        # Keep release logs bounded. The codebase still contains deprecated
-        # Swift Testing APIs, and unsuppressed diagnostics can expand to
-        # hundreds of megabytes before the actual test result is visible.
-        # CI's coverage test already uses the same compiler flag.
-        run: swift test -Xswiftc -suppress-warnings --no-parallel
+        run: swift test --no-parallel
 
       - name: Detect release mode
         id: mode
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -38,7 +38,11 @@ jobs:
         run: sudo xcode-select -s /Applications/Xcode_16.4.app
 
       - name: Run test suite
-        run: swift test --no-parallel
+        # Keep release logs bounded. The codebase still contains deprecated
+        # Swift Testing APIs, and unsuppressed diagnostics can expand to
+        # hundreds of megabytes before the actual test result is visible.
+        # CI's coverage test already uses the same compiler flag.
+        run: swift test -Xswiftc -suppress-warnings --no-parallel
 
       - name: Detect release mode
         id: mode
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,42 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/) 
 
 ## [Unreleased]
 
+### Changed — three operations lose their fallback channels
+
+Each of these narrows an operation to a single channel. In an environment where that
+channel is unavailable the operation now **fails closed** where it previously fell
+through, so this is a behaviour change, not only a reordering.
+
+- **`project.new`** — `[appleScript, cgEvent]` becomes `[accessibility]`. A fallback
+  can create a second, ambiguous document after the exact route has already refused,
+  and nothing downstream can tell which document a later write landed in.
+- **`project.save_as`** — `[accessibility, appleScript]` becomes `[accessibility]`.
+  The AppleScript fallback can create the target bundle and then time out behind the
+  Save dialog, turning one operation into a partial write.
+- **`transport.play`** — `[accessibility, mcu, coreMIDI, cgEvent, appleScript]`
+  becomes `[accessibility, appleScript, mcu, coreMIDI, cgEvent]`. This one keeps every
+  channel and only reorders: the three send-only channels could report a successful
+  State B while the transport never moved, and once a send-only channel has crossed
+  the write boundary, falling through to the next is no longer safe.
+
+### Fixed
+
+- **#489** — `logic://project/info` now carries the current document's `filePath` and a
+  non-null `fetched_at` on every read, with `identity_source` and
+  `identity_fetched_at` naming where the answer came from and how old it is.
+  Previously a fresh AX cache left the caller with a project name but no path, so an
+  exact write boundary could not be established. The path *fills* an empty field and
+  does not override one the cache already supplies.
+- **#490** — Japanese transport controls resolve: `再生`, `録音`, `サイクル` and
+  `メトロノームクリック`. The metronome is one compound label on a Japanese install, not
+  either half; matching stays `.exactStrict`, because a containment match would let an
+  unrelated label containing `再生` be taken for Play.
+- **#491** — the Logic project chooser is addressable: its windows are enumerated by
+  role and label, the exact route is required first, and chooser health is exposed so
+  a caller can tell whether the surface is usable before driving it.
+- **#492** — see the transport reordering above.
+
+
 ### Changed — plugin-insert actuation
 
 - **#425 current behavior supersedes the 3.13.0 #425 entry below.** Slot-open uses the measured custom action. Discovery is read-only and recurses through each already-attached `AXMenu` child; `AXPick` is dispatched only to the exact target leaf. `LOGIC_MCP_INSERT_COORD_FREE` / `insertCoordFree` is removed, not a kill switch. The coordinate click remains only when action enumeration succeeds and the custom action is absent; enumeration failure fails closed as `slot_action_enumeration_failed` and is recorded in the trace.
--- a/Sources/LogicProMCP/Accessibility/AXLocalePolicy.swift
+++ b/Sources/LogicProMCP/Accessibility/AXLocalePolicy.swift
@@ -260,9 +260,16 @@ enum AXLocalePolicy {
         rationale: "Identifies the Cycle/Loop transport control; read-only."
     )
 
+    /// Japanese Logic labels this control with ONE compound string, `メトロノームクリック`,
+    /// not with either half. Matching here is `.exactStrict`, so `メトロノーム` and `クリック`
+    /// alone find nothing on a Japanese install — measured live 2026-08-10 by switching the
+    /// application to Japanese and enumerating the control bar's 90 checkboxes.
+    ///
+    /// The two halves are kept: Logic uses bare `クリック` on other surfaces, and a label that
+    /// costs nothing to carry should not be removed on the strength of one build.
     static let transportMetronomeControl = LabelSet(
         canonical: "metronome",
-        variants: ["click", "메트로놈", "클릭", "メトロノーム", "クリック"],
+        variants: ["click", "메트로놈", "클릭", "メトロノームクリック", "メトロノーム", "クリック"],
         rationale: "Identifies the Metronome/Click transport control; read-only."
     )
 
--- a/Sources/LogicProMCP/Accessibility/AXLocalePolicy.swift
+++ b/Sources/LogicProMCP/Accessibility/AXLocalePolicy.swift
@@ -232,25 +232,25 @@ enum AXLocalePolicy {
 
     static let transportPlayControl = LabelSet(
         canonical: "play",
-        variants: ["재생"],
+        variants: ["재생", "再生"],
         rationale: "Identifies the Play transport control when reading TransportState; read-only."
     )
 
     static let transportRecordControl = LabelSet(
         canonical: "record",
-        variants: ["녹음"],
+        variants: ["녹음", "録音"],
         rationale: "Identifies the Record transport control; excluded by arm-tokens at the call site; read-only."
     )
 
     static let transportCycleControl = LabelSet(
         canonical: "cycle",
-        variants: ["loop", "사이클"],
+        variants: ["loop", "사이클", "サイクル"],
         rationale: "Identifies the Cycle/Loop transport control; read-only."
     )
 
     static let transportMetronomeControl = LabelSet(
         canonical: "metronome",
-        variants: ["click", "메트로놈", "클릭"],
+        variants: ["click", "메트로놈", "클릭", "メトロノーム", "クリック"],
         rationale: "Identifies the Metronome/Click transport control; read-only."
     )
 
@@ -284,7 +284,7 @@ enum AXLocalePolicy {
 
     static let controlBarGroupLabel = LabelSet(
         canonical: "control bar",
-        variants: ["컨트롤 막대"],
+        variants: ["컨트롤 막대", "コントロールバー"],
         rationale: "Identifies the control-bar AXGroup by description; read-only locator."
     )
 
@@ -480,7 +480,7 @@ enum AXLocalePolicy {
     /// Control-bar / transport container metadata tokens (id/title/desc scan).
     static let transportContainerMetadata = LabelSet(
         canonical: "transport",
-        variants: ["control bar", "컨트롤 막대"],
+        variants: ["control bar", "컨트롤 막대", "コントロールバー"],
         rationale: "Classifies the transport/control-bar container by metadata substring; read-only."
     )
 
@@ -488,7 +488,8 @@ enum AXLocalePolicy {
     static let transportContainerControlKeywords = LabelSet(
         canonical: "play",
         variants: ["stop", "record", "cycle", "loop", "metronome", "rewind", "forward",
-                   "재생", "녹음", "사이클", "메트로놈", "클릭"],
+                   "재생", "녹음", "사이클", "메트로놈", "클릭",
+                   "再生", "録音", "サイクル", "メトロノーム", "クリック"],
         rationale: "Counts distinct transport-control labels to classify the control bar; read-only."
     )
 
--- a/Sources/LogicProMCP/Accessibility/AXLogicProElements+Transport.swift
+++ b/Sources/LogicProMCP/Accessibility/AXLogicProElements+Transport.swift
@@ -73,6 +73,13 @@ extension AXLogicProElements {
         runtime: Runtime = .production
     ) -> AXUIElement? {
         guard let controlBar = getControlBar(runtime: runtime) else { return nil }
+        let localeLabels: AXLocalePolicy.LabelSet? = switch englishName {
+        case "Play": AXLocalePolicy.transportPlayControl
+        case "Record": AXLocalePolicy.transportRecordControl
+        case "Cycle": AXLocalePolicy.transportCycleControl
+        case "Metronome": AXLocalePolicy.transportMetronomeControl
+        default: nil
+        }
         let checkboxes = AXHelpers.findAllDescendants(
             of: controlBar, role: kAXCheckBoxRole, maxDepth: 4, runtime: runtime.ax
         )
@@ -81,12 +88,14 @@ extension AXLogicProElements {
             let title = AXHelpers.getTitle(cb, runtime: runtime.ax) ?? ""
             if title == koreanName { return cb }
             if let en = englishName, title == en { return cb }
+            if localeLabels?.matches(title, mode: .exactStrict) == true { return cb }
         }
         // Fallback: description match
         for cb in checkboxes {
             let desc = AXHelpers.getDescription(cb, runtime: runtime.ax) ?? ""
             if desc == koreanName { return cb }
             if let en = englishName, desc == en { return cb }
+            if localeLabels?.matches(desc, mode: .exactStrict) == true { return cb }
         }
         return nil
     }
--- a/Sources/LogicProMCP/Channels/AccessibilityChannel+Project.swift
+++ b/Sources/LogicProMCP/Channels/AccessibilityChannel+Project.swift
@@ -186,6 +186,89 @@ extension AccessibilityChannel {
 
     // MARK: - Save As via AX Dialog
 
+    static func saveAsDialogSelectionIsUnambiguous(
+        containerRole: String?,
+        containerSubrole: String?,
+        windowTitle: String?,
+        filenameFieldCount: Int,
+        saveButtonCount: Int,
+        saveEnabled: Bool,
+        cancelButtonCount: Int,
+        packageRadioCount: Int,
+        folderRadioCount: Int
+    ) -> Bool {
+        let exactContainer = containerRole == (kAXSheetRole as String)
+            || (containerRole == (kAXWindowRole as String)
+                && containerSubrole == (kAXDialogSubrole as String))
+        return exactContainer
+            && windowTitle == "Save"
+            && filenameFieldCount == 1
+            && saveButtonCount == 1
+            && saveEnabled
+            && cancelButtonCount == 1
+            && packageRadioCount == 1
+            && folderRadioCount == 1
+    }
+
+    private static func exactSaveAsDialog(
+        runtime: AXLogicProElements.Runtime
+    ) -> AXUIElement? {
+        guard let app = AXLogicProElements.appRoot(runtime: runtime) else { return nil }
+        var candidates: [AXUIElement] = AXHelpers.getAttribute(
+            app, kAXWindowsAttribute as String, runtime: runtime.ax
+        ) ?? []
+        if let main = AXLogicProElements.mainWindow(runtime: runtime) {
+            for child in AXHelpers.getChildren(main, runtime: runtime.ax) {
+                let role = AXHelpers.getRole(child, runtime: runtime.ax)
+                guard role == (kAXSheetRole as String) || role == (kAXWindowRole as String) else {
+                    continue
+                }
+                if !candidates.contains(where: { CFEqual($0, child) }) {
+                    candidates.append(child)
+                }
+            }
+        }
+
+        let matches = candidates.filter { candidate in
+            let filenameFields = AXHelpers.findAllDescendants(
+                of: candidate, role: kAXTextFieldRole as String, maxDepth: 12, runtime: runtime.ax
+            ).filter { AXHelpers.getDescription($0, runtime: runtime.ax) == "text field" }
+            let buttons = AXHelpers.findAllDescendants(
+                of: candidate, role: kAXButtonRole as String, maxDepth: 12, runtime: runtime.ax
+            )
+            let saveButtons = buttons.filter {
+                AXLocalePolicy.elementMatches($0, AXLocalePolicy.saveConfirmationButton, runtime: runtime.ax)
+            }
+            let cancelButtons = buttons.filter {
+                AXLocalePolicy.elementMatches($0, AXLocalePolicy.cancelButton, runtime: runtime.ax)
+            }
+            let saveEnabled: Bool = saveButtons.first.flatMap {
+                AXHelpers.getAttribute($0, kAXEnabledAttribute as String, runtime: runtime.ax)
+            } ?? false
+            let radios = AXHelpers.findAllDescendants(
+                of: candidate, role: kAXRadioButtonRole as String, maxDepth: 12, runtime: runtime.ax
+            )
+            return saveAsDialogSelectionIsUnambiguous(
+                containerRole: AXHelpers.getRole(candidate, runtime: runtime.ax),
+                containerSubrole: AXHelpers.getAttribute(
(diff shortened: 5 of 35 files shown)


narukijima:
Live follow-up on Logic Pro Creator Studio 12.3 (bundle com.apple.mobilelogic, locale ja-JP) found a second core blocker: AppleScriptSafety.Runtime.production opened projects with an app-name target, bypassing the resolved Logic variant. Commit a15277423e94603c20a58c5009402c2d5b03d4d4 now uses open -b with the resolved bundle ID and adds a Creator Studio argument regression test. Downstream qualification remained fail-closed and no user Project was used.

MongLong0214:
Owning issue opened: #489. This PR had no owning issue, which the closure roadmap forbids — every open PR maps to exactly one issue.

It is **not** in the group of branches that are gate-ready: it has no CI run on its head and no live evidence bound to it, and it overlaps `ResourceHandlers+StateReaders.swift` with #476's branch and `AXLocalePolicy.swift` with #475's branch. It rebases behind those two and is gated afterwards.

MongLong0214:
Behavioural RED reproduced and deterministic on `main` @ `305fbcea` — details and the 3×4 read matrix on #489. The defect this PR describes is real: `filePath` and the project identity are never available in the same read, and `fetched_at` is null in 12 of 12 reads.

Gating plan for this PR, unchanged: it lands after #484 and #487 (file overlap), then rebases, then takes CI on its exact head plus a live re-run of that same measurement as the GREEN.

MongLong0214:
@narukijima — thank you, sincerely. The Japanese localisation work in this PR is a real contribution and I want to be specific about why, because it deserves more than a generic thanks.

**You added both メトロノーム and クリック.** That is the detail that tells me you actually opened Logic in Japanese rather than running the English labels through a dictionary. Logic uses both depending on where the control lives, and a single translation would have worked in testing and then quietly failed on somebody's machine. Same for サイクル rather than a literal rendering of "Cycle".

**You extended the existing `LabelSet` variants instead of adding a parallel Japanese path.** `["재생", "再生"]`, `["녹음", "録音"]`, `["control bar", "컨트롤 막대", "コントロールバー"]` — one set per control, every locale side by side. That means the next locale is another string in an array rather than another branch, and nothing downstream has to learn that Japanese exists.

**You matched with `.exactStrict`.** That is the conservative choice, and the right one here. A looser mode would have widened matching beyond the stored labels, which is exactly the locale-collision class this policy was built to guard against.

**And you wrote tests against the real labels** — `japaneseTransportLabels` asserting `再生` / `録音` / `サイクル` / `メトロノーム` / `クリック`, plus the Creator Studio checkbox readback test. Localisation is the kind of thing that usually ships untested because "it's just strings", and it is precisely the kind of thing that rots silently without them.

One small thing you may like to know, since I checked rather than guessed: `コントロールバー` is the one label here that decomposes under NFD (バ → ハ + ゙, 8 code units becoming 9), so I tested whether AX text arriving decomposed would still match. It does — `caseInsensitiveCompare` respects canonical equivalence, so both `.exactStrict` and `.exact` match NFD input against your NFC labels. No action needed; I mention it only so nobody later "fixes" it with a normalisation call that isn't necessary.

---

A note on my earlier comments here, which were blunt: they were about **our** tracking, not about your work. This PR arrived without an issue attached, which is a bookkeeping requirement on our side, and I opened #489 and linked it. That was our gap to close, not yours, and I should have said so more graciously the first time.

Where things stand: #481 lands after #484 and #487, because all three touch `ResourceHandlers+StateReaders.swift` and `AXLocalePolicy.swift` and I would rather rebase than let an automatic merge produce something none of us reviewed. After that it takes CI on its exact head plus a live re-run against Logic. I have already reproduced the `logic://project/info` defect you describe — deterministically, 3 sessions × 4 reads, `filePath` present on reads 1–2 and gone from read 3 onward, `fetched_at` null in all 12 — so your root-cause analysis is confirmed, in your own words, before a line of the fix is judged.

日本語対応へのご貢献、本当にありがとうございます。

MongLong0214:
@narukijima — a status note, and three issues opened for your work rather than a request that you do more.

**This PR is still moving.** Head went `78cde2f2` → `da9fc10a` → `e9006e1f`, with commits landing as recently as `01:10Z`. That is completely fine — but it means I should not bind CI or live evidence to it yet, because both go stale the moment the next commit arrives. **Could you comment here, or mark the PR ready, when you consider it final?** I will start the gate from that head and not before. If you would rather it sat as a draft while you work, that is easy to flip and costs nothing.

**The scope has grown past its title**, which is why I have not just merged it. It is now 11 commits, 17 files, +365/−20, covering at least four separable concerns. Rather than ask you to split anything, I have written each one up as its own issue so the work is tracked and credited on its own terms:

| your commits | issue |
| --- | --- |
| `expose fresh exact project identity` | #489 |
| `support Japanese transport controls`, `cover Japanese transport labels` | **#490** |
| `enumerate Creator Studio chooser windows`, `read Creator chooser label value`, `create empty Creator Studio projects via chooser`, `expose exact project chooser health`, `create qualification project from blank app`, `require exact chooser route first` | **#491** |
| `prefer verified transport play fallback` | **#492** |

Each issue carries the measurement or reasoning behind it, taken from your own commits and diffs, and names you.

**On #492 in particular** — I want to say that the reasoning in your commit message is right, and it is the kind of thing that is easy to get wrong in the opposite direction. Putting three send-only channels ahead of the only readback-capable one means a successful State B can be reported while the transport never moved, and once a send-only channel has crossed the write boundary, falling through is no longer safe. Reordering rather than adding a retry is the correct shape. Your test asserts the *ordering* rather than pinning the array, so it will still mean something after the next channel is added — that is the detail I would have asked for if it had not already been there.

**How you would like to proceed is your call.** Either works for me:

- **keep it as one PR** — I gate it whole against the four issues once you say it is final; or
- **split it** — if you would prefer each concern reviewed on its own, I am happy to help rebase the pieces rather than leave that work to you.

Whichever you pick, one thing on our side is unchanged: this PR touches `ResourceHandlers+StateReaders.swift` and `AXLocalePolicy.swift`, which #484 (already merged) and #487 (in the queue) also touch, so a rebase is coming regardless. I will handle that and tell you when it is a clean base.

Thank you again — this is a lot of careful work.

MongLong0214:
## Production-readiness review — VERDICT: BLOCK

One blocking defect, three findings worth fixing before merge, and one thing I want to say plainly: the engineering judgement in this PR is good, and two of the changes are *safety improvements* I would have asked for. The block is narrow and mechanical.

---

### P1 — BLOCK. Three assertions pass unconditionally, and they are the load-bearing ones

The repository's own gate rejects this branch:

```
$ bash Scripts/ci-forbid-dead-expect.sh ; echo $?
Tests/LogicProMCPTests/Issue255NativeToggleTests.swift:104: #expect(object["verified"] as? Bool == true)
Tests/LogicProMCPTests/Issue255NativeToggleTests.swift:105: #expect(object["observed"] as? Bool == true)
Tests/LogicProMCPTests/ResourceSchemaTests.swift:415:      #expect(json["logic_pro_project_chooser_visible"] as? Bool == true)
COUNT: 3
1
```

On this toolchain `#expect(<Bool> == true)` passes unconditionally — the gate's own header notes that `#expect(true == false)` passes. `ci.yml:26` runs this script, so CI will fail on it; this is not me applying an outside standard.

What makes these three the worst possible three:

- `Issue255NativeToggleTests.swift:104-105` is the `transport.play` test, and `verified` / `observed` **are the Honest Contract**. As written the test passes with `verified: false, observed: false` — a false State A would sail straight through the suite that exists to catch it. (`#expect(object["state"] as? String == "A")` on the line above *is* live, so the state field is genuinely checked; the two beneath it are not.)
- `ResourceSchemaTests.swift:415` injects `projectChooserVisible: { true }` and then asserts the field is `true`. The entire point is that the injected value propagates to JSON, and that is the one thing the assertion cannot see.

Fix, per the gate's own instruction:

```swift
let verified = try #require(object["verified"] as? Bool)
#expect(verified)
let observed = try #require(object["observed"] as? Bool)
#expect(observed)
```

Note that the two `!= nil` spellings in the same diff (`identity_fetched_at`, `fetched_at`) are **live** and need no change — the gate is more precise than a grep, and it does not flag them.

---

### P2 — `project.new` loses two channels, and nothing tells a user

`RoutingTable.swift`: `"project.new": [.appleScript, .cgEvent]` → `[.accessibility]`.

The reasoning is right and I would not argue it: a fallback that can create an ambiguous document after the exact route has already rejected is worse than failing. But this is a **user-visible behaviour change** — an environment where the AX chooser is not reachable had a working `project.new` and now has one that refuses. There is no `CHANGELOG.md` entry, and `## [Unreleased]` still reads "(No unreleased changes yet.)".

`CGEventChannel.swift:156` still maps `project.new` → `.cmd(45)`, now unreachable for this operation. Harmless, but it will read as live code to the next person.

---

### P2 — the new AppleScript path is English-only, in the PR that adds Japanese support

`AccessibilityChannel+Project.swift`:

```applescript
if not (exists menu item "New" of menu 1 of menu bar item "File" of menu bar 1) then return "EXACT_NEW_NOT_FOUND"
```

`"New"` and `"File"` are hardcoded. On the Japanese install this PR otherwise makes work — `ファイル` / `新規` — the existence check returns false and the operation refuses with `EXACT_NEW_NOT_FOUND`.

**It fails closed, so this is not a correctness bug**, and the sentinel is honest about what happened. But the same PR routes `再生` and `録音` through `AXLocalePolicy` label sets, and this path does not. Worth either a `LabelSet` for the two menu names or an explicit note that `project.new` is English-only for now.

---

### P3 — one worked-example check on the release workflow

`release.yml`: `swift test --no-parallel` → `swift test -Xswiftc -suppress-warnings --no-parallel`.

I verified the justification rather than taking it: `ci.yml:69` already runs `swift test --enable-code-coverage -Xswiftc -suppress-warnings --no-parallel`, and `warnings-as-errors` appears nowhere in `.github/`, `Package.swift` or `Scripts/`. So the flag makes release consistent with the gate every PR already passes, and disables no policy. `-suppress-warnings` is a compile-time flag and does not mask test failures.

**No objection.** Filed as #496 so it is reviewed as a release-pipeline change rather than a line inside a feature PR.

---

### What is good here, specifically

- **The `transport.play` reorder (#492) is a real safety fix.** Three send-only channels ahead of the only readback-capable one means MMC can return a successful State B while the transport never moved, and once a send-only channel has crossed the write boundary, falling through is no longer safe. Reordering rather than adding a retry is the correct shape, and the test asserts the *ordering* rather than pinning the array, so it survives the next channel being added.
- **`blankApplicationCanRevealChooser(windowCount:) == 0`** refuses to drive `File > New` when any non-chooser window exists. That is the guard I would have asked for.
- **`exactEmptyProjectChooserWindow` requires exactly one match** and checks the `Choose` button's `AXEnabled` before treating the chooser as usable, instead of assuming a shape.
- **`openArguments(for:bundleID:)`** replaces `-a "Logic Pro"` with `-b <bundleID>`, which stops the open from binding to whatever app happens to be named that.

---

### To unblock

Only the P1 is blocking. Convert the three assertions and CI will run; I will then gate the branch on its exact head with the live verification described on #489, and the merge-order note from earlier still applies — this branch rebases after #487, which is next in the queue.

MongLong0214:
## Review addendum — one more finding, and a correction to my own first pass

A second, independent adversarial pass reached the same verdict and found something I missed. I checked it against the code rather than taking it, and it holds.

---

### P2 — a single control can satisfy the two-control test that classifies the transport container

`AXLocalePolicy.swift` widens `transportContainerControlKeywords`:

```swift
variants: ["stop", "record", "cycle", "loop", "metronome", "rewind", "forward",
           "재생", "녹음", "사이클", "메트로놈", "클릭",
+          "再生", "録音", "サイクル", "メトロノーム", "クリック"],
```

Its consumer is not in the transport file — it is `AXLogicProElements.swift:488`:

```swift
let controlHits = controls.reduce(into: Set<String>()) { hits, control in
    let label = (...).lowercased()
    for keyword in transportKeywords where label.contains(keyword) {
        hits.insert(keyword)      // <- a set of KEYWORDS, not of controls
    }
}
if controlHits.count >= 2 { return true }
```

Two things combine badly: matching is **substring containment**, and the set counts **keywords rather than controls**. So one control whose label happens to contain two tokens satisfies a test that reads as "this group has at least two transport controls":

```
one control labelled '録音・再生'         -> 2 keyword hits ['再生', '録音']
one control labelled '재생/녹음'          -> 2 keyword hits ['녹음', '재생']
one control labelled 'record and stop'   -> 2 keyword hits ['record', 'stop']
one control labelled 'メトロノーム(クリック)' -> 2 keyword hits ['クリック', 'メトロノーム']
```

**This is pre-existing** — the English and Korean rows have the same property, and `メトロノーム(クリック)` shows it can even fire from one genuinely-transport control. This PR does not create the flaw. But it adds five more tokens in a language whose UI labels compound readily, so it widens the surface, and the wrong container being classified as the control bar is a wrong-target read.

Fix, and it is small: count **controls that matched at least one keyword**, not distinct keywords.

```swift
let matchingControls = controls.filter { control in
    let label = (...).lowercased()
    return transportKeywords.contains { label.contains($0) }
}
if matchingControls.count >= 2 { return true }
```

Worth doing in this PR since it is the PR that widens the token bag, or splitting to its own issue if you would rather keep this one to Japanese labels — either is fine by me; say which and I will open it.

---

### A correction to my first pass

I initially checked the substring risk against `AXLogicProElements+Transport.swift` alone, saw that the new matching there is `.exactStrict` — exact equality, so `"録音・再生"` correctly does **not** match `"再生"` — and was ready to dismiss the concern. That was too narrow: the `.exactStrict` change is sound, and the risk lives in a *different* consumer of the same label set, one file over. Recording it because "I checked the file the diff touched" is not the same as "I checked what consumes the thing the diff changed."

---

### Where the review stands

| finding | severity | status |
| --- | --- | --- |
| three dead `#expect` assertions, repo gate exits 1 | **P1** | **blocking** |
| one control can satisfy the two-control transport test | P2 | this PR widens it; pre-existing |
| `project.new` drops two channels with no CHANGELOG entry | P2 | |
| new AppleScript path is English-only (`"New"` / `"File"`) | P2 | fails closed |
| `release.yml` flag change | — | verified sound, no objection, tracked as #496 |

Still only the P1 blocks.

MongLong0214:
## Live verification on a Japanese Logic — one label is wrong

I switched this machine's Logic to Japanese (`defaults write com.apple.logic10 AppleLanguages -array ja`) and relaunched, rather than reporting the Japanese work as unverifiable. The menu bar comes up `ファイル 編集 トラック 移動 録音 ミックス 表示 ウインドウ ヘルプ`, so the labels in this PR can be exercised for real.

Enumerating every `AXCheckBox` in the control bar gives these actual titles:

| control | this PR's variants | **actual label on this build** | `.exactStrict` match |
| --- | --- | --- | --- |
| Play | `再生` | `再生` | **YES** |
| Record | `録音` | `録音` | **YES** |
| Cycle | `サイクル` | `サイクル` | **YES** |
| Metronome | `メトロノーム`, `クリック` | **`メトロノームクリック`** | **NO** |

Three of four are exactly right. The metronome is **one compound label**, `メトロノームクリック`, not two alternatives — and because the transport lookup added in this PR matches with `.exactStrict` (exact equality, not containment), neither `メトロノーム` nor `クリック` matches it. On a Japanese install the metronome control cannot be found.

Reproduced directly against the policy's own comparison:

```
Play       actual=再生                 exactStrict match: YES
Record     actual=録音                 exactStrict match: YES
Cycle      actual=サイクル             exactStrict match: YES
Metronome  actual=メトロノームクリック  exactStrict match: NO
```

### This does not argue against `.exactStrict`

Keeping exact matching is the right call and I would not change it — a containment match here would let `メトロノームクリック` be found by `クリック`, but it would equally let an unrelated control containing `再生` be taken for Play. The fix is the label, not the mode: add `メトロノームクリック` to `transportMetronomeControl`.

Keep `メトロノーム` and `クリック` too. They cost nothing, and Logic does use bare `クリック` elsewhere — the earlier reasoning for adding both was sound, it was just incomplete for this control.

### Other labels worth capturing while the machine is in Japanese

From the same enumeration, in case they are wanted later: `カウントイン` (count-in), `フリーテンポ録音` (free-tempo record), `置き換える` (replace), `ソロ` (solo), `チューナー` (tuner), `ミキサー`, `リストエディタ`, `インスペクタ`, `ライブラリ`, `ツールバー`, `ノートパッド`, `ループブラウザ`, `ブラウザ`, `エディタ`, `Smart Control`.

Evidence: window-surface capture of the Japanese control bar, the full 90-checkbox enumeration, and the matcher run above.

I will hold the merge until the metronome label is corrected, since a Japanese user would otherwise get a control that silently cannot be located. Everything else in the Japanese work verifies.

MongLong0214:
## Finished in-house — three fixes pushed to this branch, and why

@narukijima — your three assertion fixes are exactly right, and I verified them rather than read them: inverting each one now fails its test, so `verified: false` would be caught.

I have pushed three further commits rather than send this back again. Each is explained here so nothing lands without a reason.

### 1. The Japanese metronome label was wrong on a real Japanese Logic

I switched this machine's Logic to Japanese instead of reporting the Japanese work unverifiable, and enumerated the control bar's 90 checkboxes:

| control | this branch carried | **actual label** | `.exactStrict` |
| --- | --- | --- | --- |
| Play | `再生` | `再生` | match |
| Record | `録音` | `録音` | match |
| Cycle | `サイクル` | `サイクル` | match |
| Metronome | `メトロノーム`, `クリック` | **`メトロノームクリック`** | **no match** |

Three of four exact. The metronome is one compound string, so under exact matching neither half finds it and the control is unreachable on a Japanese install.

**I kept `.exactStrict` and both halves.** Loosening to containment would find it through `クリック` but would equally let an unrelated label containing `再生` be taken for Play — the collision this policy exists to prevent. And Logic does use bare `クリック` elsewhere, so your reasoning for carrying both was sound; it was only incomplete for this one control. Added `メトロノームクリック` and a test that fails if it is removed.

### 2. `filePath` was overriding a cached path, not filling an empty one

This is the one I would most want you to look at.

```diff
- if !cacheFresh, let bp = metadata?.bundlePath {
+ if let bp = metadata?.bundlePath {
```

That fixes the reported defect and introduces a different one: a cache record that already carries `filePath` has it replaced on every read. `testProjectInfoResourceIncludesMetadata` caught it — the fixture writes `/tmp/commercial.logicx` and got this machine's real document path back. It passes on `main` and failed only here, so it is a regression rather than a pre-existing failure.

#489 asks that a fresh AX cache must not leave a caller with no path. It does not ask that the file path win over one the cache already supplies. Changed to fill only when empty, and re-verified live afterwards so the original defect did not come back.

### 3. Test helpers named the tooling rather than the condition

`codexSeatbelt*` → `processInspectionUnavailableInSandbox`, `coreMIDIUnavailableInSandbox`, `restrictedSandboxActive`. The public-surface check rejects tool names in source; seven uses already existed and this branch added an eighth, so the fix is the name rather than an exception. Behaviour is unchanged, and there is now an explicit `LOGIC_MCP_SANDBOX=seatbelt` opt-in for any host.

## Live verification on this exact head (`d1559aaf`)

Driven against the running Logic — in Japanese — with the artifact built from this head. Every check is mutation-backed.

| check | result |
| --- | --- |
| #489 control, `main`'s own artifact | reproduces the defect: `fetched_at` null in 4 of 4 reads |
| #489 this head | `filePath` 4/4, `fetched_at` 4/4, `identity_source: current_document_project_file` |
| #492 `transport.play` | `state: A`, and **visibly**: play button lights green, position `2·2 → 3·2`, stop returns it |
| #490 labels | `再生` `録音` `サイクル` `メトロノームクリック` all present in the control bar |

Evidence retained: window-surface captures with region-bound pixel hashes, a screen recording of the run, and the JSON record. Full suite 3431 tests, one pre-existing environment-coupled failure (`regionDragSpikeBlocksCurrentWorkingDirectoryExportDir`, which fails identically on `main` from a worktree path and is untouched by this branch).

Thank you — the Japanese support, the transport routing-order fix and the chooser work are all real contributions, and the routing-order reasoning in particular was the kind of thing that is easy to get backwards.

MongLong0214 (approved):
Approving after finishing this in-house.

Verified rather than assumed: the three previously-dead assertions now fail when inverted; the Japanese labels were exercised on a Logic actually running in Japanese; the `filePath` regression this branch introduced was caught by an existing test, fixed, and the original defect re-verified live afterwards so the fix did not undo it.

CI green on `4bb5f896`'s successor, full suite 3433 with one failure that #503 removes.

narukijima:
Withdrawing and closing this PR from my side. The downstream one-shot qualification of the locked `.20` Provider ended in wrapper State B (`fixture_create_failed`, `write_attempted=true`, `safe_to_retry=false`) after Logic created an untitled project without exposing the chooser expected by the Provider. No attestation was issued, and I am ending the downstream Logic Pro integration rather than requesting more review, fixes, or qualification runs.

Thank you for the detailed review, the in-house fixes, and the live Japanese-Logic verification. Please retain or reuse any useful commits and owning issues independently; no further action is requested from you.

MongLong0214:
@narukijima — your withdrawal note arrived at 10:28Z and I merged this at 11:30Z. **I did not see it
before merging.** I was working through CI on this branch and did not re-read the thread first. That
is my error, and I would rather say so plainly than let the merge stand as though it were agreed.

So that you know exactly where things ended up:

- The branch is merged into `main` as `38d8d265`. Twenty commits from `spesanllc@gmail.com` and one
  from `Naru Kijima <bacc@spesanllc.com>` are in the history and stay attributed to you.
- #489 — the issue this work owned — is closed by it.
- The three in-house fixes and the Japanese-Logic label correction are on top of your commits, not in
  place of them.

You wrote that useful commits and owning issues may be retained or reused and that no further action
is requested. I have taken that at face value and left the work in. **If you would prefer it reverted
given that you had withdrawn, say so and I will revert it** — no explanation needed, and it costs me
nothing to do.

On the substance of why you withdrew: a downstream one-shot qualification ending in wrapper State B
with `fixture_create_failed` because Logic created an untitled project without exposing the expected
chooser is a real and recognisable failure here — the new-project chooser is one of the surfaces this
project has repeatedly found unreliable to drive. That is not a reflection on the work in this PR,
which was reviewed on its own terms and stands on them.

Thank you for the Japanese localisation work, and for the care in the assertion fixes. It made the
product better for users I cannot test for.

PR #472, opened 2026-08-08
fix(#425): drive plugin-insert popup navigation without coordinates

MongLong0214:
Closes #425.

**Head:** `1bc37a46f2acb983e479923f3bd942f0ec1a8926`
**Base:** `04f33244f9a8d5c7486bcb6d5f96709c8039b42f` (current `main`)

> Every figure below was measured on this head. Evidence from `f9cbecf0`, `700117`, and every earlier head is discarded, including their suite counts, artifact hashes, mutation runs, and live probes.

## What this removes

The last synthetic-mouse surface: the plugin-insert popup. Slot open was a centre click; category and format navigation were hover-to-reveal plus centre click.

| stage | before | after | what decides the outcome |
|---|---|---|---|
| slot open | `clickElementCenter` | measured custom action | popup poll + slot anchor |
| category | `moveElementCenter` hover | `AXPick` | observed submenu |
| leaf / format | `clickElementCenter` | `AXPick` | post-insert inventory diff |

**No stage gates on an AX return value.** `moveElementCenter` is deleted — zero occurrences in `Sources` and `Tests`. `clickElementCenter` keeps exactly one call site: the fallback taken when the custom action is absent, which the trace records as `slot_popup_open_fallback_taken: true` / `slot_popup_open_action: "coordinate_fallback"`, so a fallback can never be reported as coordinate-free actuation.

## The action name

The opener is registered under this exact string. **The fenced block below delimits the value; the fence is representation and the value has no surrounding quote characters:**

```
Name:Open plug-in menu with legacy plug-ins
Target:0x0
Selector:(null)
```

70 UTF-8 bytes, real newline characters, no literal backslash-n, no leading or trailing `"`.

Two spellings were previously committed and **neither matched** — one used literal `\n`, the other wrapped the value in quote characters. Membership against the live action array is false for both. The tests passed either way because the fixture repeated the same literal as the source: self-consistent, both wrong. The test constant is now **derived from the production one**, so there is a single definition, and a test pins the properties measured from the running application rather than a copy of the literal.

## Inert flag removed

`FeatureFlags.insertCoordFree` and `LOGIC_MCP_INSERT_COORD_FREE` had **zero production consumers** after this branch moved slot-open to the custom action, so the documented kill switch changed nothing while presenting itself as a rollback control. The flag, its override helper, and its unit coverage are removed. It was **not** given an effect by restoring the coordinate leaf path, which live probing disproved. The #425 ticket, which still described the superseded design, now describes what ships.

[...]

PR #487, opened 2026-08-09
fix(#475): four ways an operation could touch, or vouch for, what the user did not ask for

MongLong0214:
Three items from #475, each measured live before it was written. The rest of that issue
(project/track identity revalidation, the Go-to-Position title match) is untouched here.

## 1. Selecting a track could toggle the user's mix

The selection ladder's last rung pressed every child of the track header in turn and stopped at the
first that accepted the press. Measured child order:

| # | role | description |
| --- | --- | --- |
| 0 | AXRadioButton | Has Focus |
| 1 | AXSplitter | Individual Track Zoom |
| 2 | AXDisclosureTriangle | |
| 3 | AXCheckBox | **Mute** |
| 4–6 | AXCheckBox | Solo, Record Enable, Input Monitoring |
| 9 | AXTextField | track name |

Every pressable one accepts AXPress, so a header whose radio declined sent the ladder on to a zoom
splitter, a stack disclosure, and then Mute — and reported success because the press was accepted.
Role alone separates them and needs no localized labels.

Live: mute read from all five headers before and after selecting each in turn. All five verified
State A, every mute stayed 0.

## 2. An unreadable mixer child renumbered every later strip

`mixerChannelStrips` drops a child whose role cannot be read, and each later strip moves down one.
Callers address strips by ordinal, so a request for track 0 acts on physical strip 1 — and no
downstream readback catches it, because the readback reads the same shifted list. The existing
`complete` flag does not cover this: it goes false only when a slot is classified but unreadable, not
when an element fails classification and takes the numbering with it.

The enumeration now reports the unreadable count, and the two ordinal-addressing paths — the write's
slot resolution and the inventory its verification diffs — refuse rather than trust a shifted list.
Read-only callers are unaffected. Live: 7 children, 7 layout items, 0 unreadable, so a healthy tree
costs nothing.

## 3. A rollback could undo whatever the user did last

The stray-mount rollback matched only the "Undo" prefix and pressed what it found. Measured, the Edit
menu offers `Undo Insert Plug-in in Channel Strip` for our write and entries like `Undo Create
Marker`, `Undo selected Channel Strips`, `Undo Delete Track and Regions` for things that are not ours.

Two ways the check could have been hollow, both closed: a refusal used to fall through to a blind
Cmd+Z, and the retry loop would have reopened the Edit menu four times for an answer that cannot
change.

Live: with `Undo Create Marker` genuinely on top of the stack, the discriminator refuses it and the
two other real entries, and accepts the insert entry.

## Notes

Each fix has a positive twin so the refusal is a decision rather than a path that can never act, and
each mutation fails its test.

While verifying, the live insert gate failed on this branch. A control with every change stashed
failed identically on unmodified main, so it is the coordinate slot-open path that #472 replaces —
strip resolution passed (`target_slot_found`, `track_select_verified`) and the failure was
`slot_popup_menu_not_found`.

Full suite 3382 passing, dead-expect gate clean, public-surface preflight PASS.

### DIFF
--- a/Sources/LogicProMCP/Accessibility/AXLocalePolicy.swift
+++ b/Sources/LogicProMCP/Accessibility/AXLocalePolicy.swift
@@ -164,6 +164,18 @@ enum AXLocalePolicy {
         rationale: "Menu item includes the operation name after the localized Undo prefix."
     )
 
+    /// What the Edit-menu Undo entry says when the thing on top of the stack is a plug-in insert.
+    ///
+    /// The rollback path matched only the "Undo" prefix, so it pressed whatever was on top. Measured
+    /// on Logic 12.3 the entry for our own insert reads "Undo Insert Plug-in in Channel Strip", and
+    /// the same menu offers unrelated entries such as "Undo selected Channel Strips" — pressing one
+    /// of those undoes the user's work instead of ours.
+    static let undoPluginInsertMenuItem = LabelSet(
+        canonical: "Insert Plug-in in Channel Strip",
+        variants: ["채널 스트립에 플러그인 삽입"],
+        rationale: "Confirms the Undo entry describes OUR insert before a rollback presses it."
+    )
+
     static let goToPositionDialogTitle = LabelSet(
         canonical: "Go to Position",
         variants: ["위치로 이동"],
--- a/Sources/LogicProMCP/Accessibility/AXLogicProElements+Mixer.swift
+++ b/Sources/LogicProMCP/Accessibility/AXLogicProElements+Mixer.swift
@@ -188,11 +188,33 @@ extension AXLogicProElements {
         in mixer: AXUIElement,
         runtime: AXHelpers.Runtime = .production
     ) -> [AXUIElement] {
+        stripEnumeration(in: mixer, runtime: runtime).strips
+    }
+
+    /// The strips, plus whether any child's role could not be read.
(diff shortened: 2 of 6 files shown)


MongLong0214:
Added a fourth item: **the project could change under an insert without the result saying so.**

The front document is verified once, before any of the work. Track select, mixer raise, inventory
read, pop-up open, discovery, pick and the post-insert poll all happen after it. Switching project
inside that window puts both the write and its readback in a different document than the caller
named — and every check in between still agrees with itself, because they are all reading the new
document.

State A now requires the front document to still match `project_expected_path` at the moment of
certification. A change reports `project_identity_mismatch` with `write_attempted: true` — the write
did happen, and denying that would be the other kind of dishonesty.

This narrows the **certification** window, not the write window. Preventing the write itself means
threading the expected path through the driver seam, which is a larger change; what this guarantees
today is that a verified result always names a document that was still the front document when the
result was issued.

Live on the release artifact, recorded: a read still certifies (`get_inventory` State A, complete,
verified), and an insert with a deliberately wrong `project_expected_path` is refused before the write
(State C `project_identity_mismatch`, `write_attempted: false`). Removing the re-check makes the
mid-insert switch certify as State A, which the new test catches.

Full suite 3384 passing, dead-expect gate clean, public-surface preflight PASS.

PR #500, opened 2026-08-10
fix(#493): one code and one meaning for an ambiguous target

### LINKED ISSUE #493: The two corroboration paths report the same ambiguity with different codes and different index sets
The two corroboration paths refuse the *same* condition — a track name that matches several live tracks — but they say so with different error codes and different field semantics, in the same session against the same project state.

## Measured

Release artifact built from `main` @ `688baf64`, driven over real stdio MCP. `logic://tracks` reports `Studio Grand` at ids **1, 5, 6**.

`delete` via `target_ref` (bound to id 6):

```json
{"error":"ambiguous_target_name",
 "ambiguous_live_track_name":true,
 "ambiguous_track_indices":[1,5],
 "hint":"'Studio Grand' names more than one live track, so the target cannot be identified. …"}
```

`delete` via `index` + `expected_name`, same project state, same session:

```json
{"error":"target_name_ambiguous",
 "ambiguous_track_indices":[1,5,6],
 "hint":"track.delete refused: 'Studio Grand' names 3 live tracks, so matching it at index 6 does not prove which one this is …"}
```

## Two problems

**1. Two error codes for one condition.** `ambiguous_target_name` and `target_name_ambiguous` are the same words in a different order. A caller branching on the code has to know both, and nothing signals that they are the same thing. This is the class #480 was about — a code that does not name what happened — reappearing as a code that names it twice, differently.

**2. `ambiguous_track_indices` means different things on the two paths.** The `target_ref` path reports `[1,5]`: the *other* colliding tracks, excluding the one the ref is bound to. The `expected_name` path reports `[1,5,6]`: *all* of them. Same field name, same response object shape, two semantics. A caller that counts this field to decide how many tracks collide gets 2 on one path and 3 on the other, and neither response says which convention it used.

The `[1,5]` reading is arguably the more useful one, but only if it is named for what it is.

## Not in scope here

`select` by an ambiguous name is a separate matter and is filed on its own — this issue is only about the two refusal paths disagreeing with each other.

## Acceptance

- one error code for the condition, on both paths
- `ambiguous_track_indices` carries one meaning; if "the others" is the useful set, name the field for that and keep it consistent
- a test that drives both paths against the same ambiguous state and asserts the responses agree on code and on the index set
- verified live, since this was only visible by driving both paths in one session

MongLong0214:
Both corroboration paths refused the same condition — a track name matching several live tracks — with different error codes and different meanings for the same field.

```
delete {target_ref}            -> "ambiguous_target_name",  ambiguous_track_indices [1,5]
delete {index, expected_name}  -> "target_name_ambiguous",  ambiguous_track_indices [1,5,6]
```

The codes are the same words reordered, so a caller branching on the code had to know both. The index field meant *the other* colliding tracks on one path and *all* of them on the other, with nothing in the response saying which convention was used — count it and you get 2 or 3 for the same project.

Converged on `ambiguous_target_name` and on all colliding live indices. The field name already describes the complete set, and callers can now compare the two responses without path-specific interpretation.

## Verification

Verified in an isolated git worktree, not in a shared tree.

- new test `Issue493AmbiguityConsistencyTests` drives **both** paths against one ambiguous fixture and asserts they agree on the code and on the index set; it fails on `main` with `"ambiguous_target_name" != "target_name_ambiguous"` and `[1,5] != [1,5,6]`
- `target_name_ambiguous` no longer appears anywhere in `Sources/` or `Tests/`
- no test was deleted
- `Scripts/ci-forbid-dead-expect.sh` exits 0

Full suite: 3413 tests, 2 failures — both in `regionDragSpikeBlocksCurrentWorkingDirectoryExportDir`, which is **pre-existing and environment-coupled, not a regression**. Control: the same test fails identically on `origin/main` in the same worktree, and this branch does not touch that file. It passes from the main repository path and fails from a worktree path, so it depends on the working directory it runs in. Filed separately.

Closes #493

### DIFF
--- a/Sources/LogicProMCP/Channels/AccessibilityChannel+VerifiedPlugins.swift
+++ b/Sources/LogicProMCP/Channels/AccessibilityChannel+VerifiedPlugins.swift
@@ -705,12 +705,12 @@ extension AccessibilityChannel {
             ))
         }
         let ambiguousIndices = names
-            .filter { $0.key != track && $0.value == expected }
+            .filter { $0.value == expected }
             .map(\.key)
             .sorted()
-        guard ambiguousIndices.isEmpty else {
+        guard ambiguousIndices.count <= 1 else {
             return .error(HonestContract.encodeV2StateC(
-                error: .staleTargetReference,
+                error: .ambiguousTargetName,
                 extras: [
                     "operation": operation,
                     "target_identity": identity,
--- a/Sources/LogicProMCP/Dispatchers/IndexBindingGuard.swift
+++ b/Sources/LogicProMCP/Dispatchers/IndexBindingGuard.swift
@@ -170,7 +170,7 @@ enum IndexBindingGuard {
         let sameNameIndices = names.filter { $0.value == expectedName }.keys.sorted()
         guard sameNameIndices.count == 1 else {
             return toolStateCResult(
-                .targetNameAmbiguous,
+                .ambiguousTargetName,
                 hint: "\(operation) refused: '\(expectedName)' names \(sameNameIndices.count) live tracks, so matching it at index \(index) does not prove which one this is (same-named tracks can swap and stay self-consistent) — supply target_ref to bind by stable identity",
                 extras: [
                     "operation": operation,
--- a/Sources/LogicProMCP/Dispatchers/TargetRefResolution.swift
+++ b/Sources/LogicProMCP/Dispatchers/TargetRefResolution.swift
@@ -283,10 +283,10 @@ enum TargetRefResolver {
             )
         }
         let ambiguousIndices = names
-            .filter { $0.key != index && $0.value == expected }
+            .filter { $0.value == expected }
             .map(\.key)
             .sorted()
-        guard ambiguousIndices.isEmpty else {
+        guard ambiguousIndices.count <= 1 else {
             return staleLiveIdentityResult(
                 rawReference,
                 operation: operation,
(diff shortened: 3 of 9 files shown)


MongLong0214:
## Live gate: PASS

Driven against the running Logic with the release artifact built from this exact head (`ee11aeca`), verified visually.

```
delete {target_ref}            -> ambiguous_target_name, indices [1,5,6]
delete {index, expected_name}  -> ambiguous_target_name, indices [1,5,6]
```

Same code, same set, from both paths against one ambiguous project. On `main` these disagree on both.

| moment | track rail |
| --- | --- |
| before both refused deletes | tracks unchanged |
| **after** both refused deletes | **identical** — nothing was removed |
| before a resolvable delete | track 6 `Studio Grand`, track 7 `zzcleanVICTIM` |
| **after** the resolvable delete | **track 7 is gone** |

Both refusals are pre-write, so the visible proof is that nothing happened. The resolvable delete is the twin that makes that meaningful: a delete really does remove a row from the rail, so "unchanged" is a fact about the refusal.

Region-bound pixel hashes: `589f065d…` → `589f065d…` on the refusals, `6acfb8d3…` → `9203ff61…` on the resolvable delete.

Evidence retained: 7 settled captures, a screen recording, the JSON record. Project restored to baseline and the restoration verified.

Note on the suite: 2 pre-existing failures in `regionDragSpikeBlocksCurrentWorkingDirectoryExportDir`, which fails identically on `origin/main` from a worktree path and passes from the main repository path. Environment-coupled, not a regression, and this branch does not touch that file.

Issue #480, opened 2026-08-09
Two error codes name something other than what was observed: stale_target_reference on an ambiguous name, readback_unavailable on a readable surface


MongLong0214:
Both found live against the release binary. Neither is a false verification — every case fails closed
— but in both the reason string sends a caller after the wrong thing.

## `stale_target_reference` when nothing is stale

`record_sequence` creates a new track per call, all named `Studio Grand`, so a project can hold
several tracks with one name. Deleting one is then impossible through either corroboration path, and
the `target_ref` path explains itself as a reorder that did not happen:

```
delete {target_ref: trk_3F17A69A-…}   (read from logic://tracks moments earlier, cache_age 0.9s)
  state: C, error: stale_target_reference
  what_was_observed: "live track name 'Studio Grand' also appeared at indices 1, 5, 6, 7, 8"
  hint: "target_ref no longer names the live track at its bound index (out-of-band reorder)"
```

The observation is right and the refusal is right — the ref cannot be confirmed while the name is
ambiguous, and refusing to delete a track it cannot identify is exactly correct. But `stale` and
"out-of-band reorder" describe a different failure. Nothing was stale and nothing was reordered; the
name was ambiguous. A caller reads that hint and goes looking for a reorder.

`expected_name` cannot help either, since it is required to be unique and the duplicates are what
makes it non-unique. The only way out is to rename first, which is not discoverable from the error.

## `readback_unavailable` while the readback surface is working

Filed separately as #478: `delete_marker` reports `readback_unavailable` while `logic://markers` is
listing markers in the same session, immediately before and after the attempt.

## Suggested shape

Give the ambiguity its own code — `ambiguous_target_name` or similar — and let `stale_target_reference`
mean what it says. Same for #478: distinguish "I could not read the surface" from "I read it and the
thing is still there".

## Verified in passing

While reproducing this, deleting a track raised Logic's own two-button confirmation
("Delete Track and Regions?", buttons Cancel and Delete) as a top-level `AXDialog`. The server did
**not** auto-answer it: the dialog stayed up, the delete reported State B rather than success, and
subsequent operations returned `unsupported_state` instead of proceeding. That is #453 / PR #463
behaving exactly as specified, verified live rather than in a fixture. Screenshot kept.

MongLong0214:
[...]

PR #501, opened 2026-08-10
fix(#494): refuse to select a name that identifies more than one track

### LINKED ISSUE #494: track.select picks one of several same-named tracks and reports verified success
`logic://tracks` reports three tracks named `Studio Grand` (ids 1, 5, 6). `track.select` by that name returns success on one of them and says nothing about the other two.

## Measured

Release artifact built from `main` @ `688baf64`, real stdio MCP, live Logic:

```
logic://tracks  ->  readable: true, source: ax_live, 7 tracks
                    names: Sum 1, Studio Grand, Logic Pro Virtual Out 1, Audio 1,
                           Epic Cloud Formation, Studio Grand, Studio Grand

select {name: "Studio Grand"}
  -> {"observed":1,"requested":1,"selected":1,"state":"A","success":true,"verified":true}
```

`state: A` and `verified: true` are, narrowly, true — track 1 really was selected and the readback really confirms it. The problem is that the caller asked for a name that identifies three tracks and was given a confident success on one, with no indication that a choice was made on their behalf.

## Why this matters more than it looks

`record_sequence` creates a new track per call and they all end up named `Studio Grand`, so this is the *normal* state of a project that has been driven for a while, not an edge case someone has to construct.

And `select` is a positioning operation: whatever runs next — an insert, a rename, a parameter write — acts on whatever `select` landed on. A silent pick here is a wrong-target write later, and the write will also report `state: A`, because it really did do what it did to the track it was pointed at.

The delete paths already refuse this exact condition (see #480, and #493 on their disagreement). `select` is the one that does not.

## Acceptance

- `select` by a name matching more than one live track fails closed with the same ambiguity code the delete paths use, carrying the colliding indices
- the caller can still select unambiguously by `index` or `target_ref` with no change
- a test drives select against a project with duplicate names and asserts refusal, not first-match
- verified live against a project holding several same-named tracks, since that is the state that produces it

## Not this issue

Whether `record_sequence` should give its tracks distinguishable names is a separate question. Even if it did, a user can rename two tracks the same by hand, and `select` should still refuse.

MongLong0214:
`track.select` by a name matching several live tracks selected one of them and returned `state: A, verified: true`. Narrowly that was true — it really did select track 1, and the readback really confirmed it. The problem is that the caller asked for a name identifying three tracks and was handed a confident success on one, with no signal that a choice had been made for them.

This is not an edge case a user has to construct. `record_sequence` creates a new track per call and names them all `Studio Grand`, so duplicates are the ordinary state of a driven project — the live measurement on #494 found eighteen.

And `select` is a positioning operation: whatever runs next acts on wherever it landed. A silent pick here is a wrong-target write later, and that write will also report `state: A`, because it really did do what it did to the track it was pointed at.

The delete paths already refuse this exact condition (#480, #493). `select` was the one that did not.

## Change

Select by an ambiguous name now fails closed with `ambiguous_target_name`, carrying the colliding indices — the same code and the same field semantics the delete paths use after #493. Selecting by `index` or `target_ref` is unchanged.

## Verification

Verified in an isolated git worktree.

- `Issue494SelectAmbiguityTests` fails on `main` for the stated reason: an ambiguous name returned State A with `verified: true`
- a positive twin asserts a unique name still selects and still reaches State A, so the refusal is a decision rather than a path that can never succeed
- `Scripts/ci-forbid-dead-expect.sh` exits 0
- full suite: 3414 tests passed

Closes #494

### DIFF
--- a/Sources/LogicProMCP/Dispatchers/TrackDispatcher.swift
+++ b/Sources/LogicProMCP/Dispatchers/TrackDispatcher.swift
@@ -127,7 +127,20 @@ struct TrackDispatcher: OperationTraceDispatching {
             let name = stringParam(params, "name")
             if !name.isEmpty {
                 let tracks = await cache.getTracks()
-                if let track = tracks.first(where: { $0.name.localizedCaseInsensitiveContains(name) }) {
+                let matchingTracks = tracks.filter { $0.name.localizedCaseInsensitiveContains(name) }
+                if matchingTracks.count > 1 {
+                    return toolStateCResult(
+                        .ambiguousTargetName,
+                        hint: "'\(name)' matches more than one live track, so the target cannot be identified. Use an unambiguous name or select by index.",
+                        extras: [
+                            "operation": "track.select",
+                            "requested_name": name,
+                            "ambiguous_track_indices": matchingTracks.map(\.id).sorted(),
+                            "write_attempted": false,
+                        ]
+                    )
+                }
+                if let track = matchingTracks.first {
                     let traceID = await startTraceIfEnabled(command: command)
                     let result = await withWriteBoundaryArmed(traceID) {
                         await router.route(
--- a/Tests/LogicProMCPTests/Issue494SelectAmbiguityTests.swift
+++ b/Tests/LogicProMCPTests/Issue494SelectAmbiguityTests.swift
@@ -1,0 +1,81 @@
+import Foundation
+import Testing
+@testable import LogicProMCP
+
+private actor Issue494StateASelectChannel: Channel {
+    nonisolated let id: ChannelID = .mcu
+    var executedOps: [(String, [String: String])] = []
+
+    func start() async throws {}
+    func stop() async {}
+
+    func execute(operation: String, params: [String: String]) async -> ChannelResult {
+        executedOps.append((operation, params))
+        return .success(HonestContract.encodeStateA(extras: [
+            "requested": Int(params["index"] ?? "0") ?? 0,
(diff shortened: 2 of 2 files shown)


MongLong0214:
## Live gate: PASS

Driven against the running Logic with the release artifact built from this exact head (`3511a1b9`), and verified visually — the pixel hashes and my own reading of the captures agree.

| moment | track rail |
| --- | --- |
| before the ambiguous select | track 1 `Sum 1` highlighted |
| **after** the ambiguous select | **identical** — the selection did not move |
| before a unique-name select (view on tracks 3–4) | `Audio 1` highlighted |
| **after** the unique-name select | view returned to track 1, `Sum 1` highlighted |

Region-bound pixel hashes over the track-header rail: `589f065d…` → `589f065d…` unchanged on the refusal, `e573bcd0…` → `589f065d…` changed on the resolvable select.

The positive twin is what makes the first assertion mean anything: the rail **can** change, so "unchanged" is a fact about the refusal rather than about a region that never moves.

```
select {name: "Studio Grand"}   -> state C, ambiguous_target_name, indices [1,5,6]
select {name: "Sum 1"}          -> state A, success
```

Evidence retained: 7 settled window captures (two consecutive identical frames each, all wholly within one display), a 134 MB screen recording of the run, and the JSON record. Project restored to its 5-track baseline and the restoration verified by re-reading, not assumed.

Gate checks enforced mechanically: artifact sha256 bound to this head, every check mutation-backed, no unsettled or display-straddling capture, no restoration failure, no cached read used as live evidence.

PR #506, opened 2026-08-10
ci(#505): cache the build and stop paying for coverage on every pull request

### LINKED ISSUE #505: CI takes 23 minutes and strict up-to-date turns N ready PRs into N serial re-runs
Landing one small fix currently costs 20–25 minutes of CI, and landing several costs that **per merge, serially**. Today four one-file fixes took most of a working session, almost none of it spent on the fixes. The cause is three things multiplying, and each is measured below rather than guessed.

## Where the time goes

From run `31367812235` (a normal green run on `main`):

| step | duration |
| --- | --- |
| checkout, Xcode select, dead-expect lint, Package.resolved check | 5s total |
| **Build** | **281s** |
| **Coverage report** (`swift test --enable-code-coverage --no-parallel`) | **1114s** |

**1395 of 1400 seconds are build + test.** Nothing else matters.

## Why it multiplies

**1. No build cache.** `ci.yml` never caches `.build` or the SwiftPM cache, so every run compiles the package and the whole test target from scratch. The 281s build is paid on every single run, including runs that differ from the previous one by a one-line test change.

**2. `--no-parallel` on a 3400-test suite.** The coverage step runs the entire suite on one thread. It is `--no-parallel` for a real reason — this repository has had order-dependent tests and shared-stdout corruption — but the cost is now 18½ minutes per run.

**3. `strict_required_status_checks_policy: true`.** Every merge into `main` marks every other open PR out of date. Each one must then merge `main`, which produces a new head, which re-runs the full 23 minutes. With N ready PRs the wait is not N × 23 minutes of useful work — it is N × 23 minutes of *re-running CI on code that already passed*, serialised.

On top of that, `concurrency: group: ${{ github.workflow }}-${{ github.ref }}` with `cancel-in-progress: true` does not always cancel the older run on the same PR, so a superseded run keeps the group and the current head sits at `pending` with zero jobs — indistinguishable, from `gh pr checks`, from "no CI configured".

## What would actually fix it

In order of effect per unit of risk:

**A. Cache the build.** `actions/cache` keyed on `Package.resolved` + Swift version, restoring `.build` and `~/.cache/org.swift.swiftpm`. A warm build should cut most of the 281s, and it removes the cost from every subsequent run rather than one.

**B. Split the suite from coverage.** Run `swift test` for the verdict and produce coverage only on `main` pushes or on demand. The gate needs the tests to pass; it does not need a coverage report on every PR head.

**C. Turn off the strict up-to-date requirement, or adopt a merge queue.** Strict mode is what turns N independent PRs into N serial CI runs. A merge queue keeps the safety — it tests the *combined* result — without re-running each PR after every unrelated merge. If a queue is too much, dropping `strict` is defensible here because the repository already requires a full local suite on the merged result before push.

**D. Parallelise selectively.** `--no-parallel` exists for known-bad interactions. Rather than flipping it globally, the suites that are order-sensitive can be marked `.serialized` and the rest allowed to run in parallel.

**E. Make a stuck run visible.** A `pending` run with zero jobs should not look like a missing configuration. Whatever watches CI should distinguish *no run*, *awaiting approval* (fork PRs park at `action_required` forever without it), *blocked by concurrency*, *running*, and *finished*.

A and B are contained changes to `ci.yml`. C is a repository setting and is the one that removes the quadratic term.

## Not proposed

Weakening what the gate checks. The suite must still run in full on the exact head that merges; the argument above is only about not running it repeatedly on code that has not changed.

MongLong0214:
Measured on run `31367812235`, a normal green run: of 1400 seconds, **Build took 281 and the coverage step took 1114**. Everything else totalled five. Nothing was cached, so both were paid in full on every run — including runs differing from the previous one by a single line.

## Two changes here

**Cache `.build` and the SwiftPM cache**, keyed on `Package.swift`, `Package.resolved` and the Xcode pin, so a dependency bump or toolchain change misses rather than restoring something stale.

**Split the verdict from the report.** `swift test` decides whether the branch is good; coverage instrumentation roughly doubles the run and is only read on `main`, so it now runs on pushes to `main` instead of sitting between every pull request and its answer.

The suite still runs **in full, on the exact head, with `--no-parallel`**. Nothing the gate checks is removed.

## The third term, done separately

`strict_required_status_checks_policy` was the multiplier: every merge into `main` marked every other open PR out of date, so N ready branches cost N serial 23-minute re-runs of code that had already passed. That setting is now off. The ruleset still requires the `build` check itself, and this repository already runs the full suite locally on the merged result before pushing.

## Deliberately not changed

`--no-parallel` stays. It guards order-dependent tests and shared-stdout corruption this repository has actually hit, and those are unfixed. Narrowing it to `.serialized` on the affected suites is the right follow-up and is described in #505.

Closes #505

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,6 +28,20 @@ jobs:
         # prove nothing. Fail fast (no Xcode needed) before the build if any reappear.
         run: bash Scripts/ci-forbid-dead-expect.sh
 
+      - name: Cache SwiftPM and build products
+        # 1395 of a 1400-second run is build + test and none of it was cached, so every
+        # run recompiled the package and the whole test target from scratch. The key
+        # pins the manifest, the resolved dependency set and the Xcode selection, so a
+        # dependency bump or toolchain change misses rather than restoring a stale tree.
+        uses: actions/cache@0400d5f644dc74513175e3cd8d07132dd4860809 # v4.2.4
+        with:
+          path: |
+            .build
+            ~/Library/Caches/org.swift.swiftpm
+          key: spm-macos15-xcode164-${{ hashFiles('Package.resolved', 'Package.swift') }}
+          restore-keys: |
+            spm-macos15-xcode164-
+
       - name: Select Xcode
         # Xcode 16.4 ships Swift 6.2 — required by swift-sdk 0.11.0+ which
         # adopts the short-form `withThrowingTaskGroup { ... }` syntax. The
@@ -40,7 +54,13 @@ jobs:
       - name: Verify Package.resolved is unchanged
         run: git diff --exit-code Package.resolved
 
+      - name: Test suite
+        # The gate's verdict. Coverage instrumentation roughly doubles the run and is
+        # only read on main, so it no longer sits between a pull request and its answer.
+        run: swift test -Xswiftc -suppress-warnings --no-parallel
+
       - name: Coverage report
+        if: ${{ github.event_name == 'push' }}
         # H-4 (2026-05-08 enterprise review) closed in v3.4.0: re-armed
         # the coverage gate. Pre-fix this step disabled shell error handling
         # and exited successfully, so coverage regressions never blocked


MongLong0214:
Isolated: the cache half alone is accepted (#509, run dispatched normally). So the element GitHub rejects at parse time is the coverage split — either the added `Test suite` step or the `if:` on `Coverage report`.

Superseded by #509 for the cache. Closing this so it stops occupying a slot; the coverage split returns as its own change once the rejected element is identified.

PR #513, opened 2026-08-10
DO NOT MERGE — proof that the required aggregator fails closed

MongLong0214:
Temporary. A deliberate compile error, opened only to observe whether the `build` aggregator reports **failure** or silently reports **skipped** when a dependency fails. Closed once the result is recorded on #512. Nothing here is intended for main.


MongLong0214:
Result recorded on #512: the aggregator reported **failure**, not skipped. Purpose served.

PR #514, opened 2026-08-10
fix(#289): a refresh that finished late must not overwrite newer state

MongLong0214:
Closes part of #289 (ADR-006). This is the correctness half; single-flight, cancellation and
notification backpressure are not in this PR.

## The defect

`StateCache` assigned every section write unconditionally, and `StatePoller` called those writes
without comparing against anything — no request sequence, no epoch, no revision, no observation time.
A refresh that **started** before a change and **finished** after it therefore replaced the newer
value with the older one. `VersionedSnapshot` has carried `projectEpoch` and `sectionRevision` since
it was written; the write path simply never read them.

## The fix

Section writes now take the `(projectEpoch, sectionRevision)` observed at the moment the read
**started**, and apply the value only when both still match. Observing after the read returns would
make a stale result look current and reintroduce the exact race being closed, so the capture is
before `execute`, not after. Rejections are counted per section — a guard that leaves no trace cannot
be tested.

Not behind a feature flag. Overwriting fresh state with stale state is never desired, and a
correctness fix that ships disabled fixes nothing.

## A defect this fix introduced, and removed

The first version returned the write outcome from `poll()`. That value feeds
`projectReady`/`tracksReady` → `hasDocument`, and after `failureThreshold` consecutive false results
the poller calls `updateDocumentState(false)` — it declares the user's project **closed**. Since the
guard fires precisely during concurrent activity, that would have been a worse failure than the race.
`poll()` now answers "is this section readable", which a rejection does not contradict: a rejected
write means the cache already holds something newer.

`Issue289PollerReadinessTests` locks this. Reverting `poll()` to return the write outcome makes it
fail — verified by running the mutation, not by inspection.

## What the live run does and does not show

Driven against the release artifact built from this branch, screen-recorded, with the track rail
asserted on a bound region: a rename reaches State A through the guarded write path, is **visible on
screen** rather than only in the envelope, and the restore is visible too.

It does **not** demonstrate the race. The revert probe — rename a track, then read the tracks
resource 40 times across several refresh cycles looking for the old name returning — was run against
**both** binaries:

| binary | samples | reverts observed |
| --- | ---: | ---: |
| pre-fix (`origin/main`) | 40 | **0** |
| post-fix (this branch) | 40 | 0 |

The pre-fix binary never reverted either, so a zero on this branch separates nothing. The probe has no
power to distinguish them at the timing this machine produces through the public surface, and the
evidence records that rather than crediting the fix with a pass it did not earn. The race is
demonstrated at the unit level, where the interleaving can actually be forced.

## Stated limit

Markers stay on the unversioned path: `CacheSectionID` has no `markers` case, so there is no revision
to compare. Markers remain exposed to this race. That is written down here rather than left for
someone to discover.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,13 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/) 
 
 ## [Unreleased]
 
+### Added
+- `logic://` resource envelopes served under `LOGIC_MCP_ADR006_VERSIONED_CACHE=1` now carry
+  `dropped_stale_writes` beside `section_revision`: the number of refreshes whose value was refused
+  for that section because it finished after the section had already moved on (#289). The field is
+  always present and starts at zero, so a client can tell "the guard never had to fire" apart from
+  "this build does not report it".
+
 ### Changed — three operations lose their fallback channels
 
 Each of these narrows an operation to a single channel. In an environment where that
--- a/Sources/LogicProMCP/Resources/ResourceHandlers+CacheEnvelope.swift
+++ b/Sources/LogicProMCP/Resources/ResourceHandlers+CacheEnvelope.swift
@@ -68,11 +68,18 @@ extension ResourceHandlers {
     static func versionedCacheExtras(
         projectEpoch: UInt64,
         sectionRevision: UInt64,
+        droppedStaleWrites: UInt64 = 0,
         bodyJSON: String
     ) -> [String: Any] {
         [
             "project_epoch": projectEpoch,
             "section_revision": sectionRevision,
+            // How many times a refresh finished so late that its value was refused for this section
+            // (#289). A guard whose only evidence is the absence of a symptom cannot be checked: the
+            // symptom is transient and rare, so "we saw no corruption" is equally consistent with the
+            // guard working and with it never running. Counting refusals turns that into something a
+            // client — or a live test — can observe directly at the boundary where it matters.
+            "dropped_stale_writes": droppedStaleWrites,
             "etag": etag(of: bodyJSON),
         ]
     }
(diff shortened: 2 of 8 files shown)


MongLong0214:
**Root-caused why no live probe could observe this race, with measurements.**

I could not produce a positive observation of a stale overwrite on the pre-fix binary. Rather than
report that as "not reproducible", I instrumented the poller and measured the window.

| measured | value |
| --- | --- |
| poll cycle | ~5 s (3 s sleep + cycle work) |
| the poller's tracks read — the whole window a competing write must land inside | **107–150 ms** |
| that window as a share of the cycle | **~2.4 %** |
| other activity logged inside six consecutive windows | **0 lines** |
| one `track.rename` | ~170 ms |
| one `track.select` | ~634 ms |
| attempts made | 90 renames, 71 selections, 20 poll cycles |
| poll writes refused | **0** — all 11 instrumented writes applied |

The fastest write the shipped MCP surface offers is **slower than the window it would have to hit**.
That is why nothing collided: not chance, and not a missing race, but a client that cannot issue a
write quickly enough to interleave. My earlier "555 ms tracks read" figure was wrong — that was a tool
round-trip, not the poller's raw read; the real window is four times narrower.

Two consequences, both now reflected in the PR:

1. **This is cheap insurance on a narrow window, not a fix for something users hit today.** The
   original PR text implied more than the evidence supports, and is corrected.
2. **The counter, not the symptom, is the thing worth exposing.** A guard whose only evidence is the
   absence of a rare transient cannot be checked — silence is equally consistent with "working" and
   "never ran". Envelopes under `LOGIC_MCP_ADR006_VERSIONED_CACHE=1` now carry
   `dropped_stale_writes` per section, always present and starting at zero. If a later change widens
   the window — an `await` added between the read and the write, a shorter poll interval, a writer
   that is not AX-bound — the number starts moving and somebody sees it.

The race remains real in the code: nothing compared versions before this change. What the measurement
establishes is its reachability, which is much lower than I claimed.

Issue #527, opened 2026-08-11
navigate.create_marker drives Navigate > Create Marker but no marker appears; two readback paths disagree about the same project state


MongLong0214:
### `navigate.create_marker` reports success while its own readback shows nothing was created

Driven through the release artifact against a project with 13 markers:

```json
{"operation":"nav.create_marker","method":"accessibility_menu",
 "menu_path":"Navigate > Create Marker","sent":true,
 "marker_count_before":13,"marker_count_after":13,
 "marker_count_before_channel":13,"marker_count_after_channel":13,
 "reason":"readback_mismatch","state":"B","success":true}
```

The marker count is unchanged — measured by the operation itself, twice, from two sources — and the
envelope still carries `success: true`. Re-reading `logic://markers` afterwards confirms 13. The
Marker List on screen confirms 13.

### What is right about it

`reason: "readback_mismatch"` is honest, and State B is the correct grade for "the write boundary was
crossed and the result cannot be confirmed". The operation is not lying about its evidence.

### What is wrong

`success: true` alongside `marker_count_before == marker_count_after` is a contradiction the caller
cannot be expected to unpick. The operation **has** the counts. It compared them — that is what
produced `readback_mismatch`. Having established that nothing changed, reporting success is the one
thing the envelope should not do.

This is the same shape as #523, where the caller was told success for a marker that was never
deleted, and it is worth treating as one class rather than two incidents: **an operation that
measures its own before/after and finds them equal should not answer `success: true`.**

### Note on the earlier marker work

`create_marker` did succeed earlier in the same session — a marker `마커 15` appeared in the Marker
List while Logic was in Korean, so the menu path does work under some conditions. What changed
between then and now has not been established, and that is precisely why the mismatch should be
surfaced as a failure rather than a success: it is a real, intermittent condition, not a readback
weakness.

### Acceptance

- an unchanged before/after count yields `success: false` (State B may stay, with the reason);
- an actual creation still reports success with the count advancing by one;
- the condition under which the menu path stops creating is identified, or the operation says which
  of its two readback sources disagreed.

MongLong0214:
**Correcting this issue. Its title and central claim are wrong; the underlying defect is real but
different.**

I reported that `create_marker` "answers `success: true` while its own counts are equal". I read the
JSON body's `success` field and did not check the MCP-level flag beside it. Measured through a raw
MCP client:

```
MCP isError: True
[...]

Issue #526, opened 2026-08-11
ChannelRouter falls through on a deliberate safety refusal, so a guarded operation is reported as succeeding via a weaker channel


MongLong0214:
### The router converts a deliberate safety refusal into a reported success

`ChannelRouter` tries channels in order and moves to the next when one returns an error. It does not
distinguish **"this channel could not act"** from **"this channel refused on purpose because acting
would have been unsafe"**.

Observed on `nav.delete_marker`, routed `[.accessibility, .midiKeyCommands, .cgEvent]`:

```
[router] nav.delete_marker failed via Accessibility:
  {"error":"ax_write_failed",
   "hint":"The Marker List did not hold keyboard focus, so Delete was not pressed —
           the same key deletes a region or a track when focus is in the arrange area.",
   "write_attempted":false,"state":"C"}
  , trying next
[router] nav.delete_marker succeeded via MIDIKeyCommands
```

The caller receives `success: true, state: B`. Nothing was deleted. The refusal that protected the
user's regions and tracks was discarded, and the operation was reported as having been attempted
successfully by a channel that fired a MIDI CC nothing is bound to.

### Why this is more than one operation's problem

`write_attempted: false` on a State C is the Honest Contract's way of saying **the write boundary was
not crossed and the channel decided not to cross it**. That is a verdict about the operation, not a
report of channel weakness. Falling through past it means:

- the safest channel's judgement is overridden by a less capable one;
- the caller is told `success` for an operation the product declined to perform;
- the `state: B` that results is indistinguishable from a genuine attempted-but-unverifiable write.

A channel that says "I cannot reach this surface" should fall through. A channel that says "I reached
it and refused" should not.

### Where it was found

Fixing #523. That operation no longer needs the fallback — it now acts through the Marker List's own
Edit menu — but the routing behaviour that turned its refusal into a success is unchanged and applies
to every operation with more than one channel.

### Suggested shape

Treat a State C carrying `write_attempted: false` **together with a refusal-class error** as terminal
for the operation rather than as a channel failure. `element_not_found` or a channel that never
started are genuine fall-through cases; `ax_write_failed` raised by a precondition guard is not.

The distinction has to be explicit in the envelope rather than inferred from the error string — the
router should not be pattern-matching hints.

### Acceptance

- an operation whose first channel refuses with a guard returns that refusal to the caller;
- an operation whose first channel is genuinely unavailable still falls through;
- a test that a refusal is not converted into a later channel's success;
[...]

PR #528, opened 2026-08-11
fix(#523): delete a marker through the list's own menu, not a global keystroke

### LINKED ISSUE #523: navigate.delete_marker selects the row and returns State B without deleting it
### Observed

`navigate.delete_marker` at a **valid** index reports success in State B and the marker is still
there. Driven through the release artifact against a project with 15 markers:

```
delete_marker { index: 14 }  →  success: true   state: B   verified: false
```

The Marker List afterwards still lists all 15, with the target row **selected** but not deleted. No
confirmation dialog appeared. Repeated three times with the same result.

Out-of-range is handled correctly: `index: 15` returns State C `element_not_found`, so the index
contract itself is 0-based and enforced.

### What this is not

- **Not a false State A.** The implementation is careful: it selects the row, refuses unless the
  Marker List holds keyboard focus (the same keystroke deletes a region or a track elsewhere), posts
  `kVK_Delete`, then requires **two consecutive identical readings** before certifying and reports
  State B when they will not settle. The envelope is honest — `verified: false` says exactly that.
- **Not a blocked key post.** Measured in the same session: a `CGEvent` key post does reach Logic —
  sending Space flipped `logic://transport/state` `isPlaying` from `false` to `true`. So the channel
  the delete uses is live.
- **Not locale.** Observed with Logic in English.

### What is unresolved

Why the Delete keystroke selects the row and does not remove it. It gets past the keyboard-focus
guard, so focus is where the code requires. I have not established whether Logic ignores the key for
this table, whether the marker needs a different actuation, or whether the focus guard passes while
focus is somewhere the key does not act.

### Why it is worth a ticket even though the envelope is honest

A caller sees `success: true`. State B means "attempted, unverifiable", and a caller that treats
success as done will believe a marker was removed that is still in the project. If this operation can
never verify on this surface, that is worth knowing — an operation that is structurally unable to
reach State A should say so once rather than return State B forever.

### Reproduction

1. A project with several markers.
2. `navigate.delete_marker { index: <a valid one> }`.
3. Read `logic://markers`, or look at the Marker List.

Found while cleaning up a marker created during a Korean-locale sweep (#519); the marker `마커 15` is
still in that test project as a result.

MongLong0214:
Closes #523.

## The cause was not the one the issue guessed

`nav.delete_marker` answered `success: true, state: B` while the marker stayed. The server log named
it:

```
[router] nav.delete_marker failed via Accessibility:
  {"error":"ax_write_failed",
   "hint":"The Marker List did not hold keyboard focus, so Delete was not pressed —
           the same key deletes a region or a track when focus is in the arrange area.",
   "write_attempted":false,"state":"C"}, trying next
[router] nav.delete_marker succeeded via MIDIKeyCommands
```

The accessibility path **refused on purpose**, and the router read that deliberate
`write_attempted: false` refusal as a failure, fell through to the MIDI key-command channel, and that
channel fired a CC nothing is bound to and reported success. A safety refusal was being converted
into a reported success for an operation that never happened.

## The fix

The Marker List window has its own Edit menu whose `Delete` entry is aimed at that list and needs no
keyboard focus. Acting through it removes the reason to be in that position. The keystroke and its
focus guard remain as the fallback, and the refusal message now says both routes were unavailable
rather than naming only focus.

`Delete` is matched **exactly** against a locale policy entry. The same menu offers
`Delete Undo History`, which a prefix match reaches first — not hypothetical: it ran during this
investigation and discarded a project's undo history. This repository had already recorded the
identical trap for the Undo prefix.

No coordinate or mouse event is involved: the menu button is found by label and opened with
`AXShowMenu`.

## Live evidence

| check | result |
| --- | --- |
| the accessibility path handles it — no fall-through to the key-command channel | `actuation: marker_list_edit_menu`, `method: nil` |
| the count drops by exactly one, from two settled readings | `14 → 13`, `readback_settled: true` |
| an independent read agrees the marker is gone | back to the baseline the run started from |
| the row disappears from the Marker List **on screen** | region-bound visual assertion |

Screen-recorded. The marker the run deletes is created by the run itself, through the Marker List's
own `+` button — **not** through `nav.create_marker`, because every marker in the project sat at bar 1 and Logic
will not stack a second one there — so the run creates the marker it deletes rather than deleting one
the project already had.

## Two findings this surfaced, filed separately

- **#526** — the router converts a deliberate safety refusal into a fallback success. That applies to
  every multi-channel operation, not just this one, and is not addressed here.
- ~~**#527**~~ — I filed this claiming `nav.create_marker` was broken. It is not: Logic refuses a
  second marker at a position that already has one, and every marker in the test project sat at
  bar 1. Measured at an empty bar it answers State A, verified. The issue is closed.

An earlier note here said `logic://markers` disagreed with the operation's own reading. That does not
reproduce — a macOS consent dialog had been sitting over the screen during those readings. The
evidence remains keyed to the product's own measurement, which is the stricter source either way.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,15 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/) 
 ## [Unreleased]
 
 ### Fixed
+- **`navigate.delete_marker` deletes the marker, and says so honestly (#523).** It reported
+  `success: true, state: B` while the marker stayed. The accessibility path's safety guard was
+  refusing — correctly, because the same keystroke deletes a region or a track when focus is not in
+  the Marker List — and the router treated that deliberate `write_attempted: false` refusal as a
+  failure, fell through to the MIDI key-command channel, and that channel fired a CC nothing is bound
+  to and answered success. The operation now acts through the Marker List window's own Edit ▸ Delete,
+  which is aimed at that list and needs no keyboard focus; the keystroke and its guard remain as the
+  fallback. Verified live: `state A`, `verified true`, 15 markers to 14.
+
 - **The MIDI readback collector can resolve the Event List display mode (#524).** It searched the
   **application** menu bar for `Event Position and Length as Time`; that menu has sixteen entries on
   Logic 12.3 and none of them is the setting, which lives in the Event pane's own View menu beside the
--- a/Sources/LogicProMCP/Accessibility/AXLocalePolicy.swift
+++ b/Sources/LogicProMCP/Accessibility/AXLocalePolicy.swift
@@ -212,6 +212,17 @@ enum AXLocalePolicy {
         rationale: "Reveals the New Project chooser, or creates the project directly where Logic skips it."
     )
 
+    /// The plain `Delete` entry on a list window's Edit menu.
+    ///
+    /// Matched with `.exactStrict`, never as a prefix: the same menu offers `Delete Undo History`,
+    /// and a prefix match reaches it first. That is not a hypothetical — it happened during the
+    /// investigation of #523 and discarded a project's undo history.
+    static let deleteMenuItem = LabelSet(
+        canonical: "Delete",
+        variants: ["삭제"],
+        rationale: "Removes the selected row from a list window; exact match only."
+    )
+
     static let editMenuBar = LabelSet(
         canonical: "Edit",
         variants: ["편집"],
--- a/Sources/LogicProMCP/Channels/AccessibilityChannel+MarkerDelete.swift
+++ b/Sources/LogicProMCP/Channels/AccessibilityChannel+MarkerDelete.swift
@@ -80,19 +80,37 @@ extension AccessibilityChannel {
             ))
         }
 
-        guard markerTableHasKeyboardFocus(runtime: runtime) else {
-            // Refusing here is the point: the same keystroke elsewhere deletes a region or a track.
-            extras["write_attempted"] = false
-            return .error(HonestContract.encodeStateC(
-                error: .axWriteFailed,
-                hint: "The Marker List did not hold keyboard focus, so Delete was not pressed — "
-                    + "the same key deletes a region or a track when focus is in the arrange area.",
-                extras: extras
-            ))
+        // The window's own Edit ▸ Delete is aimed at THIS list and needs no keyboard focus, so it is
+        // tried first. The keystroke below does need focus, and its guard was refusing nearly every
+        // call — measured (#523). The refusal is right in itself; what it exposed is worse. The
+        // router treats a deliberate `write_attempted: false` State C as a failure and falls through
+        // to the MIDI key-command channel, which fires a CC nothing is bound to and answers
+        // `success: true, state: B`. A safety refusal was being turned into a reported success for an
+        // operation that never happened. Acting through the menu removes the reason to be there.
+        //
+        // The entry is matched EXACTLY: the same menu offers `Delete Undo History`, which a prefix
+        // match reaches first — that is not hypothetical, it discarded a project's undo history while
+        // this was being investigated.
+        if deleteViaMarkerListEditMenu(in: window, runtime: runtime) {
+            extras["write_attempted"] = true
+            extras["actuation"] = "marker_list_edit_menu"
+        } else {
+            guard markerTableHasKeyboardFocus(runtime: runtime) else {
+                // Refusing here is the point: the same keystroke elsewhere deletes a region or a track.
+                extras["write_attempted"] = false
+                extras["actuation"] = "refused_no_keyboard_focus"
+                return .error(HonestContract.encodeStateC(
+                    error: .axWriteFailed,
+                    hint: "The Marker List's Edit menu offered no exact Delete entry and the list did "
+                        + "not hold keyboard focus, so Delete was not pressed — the same key deletes "
+                        + "a region or a track when focus is in the arrange area.",
+                    extras: extras
+                ))
+            }
+            extras["write_attempted"] = true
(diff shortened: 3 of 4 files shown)


MongLong0214:
**Not merging this. An independent review blocked it, and I have confirmed the critical finding in
the code.**

## Critical — the AX return code can drive a second deletion

`deleteViaMarkerListEditMenu` ends with:

```swift
return AXHelpers.performAction(entry, kAXPickAction as String, runtime: runtime.ax)
```

and the caller treats `false` as "nothing happened":

```swift
if deleteViaMarkerListEditMenu(in: window, runtime: runtime) {
    ...
} else {
    guard markerTableHasKeyboardFocus(runtime: runtime) else { ...refuse... }
    _ = mouse.postKeyEvent(0x33)   // presses Delete
}
```

This repository's own recorded rule is that **AX return codes lie in both directions** — `rc = 0`
with no effect and `rc = -25204` after a successful effect are both observed here. So if `AXPick`
delivers the deletion and still reports false:

- with the table focused, the code presses Delete **again** and deletes a second marker;
- without focus, it returns `write_attempted: false` for a write that was in fact attempted and may
  have landed.

The first is destructive. Once `AXPick` has been issued the operation is State B-or-A pending
readback, and no other actuator may be selected on the strength of the return value.

## Also raised, and I accept them

- **The original fallback problem is still reachable.** When the menu entry is not found and focus is
  unsafe, the refusal is still a non-terminal `ax_write_failed`, so the router still falls through to
  the MIDI channel. That is #526, and its fix (`fallback_unsafe`) is now in #530 — this branch should
  build on it rather than leave the hole open.
- **Japanese is missing.** `deleteMenuItem` carries only `Delete` and `삭제`. I will measure the
  Japanese labels from a Japanese Logic rather than translate them, the way the other locale sets in
  this repository were produced.
- **`safe_to_retry` is absent** from the paths this branch touches.

## And the tests do not discriminate

The review's sharpest point: every assertion in `Issue523MarkerDeleteMenuTests` still passes if the
behavioural menu-first hunk is reverted while the policy constant stays. Asserting that a two-element
label list does not contain `"Delete Undo History"` cannot detect production using a prefix match —
it tests the literal I just wrote, not the code path.

That is the same failure mode this repository already has a CI check for, one level up: not a dead
assertion, but a live assertion aimed at nothing.

## Plan

#530 (#526) lands first. Then this branch is rebuilt on it: the pick result becomes an outcome that
distinguishes "issued" from "never issued" so the return code can never select a second actuator,
Japanese labels get measured live, and the tests are rewritten against an AX tree so they fail when
the behaviour is reverted. Re-verified live and re-reviewed before it goes anywhere near main.

MongLong0214:
**Japanese labels measured on a Japanese Logic, not translated.** Switched Logic Pro 12.3 with
`defaults write com.apple.logic10 AppleLanguages -array ja`, restarted, drove it, and restored
English afterwards.

Why this had to be measured: `Navigate` is `移動` — and the `Go To` submenu inside it is **also**
`移動`. `Mix` is `ミックス` in Japanese while it stays `Mix` in Korean. Nothing here is derivable.

## What the delete path needs

Marker List's own Edit menu, in order:

```
取り消す- マーカーを作成 | やり直しできません | 取り消し履歴… | 取り消し履歴を削除 |
カット | コピー | ペースト | 置き換えてペースト | オリジナルの位置にペースト | 削除 |
すべてを選択 | 後続のものすべてを選択
```

`削除` is Delete. **The trap is worse in Japanese than in English:**

- a prefix match on `削除` also matches `削除して移動` (Delete and Move, in the main Edit menu);
- a *contains* match on `削除` matches `取り消し履歴を削除` — Delete Undo History, the entry that
  actually discarded a project's undo history during this investigation.

In English only the prefix direction was dangerous. In Japanese both are. `.exactStrict` is not
belt-and-braces here, it is the only safe mode.

## Everything else measured

| what | Japanese |
| --- | --- |
| menu bar | `ファイル · 編集 · トラック · 移動 · 録音 · ミックス · 表示 · ウインドウ · ヘルプ` |
| File ▸ New | `新規` |
| Go To ▸ Position… | `移動 ▸ 移動 ▸ 位置…` |
| Go To Position dialog | **`位置の移動`** (buttons `キャンセル` / `OK`, `AXFloatingWindow`) |
| Marker List window | `名称未設定 - マーカーリスト` |
| marker container `AXDescription` | **`マーカー`** |
| marker count `AXDescription` | `項目数`, value `N個のマーカー` |
| Marker List toolbar menus | `編集 · オプション · 表示` |
| Marker List columns | `L · ポジション · マーカー名 · 長さ` |
| Event List tabs | `イベント · マーカー · テンポ · キー/拍子` |
| Event List region-level headers | `L · M · ポジション · 名前 · Trk · 長さ` |
| arrange window | `名称未設定 - トラック` (`Untitled` = `名称未設定`) |

**Not measured, so not to be written into a policy: the note-level Event List headers.** The project
was empty, so there was no MIDI region to open. Anyone adding those must measure them.

## A defect this turned up

`project.new` fails on Japanese Logic. It reported

```
channels_exhausted / phase: unexpected_blocking_sheet / sheet_kind: unknownSheet / state B
```

while the project window `名称未設定 - トラック` had in fact been created. The "unknown" sheet was the
ordinary `新規トラックを作成` (New Tracks) dialog every new project shows, with buttons `作成` and
`キャンセル`. The sheet classifier does not recognise it in Japanese, so it fails closed on a normal
startup dialog. Filing separately.

MongLong0214:
Superseded by a rebuild on merged main. The blocking defect here — returning `AXPick`'s Bool so a false negative pressed Delete a second time — is fixed by an outcome type rather than patched, and the tests now run against an AX tree instead of asserting properties of a label constant. New PR replaces this one.

PR #530, opened 2026-08-11
fix(#526): let a channel forbid fallback when trying another one would be unsafe

### LINKED ISSUE #526: ChannelRouter falls through on a deliberate safety refusal, so a guarded operation is reported as succeeding via a weaker channel

MongLong0214:
Closes #526.

## The defect

`ChannelRouter` decided whether to fall through to the next channel by asking whether the error
**code** was terminal. A channel that refuses an operation for safety emits `ax_write_failed` with
`write_attempted: false` — which by code alone is indistinguishable from "this channel could not
reach the control, try another". So a deliberate refusal was routed to a weaker channel, which fired
blindly and reported success.

```
[router] nav.delete_marker failed via Accessibility:
  {"error":"ax_write_failed","write_attempted":false,"state":"C"}, trying next
[router] nav.delete_marker succeeded via MIDIKeyCommands
```

`ax_write_failed` has to stay non-terminal: for other operations another channel genuinely can
deliver. The information the router is missing is not *what went wrong* but *may another channel try
at all*.

## The change

A State C envelope may carry `fallback_unsafe: true`, and the router returns such an envelope
verbatim instead of continuing. Absent or false, behaviour is byte-for-byte what it was.

Deliberately narrow, and each of these was a live option that was rejected:

- **`safe_to_retry` is not reused.** It answers whether the *caller* may retry. A safety refusal is
  usually safe for the caller to retry after fixing focus while still being unsafe to route to
  another channel. Widening a shared field to carry a second meaning would have made both weaker.
- **Nothing is added to `terminalErrorCodes`.** The code is the wrong axis; a code that is
  legitimately recoverable elsewhere must stay recoverable.

The marker-delete focus refusal is the single call site that sets it.

## Live evidence

Release artifact built from this head, driven over real stdio MCP against Logic Pro 12.3, with the
Marker List deliberately not holding keyboard focus.

| | guard on (this branch) | guard off (same source, one line) |
| --- | --- | --- |
| MCP `isError` | `true` | `false` |
| envelope | State C verbatim, `fallback_unsafe: true`, `write_attempted: false` | `success: true`, State B, `method: midi_key_command`, CC 45 CH 16 |
| Logic's own marker count | 15 → 15 | 15 → 15 — **nothing was deleted either way** |

The negative control is the point: with the guard removed the caller is told a delete was attempted
and is retryable, while a CC nothing is bound to was fired and no marker moved.

Evidence document: 3 checks, all three mutation-backed; 3 settled captures wholly within one display;
1 visual assertion; 1 screen recording; marker count restored. The visual assertion expects the row
region to be **unchanged**, so it was separately controlled — adding one marker through the Marker
List's own `+` button changes that same region's pixel hash, which is what makes "unchanged" a
result rather than an artefact.

## Tests

`Issue526FallbackUnsafeTests` — three cases: a multi-channel chain whose first channel sets the flag
never executes the second and returns verbatim; the same chain without the flag still falls through;
the flag alongside an already-terminal code stays verbatim.

Mutation-tested rather than assumed: disabling the router guard fails the first test on four
assertions, including `keyCommands.executions() == 0`. The other two still pass under that mutation —
they are regression guards for the paths this change must not disturb, not evidence for it.

Full suite: 3476 tests pass.

### DIFF
--- a/Sources/LogicProMCP/Channels/AccessibilityChannel+MarkerDelete.swift
+++ b/Sources/LogicProMCP/Channels/AccessibilityChannel+MarkerDelete.swift
@@ -72,7 +72,10 @@ extension AccessibilityChannel {
         ]
 
         guard selectMarkerRowForDeletion(index, in: window, runtime: runtime.ax) else {
+            // A weaker key-command channel cannot answer an indexed delete: it ignores `index`
+            // and fires CC 45 blindly, without proving which marker would be removed.
             extras["write_attempted"] = false
+            extras["fallback_unsafe"] = true
             return .error(HonestContract.encodeStateC(
                 error: .axWriteFailed,
                 hint: "Marker index \(index) could not be selected, so nothing was pressed",
--- a/Sources/LogicProMCP/Channels/AccessibilityChannel+MarkerDelete.swift
+++ b/Sources/LogicProMCP/Channels/AccessibilityChannel+MarkerDelete.swift
@@ -73,11 +73,19 @@ extension AccessibilityChannel {
         let expectedSurvivors = expectedSurvivorMarkers
             .map { canonicalMarkerIdentity(name: $0.name, position: $0.position) }
             .sorted()
+        let targetPositionMatchCount = before.filter { $0.position == target.position }.count
+        let targetPositionUnique = targetPositionMatchCount == 1
+        // A parse fallback manufactures `ordinal + 1.1.1.1`, which can happen to equal a
+        // different marker's parsed position. Do not let a synthetic position participate in a
+        // State-A identity proof, even if it happens not to collide in this particular reading.
+        let prewritePositionEvidenceCanonical = before.allSatisfy { $0.positionSource == .parser }
         var extras: [String: Any] = [
             "operation": "nav.delete_marker",
             "requested_index": index,
             "target_name": target.name,
             "target_position": target.position,
+            "target_position_unique": targetPositionUnique,
+            "prewrite_position_evidence_canonical": prewritePositionEvidenceCanonical,
             "marker_count_before": before.count,
         ]
 
@@ -121,6 +129,7 @@ extension AccessibilityChannel {
         // report State B when they will not settle rather than certifying a guess.
         var observedSurvivorPositions: [String] = []
         var observedSurvivors: [String] = []
+        var observedPositionEvidenceCanonical = false
         var settled = false
         var previousPositions: [String]?
         for _ in 0..<6 {
@@ -133,6 +142,9 @@ extension AccessibilityChannel {
                 observedSurvivors = reading
                     .map { canonicalMarkerIdentity(name: $0.name, position: $0.position) }
                     .sorted()
+                observedPositionEvidenceCanonical = reading.allSatisfy {
+                    $0.positionSource == .parser
+                }
                 settled = true
                 break
             }
@@ -145,6 +157,9 @@ extension AccessibilityChannel {
         }
         extras["readback_settled"] = true
         extras["marker_count_after"] = observedSurvivorPositions.count
+        extras["observed_position_evidence_canonical"] = observedPositionEvidenceCanonical
+        extras["position_evidence_canonical"] = prewritePositionEvidenceCanonical
+            && observedPositionEvidenceCanonical
 
         // This proves that precisely one position occurrence disappeared and that it was an
         // occurrence at the target's position. It cannot establish WHICH marker was removed when
@@ -162,6 +177,18 @@ extension AccessibilityChannel {
         // position multiset in State A. Names are not exposed here because Logic may renumber them.
         extras["expected_survivor_position_multiset"] = expectedSurvivorPositions.joined()
         extras["observed_survivor_position_multiset"] = observedSurvivorPositions.joined()
+        guard targetPositionUnique else {
+            extras["reason_detail"] = "Delete was attempted and the marker count moved, but "
+                + "\(targetPositionMatchCount) pre-write markers shared target position "
+                + "\(target.position), so this readback cannot establish which marker was deleted."
+            return .success(HonestContract.encodeStateB(reason: .readbackUnavailable, extras: extras))
+        }
+        guard prewritePositionEvidenceCanonical, observedPositionEvidenceCanonical else {
+            extras["reason_detail"] = "Delete was attempted and the marker count moved, but at "
+                + "least one marker position came from an ordinal parse fallback; synthetic "
+                + "positions cannot support a verified target identity."
+            return .success(HonestContract.encodeStateB(reason: .readbackUnavailable, extras: extras))
+        }
         return .success(HonestContract.encodeStateA(extras: extras))
     }
 
--- a/Sources/LogicProMCP/Channels/AccessibilityChannel+MarkerDelete.swift
+++ b/Sources/LogicProMCP/Channels/AccessibilityChannel+MarkerDelete.swift
@@ -58,10 +58,20 @@ extension AccessibilityChannel {
             ))
         }
 
-        // Identify the survivor set by name+position rather than by index: indices renumber when a
-        // row disappears, so comparing indices after the write would prove nothing.
-        let expectedSurvivors = before.filter { $0.id != index }
-            .map { "\($0.name)@\($0.position)" }
+        // Indices renumber when a row disappears, so comparing them after the write would prove
+        // nothing. Instead, expect the pre-write POSITION multiset with exactly one instance of
+        // the target's position removed. Position entries are length-prefixed before sorting and
+        // joining, so their concatenation has unambiguous boundaries.
+        let expectedSurvivorMarkers = before.filter { $0.id != index }
+        let expectedSurvivorPositions = expectedSurvivorMarkers
+            .map(\.position)
+            .map(canonicalMarkerPosition)
+            .sorted()
+        // Names are deliberately not part of the State-A proof: Logic renumbers its generated
+        // marker names after a delete. Retain unambiguous name/position identities for a State-B
+        // mismatch diagnostic, where they help a human understand what was read back.
+        let expectedSurvivors = expectedSurvivorMarkers
+            .map { canonicalMarkerIdentity(name: $0.name, position: $0.position) }
             .sorted()
         var extras: [String: Any] = [
             "operation": "nav.delete_marker",
@@ -109,19 +119,24 @@ extension AccessibilityChannel {
         // name. Verifying against a single read produced a State A for a marker that was still
         // there. So require two consecutive identical readings before believing either of them, and
         // report State B when they will not settle rather than certifying a guess.
-        var survivors: [String] = []
+        var observedSurvivorPositions: [String] = []
+        var observedSurvivors: [String] = []
         var settled = false
-        var previous: [String]?
+        var previousPositions: [String]?
         for _ in 0..<6 {
             let reading = AXLogicProElements.enumerateMarkersFromListWindow(
                 afterBinding.window, runtime: runtime.ax
-            ).map { "\($0.name)@\($0.position)" }.sorted()
-            if let previous, previous == reading {
-                survivors = reading
+            )
+            let positions = reading.map(\.position).map(canonicalMarkerPosition).sorted()
+            if let previousPositions, previousPositions == positions {
+                observedSurvivorPositions = positions
+                observedSurvivors = reading
+                    .map { canonicalMarkerIdentity(name: $0.name, position: $0.position) }
+                    .sorted()
                 settled = true
                 break
             }
-            previous = reading
+            previousPositions = positions
             mouse.sleepMicros(250_000)
         }
         guard settled else {
@@ -129,16 +144,42 @@ extension AccessibilityChannel {
             return .success(HonestContract.encodeStateB(reason: .readbackUnavailable, extras: extras))
         }
         extras["readback_settled"] = true
-        extras["marker_count_after"] = survivors.count
+        extras["marker_count_after"] = observedSurvivorPositions.count
 
-        guard survivors == expectedSurvivors else {
-            // Either the target survived, or something else went with it. Both are mismatches, and
-            // saying which is more useful than a bare count.
-            extras["observed_survivors"] = survivors
+        // This proves that precisely one position occurrence disappeared and that it was an
+        // occurrence at the target's position. It cannot establish WHICH marker was removed when
+        // multiple markers share that position; the position multiset contains no such identity.
+        guard observedSurvivorPositions.count == before.count - 1,
+              observedSurvivorPositions == expectedSurvivorPositions else {
+            // Either the target position survived, or another position disappeared. The
+            // name-carrying diagnostics make that mismatch useful to a human without making names
+            // part of the stable State-A comparison.
+            extras["observed_survivors"] = observedSurvivors
             extras["expected_survivors"] = expectedSurvivors
             return .success(HonestContract.encodeStateB(reason: .readbackMismatch, extras: extras))
         }
+        // Bind the request-derived position-multiset expectation to the independently read-back
+        // position multiset in State A. Names are not exposed here because Logic may renumber them.
+        extras["expected_survivor_position_multiset"] = expectedSurvivorPositions.joined()
+        extras["observed_survivor_position_multiset"] = observedSurvivorPositions.joined()
         return .success(HonestContract.encodeStateA(extras: extras))
+    }
+
+    /// An unambiguous element of a position multiset.
+    ///
+    /// The wire form is `<position UTF-8 byte count>:<position>`. Sorting these elements and
+    /// concatenating them creates an unambiguous multiset encoding without trusting a delimiter.
+    private static func canonicalMarkerPosition(_ position: String) -> String {
+        "\(position.lengthOfBytes(using: .utf8)):\(position)"
+    }
+
+    /// An unambiguous, stable marker identity for survivor-set comparison and reporting.
+    ///
+    /// The wire form is `<name UTF-8 byte count>:<name><position UTF-8 byte count>:<position>`.
+    /// Because the two payload strings are length-prefixed, concatenating sorted entries cannot
+    /// be forged by putting a former joining character (such as `@`) in a marker name.
+    private static func canonicalMarkerIdentity(name: String, position: String) -> String {
+        "\(name.lengthOfBytes(using: .utf8)):\(name)\(position.lengthOfBytes(using: .utf8)):\(position)"
     }
 
     /// Selects the row and confirms the table agrees, since a write that reports success without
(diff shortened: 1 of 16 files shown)


MongLong0214:
Independent review found real gaps. Recording them here before revising.

**1 — I marked one refusal branch out of two.** `nav.delete_marker` refuses in two places, both
`ax_write_failed` with `write_attempted: false`:

```swift
guard selectMarkerRowForDeletion(index, in: window, runtime: runtime.ax) else { ... }   // unmarked
guard markerTableHasKeyboardFocus(runtime: runtime) else { ... }                        // marked
```

The first is the same defect this PR exists to close: it still falls through to the MIDI key-command
channel, which ignores `index` entirely and fires CC 45 blindly. Fixing one branch and leaving its
sibling is exactly the "fix the named instance instead of the property" mistake.

**2 — `isFallbackUnsafeStateC` does not check that the envelope is State C.** It tests
`success == false` and the marker. All six encoders merge `extras` *after* setting their canonical
fields, so a caller can overwrite `success`, and `.error(...)` does not imply State C either — there
are call sites that return `.error(encodeStateB(...))`. The predicate must require `state == "C"` and
a String `error` as well.

**3 — Japanese is unproved for the focus gate.** `markerContainerKeywords` carries English and Korean
only. If Japanese Logic localizes that description, a valid Marker List focus becomes a terminal
refusal. That fails closed rather than acting unsafely, but three-locale operation is not
demonstrated. I will measure the Japanese labels from a Japanese Logic rather than translate them,
and that is deliberately not folded into this PR.

**4 — an ordering consequence worth stating.** The new check runs before terminal classification, so
a *marked* terminal `element_not_found` would bypass `shouldContinueAfterTerminalStateC`. No existing
envelope carries the marker, so nothing currently changes — but it is a real ordering change and
belongs in the comment rather than being discovered later.

Also noted and not changed here: the marked refusal happens after `selectMarkerRowForDeletion` has
already written a selection, so `write_attempted: false` is describing the *delete*, not every AX
write in the call. That predates this PR; widening it here would be a second change hiding inside a
blocker fix.

Revising for 1, 2 and 4.

MongLong0214:
Revised. All three findings addressed, plus one the live evidence turned up on its own.

**1 — both refusal branches marked.** The selection-failure branch now carries `fallback_unsafe` too,
with a comment saying why a key-command channel cannot answer an indexed delete: it ignores `index`
and fires CC 45 without proving which marker it would remove.

**2 — `isFallbackUnsafeStateC` now requires the whole State C shape**: `success:false`,
`state:"C"`, and a String `error`, in addition to the marker. `success:false` alone was not proof,
exactly as the review said — the encoders merge `extras` after their canonical fields, so a caller
can overwrite it.

**3 — the ordering consequence is written at the check**, and corrected while writing it: the first
draft of that comment claimed no envelope carries the marker, which stopped being true the moment
this commit marked two of them. It now says what is actually load-bearing — both carry
`ax_write_failed`, which is not terminal, so the ordering changes nothing.

Both new behaviours are mutation-backed, not assumed:

- removing the `state == "C"` requirement fails
  `testIssue526StateAExtrasCannotMasqueradeAsFallbackUnsafeStateC`;
- removing the marker from the selection refusal fails
  `testIssue526SelectionFailureRefusalIsFallbackUnsafeAndStopsChain` on four assertions, including
  that the weaker channel ran.

7 tests; full suite 3480.

## The review's fifth point, confirmed by the screenshots

The review noted that the marked refusal happens after a selection has already been written, so it is
not a pure "nothing attempted" State C. The live run demonstrated it independently and I did not
plant it: my visual assertion claimed the marker rows would be **pixel-identical** across a refusal,
and it **failed**. Comparing the two captures, all sixteen rows are present in both — nothing was
deleted — but the row highlight moved.

So the assertion was the thing that was wrong, not the product. It now asserts on Logic's own marker
count readout, which is what "nothing was deleted" actually means, and the row-region change is
recorded as an observation instead of being asserted away. The count region was separately controlled
— adding one marker changes it — so "unchanged" is a result rather than an artefact.

The underlying behaviour is real and predates this PR: a State C whose `write_attempted:false`
describes the delete, while a selection write did happen. Filing it separately rather than widening
this change to cover it.

MongLong0214:
Revised again. The review's in-scope finding was correct and is fixed at the routing table.

## `fallback_unsafe` alone could not close #526

It only takes effect once the Accessibility channel has actually run and returned a marked envelope.
`ChannelRouter.route` skips a channel entirely — `continue`, producing **no envelope at all** — when
it is not registered, reports `available: false`, or is not runtime-ready. On any of those the chain
reached the MIDI key-command rung, which ignores `index` completely and fires CC 45 at whichever
marker happens to be selected.

So the operation this PR exists to protect could still remove the wrong marker and report success,
through a route the marker never touched.

`nav.delete_marker` is now `[.accessibility]`. The principle written at the change, because it is the
reason rather than the effect: **a channel that cannot express the operation's target must not be in
that operation's chain.** `delete_marker` takes an `index`; the key-command rung has no way to carry
one, so it cannot answer the request even in principle. (It also required a manual MIDI Learn before
it did anything at all.)

`fallback_unsafe` stays — it remains correct for the in-chain refusal and for other multi-channel
operations. This closes the other route to the same outcome.

## Both flagged assertions are now attributable

- the terminal-State-C case asserted zero key-command executions, which already held because
  `.elementNotFound` is terminal regardless of the new gate. It now uses a non-terminal code, and
  removing `fallback_unsafe` makes MIDI execute once and fail it.
- the selection-failure case was masked by a fixture with no focused element — had selection wrongly
  trusted AX success, the later focus guard would have emitted the same refusal and the assertion
  would still have passed. Making `AXSelectedRows` confirm the row now fails it.

`nav.delete_marker` leaves `expectedKeycmdOnlyOps` because its chain no longer contains the
key-command channel, which is the property that set is defined by — not a weakened check.

## Audited while here

`nav.rename_marker` is already accessibility-only. **`nav.goto_marker` is
`[.midiKeyCommands, .cgEvent]`** and takes an `index` or a `name` that neither channel can carry —
the same class of defect, reported rather than changed here.

## Still deliberately out of scope

The focus guard verifies the *kind* of table rather than the *identity* of the one the row was
selected in (#535), and the Japanese label gaps fail closed. Neither is introduced or worsened here.

Live evidence re-recorded against this head: three mutation-backed checks, the visual assertion on
Logic's own marker count, restoration verified, screen recording. Full suite 3480.

One honesty note about that run: the first attempt recorded a failed check because the count readback
came back **empty**, not because a marker was deleted. I read the count independently (`14 Markers`,
unchanged), confirmed nothing was lost, and re-ran. The evidence document is the clean run; this is
the record that it was not the only one.

MongLong0214:
Revised again — the contract surfaces now agree with the router.

Making the chain accessibility-only left four public declarations saying the opposite, which the
review caught:

| surface | was | now |
| --- | --- | --- |
| `OperationRegistry` | `.none` / `.requiresKeyBinding` | `.readbackRequired` / `.defaultInstall` |
| `SemanticOracleTable` | "send-only … State B only", structurally excluded | a real oracle; exclusion removed |
| `MIDIKeyCommandsChannel` health detail | listed as keycmd-only | removed |
| `docs/SETUP.md` | listed under manual MIDI Learn | removed |

That mattered beyond tidiness: release qualification would have kept deferring an operation that can
reach State A, and the public catalog would have told callers it needs a MIDI Learn it does not.

`.readbackRequired` describes what the AX path already does — it compares the complete expected
survivor set (name@position for every marker except the target) against a settled readback and
issues State A only on a match. The new oracle pins exactly that: operation identity, target name and
position, both counts, and `readback_settled`.

`nav.goto_marker` stays keycmd-only in all four places. It really is (#537).

All three weak assertions the review named now have a mutation that fails them, and I ran each:

- reintroducing `nav.delete_marker` into the health-detail text fails the exact-set-equality check
  (it could not before);
- allowing fallback to the probe fails the refusal-envelope assertion;
- collapsing the refusal to its message alone fails the not-a-success assertion.

Two census counts moved with the sets they count (B2 15 → 16, covered total 48 → 49), with the reason
written beside them.

## One thing worth recording about the suite

Run in parallel, this tree reports 58 issues. Run with `--no-parallel`, it reports 2 — the two census
counts, which are now fixed, and then 3480 pass. The other 56 are scheduling artefacts, not
behaviour; the ship gate uses `--no-parallel` for that reason. I checked this rather than assuming,
because "58 failures" and "2 failures" would have led to very different conclusions.

Live evidence re-recorded against this head. Full suite 3480.

MongLong0214:
Blocked again, and this one is on me twice over.

The oracle I added does not prove a deletion. It pins `readback_settled: true` (which proves the
reads were stable, not that anything changed), type-checks both counts without requiring `after` to
be `before - 1`, and pins `target_name` / `target_position` — which are **pre-write metadata echoed
from the request**. A State A payload with equal counts, or with the wrong survivor removed, passes
it.

I noticed this myself while reviewing the change — I wrote that it "doesn't assert
`marker_count_after == marker_count_before - 1`" — and then talked myself out of acting on it because
the comment said State A is only reached when the survivor set matches. That reasoning was wrong in a
specific way worth naming: **the guarantee lives in the handler, and an oracle exists precisely to
not take the handler's word for it.** If the oracle only restates the handler's shape, it verifies
nothing.

The fix is to expose what the handler already computes. The State A branch builds the complete
expected survivor set (`name@position` for every marker except the target) and compares it with the
observed one; that comparison is the proof and it never reaches the envelope. Both sets go into the
State A extras and the oracle asserts they are equal — the same shape as `navigateRenameMarker`'s
`.fieldsEqual(requested_name, observed_name)`, which is a real request-versus-readback bound rather
than a shape check.

With the negative cases the review asked for: a settled State A whose counts do not decrease is
rejected, and one whose observed survivors differ from the expected set is rejected.

One trap being handled explicitly: this repository has already recorded that a naive `key=value`
join lets a marker *name* forge an entry. The exposed form must not be spoofable by a name.

The index-blind fallback finding from the previous round is confirmed addressed.

MongLong0214:
**Live measurement invalidates part of this PR's claim. Recording it before revising.**

Chasing the review's oracle finding into a real delete turned up something neither of us had:
**Logic renumbers its auto-generated marker names when one is deleted.**

Measured, one delete of index 0, with the survivor sets now exposed in the envelope:

```
expected  Marker 2 @1.1.1.1 … Marker 12 @1.1.1.1, Marker 13 @5.1.1.1, Marker 14 @5.1.1.1, Marker 15 @5.1.1.1
observed  Marker 1 @1.1.1.1 … Marker 11 @1.1.1.1, Marker 12 @5.1.1.1, Marker 13 @5.1.1.1, …
```

`marker_count_before: 15`, `marker_count_after: 14`, and Logic's own readout went `15 Markers` →
`14 Markers`. **The delete worked.** But every default-named survivor shifted its name by one, so the
`name@position` comparison can never match.

## What that means for this PR

`delete_marker` therefore **cannot reach State A in a project with default marker names** — which is
the common case. It answers State B `readback_mismatch`, honestly, for a deletion that happened.

So the registry change to `.readbackRequired` / `.defaultInstall` and the new oracle describe a State
A that most real projects will never see. That is not wrong in principle — the operation does perform
a readback, and with user-named markers the comparison does match — but shipping it without saying so
would leave a catalog entry that is true only in the uncommon case.

## The fix, which live testing also handed me

The identity used for the comparison is the thing that breaks: an auto-generated name is not stable
across the very operation being verified. Positions are stable. So the provable claim is:

> the multiset of survivor **positions** lost exactly one instance of the target's position, and the
> count decreased by exactly one

That survives renumbering, does not depend on names at all, and is a real request-versus-readback
bound rather than a shape check. It is weaker than name+position identity in one specific way —
with two markers at the same position it cannot say *which* of them went — and that limitation
belongs in the comment rather than being papered over.

Revising to that. The length-prefixed encoding stays (a name can still forge an entry in the
name-carrying diagnostic fields).

This is the third time today that driving the real application contradicted something the code and I
both believed. It is also why the oracle finding was worth chasing rather than patching around.

MongLong0214:
Fixed, and the fix is confirmed against the running application.

State A now compares the **multiset of survivor positions** — the pre-write positions with exactly
one instance of the target's position removed — instead of name+position identities that Logic
invalidates by renumbering.

Measured on the same project, before and after the change:

| | result |
| --- | --- |
| name+position comparison | `state B`, `readback_mismatch`, 15 → 14, Logic's readout `15 Markers` → `14 Markers` |
| position multiset | **`state A`, `verified: true`**, 14 → 13, Logic's readout `14 Markers` → `13 Markers` |

So the operation could not reach State A in any project with default marker names, and now it does.

The limitation is written at the comparison rather than hidden: **with two markers at the same
position this cannot establish which of them was removed.** Names are kept for the State B mismatch
diagnostic, where they help a human read a failure, with the length-prefixed encoding so a name
cannot forge an entry there.

The oracle binds the same equality, and the negative cases hold: a settled State A whose positions
did not lose an instance of the target position is rejected; one whose observed positions differ from
the expectation is rejected; and — the case that used to fail in the real world — one where every
survivor NAME shifted but the positions are correct is accepted.

Full suite 3481 serially. Live evidence re-recorded against this head.

MongLong0214:
Blocked, and the first finding is the sharp one: **I wrote the limitation in a comment and then
claimed State A anyway.**

The position multiset proves that *one* marker at the target's position went. It cannot prove *which*
one when several share that position — which is common, and is exactly the shape of the project I
measured on. So a delete of the wrong requested index can be certified as verified. Documenting a
limitation does not exempt the claim from it.

Fixing the claim rather than the comment: **State A only when the target's position is unique among
the pre-write markers** — then losing one instance of it identifies the target. When the position is
shared, the honest answer is State B: the delete was attempted, the count moved, and which marker
went cannot be established from this readback. A field will record which case applied so a reader can
tell a verified delete from an unverifiable one.

Second finding, also correct: the oracle binds two **self-reported** fields to each other, so a
handler writing the same arbitrary value twice passes. The negative tests mutate only one side and
miss it. The counts and the multisets will have to be consistent with each other, with a negative
case where both fields carry the same junk and the counts disagree.

Third: `8:5.1.1.1` in a test fixture is a malformed length prefix — `5.1.1.1` is **seven** UTF-8
bytes. A wrong prefix makes that fixture prove nothing about the real encoding. Checking every
hand-written length-prefixed literal for the same error.

MongLong0214:
Fixed, and both halves are verified against the running application.

**State A now requires the target's position to be unique** among the pre-write markers. When it is
shared, the answer is State B — the delete was attempted, the count moved, and which marker went
cannot be established from this readback. `target_position_unique` records which case applied.

Measured live:

```
shared position   index 0, Marker 1 @1.1.1.1   state=B  verified=false  unique=false  19 -> 18
```

The reviewer also asked whether an ordinal fallback position after a parse failure could make two
markers look co-located. It can, and that guard is now in and **fired on the first real project I
pointed it at**:

```
unique position, multisets identical, count 18 -> 17
  state=B  reason_detail: "at least one marker position came from an ordinal parse fallback;
                           synthetic positions cannot support a verified target identity"
```

So on that project the position column is not parsing and the product substitutes ordinals — and it
correctly refuses to certify identity on synthetic data. That is the conservative answer and I am
recording it rather than filing it: **I have no live State A for this operation, only the two live
refusals plus the unit and oracle coverage.** Saying otherwise would be the exact over-claim this PR
has spent several rounds removing.

The oracle no longer binds two self-reported fields to each other — the multiset entry counts must
agree with the reported counts, with a negative case carrying identical junk in both fields and
disagreeing counts.

One fixture used `8:5.1.1.1` as a length-prefixed entry; `5.1.1.1` is **seven** UTF-8 bytes. I
re-parsed every hand-written length-prefixed literal in the tests with the real format rather than
eyeballing them — that was the only genuine error, and my first grep-based audit reported 87 of 92
"wrong" because the regex was wrong, not the fixtures.

Also fixed: seven `#expect(x as? Bool == true/false)` assertions the public-surface preflight
flags as always-pass. Bound with `try #require` and asserted bare or negated.

Full suite 3488. Live evidence re-recorded against this head.

MongLong0214:
Follow-up on why the live State A did not appear, recorded here rather than filed.

The ordinal fallback that tripped the new guard is `StateModels.swift`:

```swift
position: parsed ?? "\(ordinal + 1).1.1.1",
positionSource: parsed != nil ? .parser : .fallback
```

So when the Marker List's Position cell does not parse, the marker is given **`<ordinal+1>.1.1.1`** —
a synthetic value shaped exactly like a real bar position. That is why the marker I created at bar 77
was reported at `19.1.1.1`: it was the nineteenth row.

Two consequences worth having written down:

1. A synthetic position is indistinguishable from a real one **by value**. Only `positionSource`
   separates them, which is precisely why this PR's guard reads that field rather than the position
   itself. Anything else that compares marker positions has the same exposure.
2. Because the fallback is derived from the row ordinal, two rows can never collide *within one
   read* — but a fallback position can collide with a **real** position from another row, which is
   the case the guard exists for.

I did not chase why the cell fails to parse: reaching that table needs the product's own descendant
search (System Events does not vend it at the depth I could walk), and the guard makes the current
behaviour correct regardless. Noting the mechanism so whoever does chase it starts from the right
place instead of from the symptom.

Issue #519, opened 2026-08-11
Mixer volume/pan is unusable on any Logic UI language other than English or Korean


MongLong0214:
### Problem

Ten AppleScript menu drives hard-code the **English** menu-bar item name, so every operation that
routes through them fails on a Logic running in any other language:

| menu bar item | occurrences | files |
| --- | ---: | --- |
| `"Navigate"` | 5 | `AccessibilityChannel+Transport.swift`, `+Regions.swift` |
| `"File"` | 4 | `AccessibilityChannel+Project.swift`, `+MIDIImport.swift` |
| `"Edit"` | 1 | `AccessibilityChannel+Markers.swift` |

### Measured, not inferred

Logic switched to Korean (`defaults write com.apple.logic10 AppleLanguages -array ko`) on this
machine, then `project.new` driven through the release artifact:

```
2.8 s → success: false   state: C   error: channels_exhausted
        hint: "Could not reveal Creator Studio chooser through exact File > New"
```

It fails **before** reaching either creation branch. The script asks for
`menu item "New" of menu 1 of menu bar item "File"`; on a Korean Logic those are `파일` and
`새로 만들기`, so it returns `EXACT_NEW_NOT_FOUND` and the operation gives up. The same machine in
English completes the operation in 5.7 s.

Two more English-only assumptions were measured in the same session and are the same defect class:

- The arrange window's title suffix is localized — `Untitled 55 - Tracks` in English,
  `Untitled 55 - 트랙` (U+D2B8 U+B799) in Korean. `isCreatedProjectWindowTitle` matched only
  `" - Tracks"`, so even after the menu is reached, the created-project witness would not fire.
- The Event List column header is localized — `["L","M","Position","Status","Ch","Num","Val","Length/Info"]`
  in English, `["L","M","위치","상태","채널","번호","값","길이/정보"]` in Korean. The MIDI readback
  collector compares against English literals, so it throws `headerMismatch` on a Korean Logic.

### Why it matters beyond one operation

The affected surfaces are `project.new` and bounce, `midi.import_file`, marker operations, region
operations, and transport/navigation. On a non-English Logic these do not degrade — they fail closed
with an error that names an English menu the user does not have. An outside integrator ended their
Logic Pro integration over the `project.new` half of this (#516), on a machine configured differently
from the maintainer's.

### Proposed fix

Resolve the menu-bar item by **enumerating the live menu bar** and matching each candidate against an
[...]

PR #531, opened 2026-08-11
fix(#529): validate the Go To menu before reading whether it is enabled

### LINKED ISSUE #529: goto_bar falls back to the slider route (one-bar nudge) because its dialog-route guard reads a stale cached menu-validation value
`goto_bar` silently loses its exact route and degrades to a one-bar nudge, because the guard that
selects the route reads a **cached** menu-validation value that can be stale.

## What I measured

`AccessibilityChannel+Transport.gotoPositionViaDialog` gates the dialog route on:

```applescript
set mi to menu item "Position…" of menu 1 of menu item "Go To" of menu 1 of menu bar item "Navigate" of menu bar 1
if not (enabled of mi) then
    return "MENU_DISABLED"
end if
```

`MENU_DISABLED` becomes an error, the router falls through, and the slider route runs instead.

**The same read, on the same menu item, with the menu closed, returned different values minutes
apart with no project change in between:**

```
15:04  closed-menu read → enabled=false      (every item in the submenu read false)
15:05  menu opened by hand → enabled=true    (Position…, Left Locator, Right Locator all true)
15:09  closed-menu read → enabled=true
```

The item was actionable the whole time: driving `Navigate ▸ Go To ▸ Position…` by hand at 15:05
opened the `Go To Position` dialog immediately.

## What the product did in each state

While the cached value was false — release build, real Logic, requesting bar 25 and 37:

```
{"state":"B","verified":false,"requested":"25.1.1.1","observed":"10.1.1.1","via":"slider"}
{"state":"B","verified":false,"requested":"25.1.1.1","observed":"11.1.1.1","via":"slider"}
{"state":"B","verified":false,"requested":"25.1.1.1","observed":"12.1.1.1","via":"slider"}
```

**The playhead advances exactly one bar per call** rather than going where it was asked. Each call
reports State B with `isError: true` and the true `observed` value, so the envelope is honest — but
the playhead has moved, to a position nobody requested, and repeated calls walk it forward.

After the menu was opened once, the same build, same project, same request:

```
{"state":"A","verified":true,"requested":"25.1.1.1","observed":"25.1.1.1","via":"dialog"}
{"state":"A","verified":true,"requested":"37.1.1.1","observed":"37.1.1.1","via":"dialog"}
```

## Why this matters beyond the route choice

`FrontmostGate`'s own doc comment records the symptom this produces:

> `goto_position` driven from the background lands on the wrong bar (requested 1.1.1.1, observed
> 3.3.1.1 and 2.3.1.1 on two runs, against 2/2 exact hits with Logic frontmost)

That is the same signature — an exact request landing a short distance away — and the slider fallback
produces it deterministically. So at least some of what has been attributed to focus may be this
guard instead. Worth re-examining rather than assuming.

## What I have NOT established

**What makes the cached value go stale-false.** I can reproduce both states and I can restore the
working one by opening the menu, but I did not isolate the trigger. Do not write a fix that assumes a
cause I have not shown.

## Direction, not a prescription

macOS validates menu items when a menu opens; reading `enabled` on a closed menu returns whatever the
last validation left behind. So the guard needs the menu opened (and therefore validated) before it
reads, rather than reading a value that has no defined freshness. This repository has already
recorded that the guard's strictness is deliberate — *"an unreadable `AXEnabled` must not authorise a
pick"* — so the fix is to make the read meaningful, **not** to drop the guard.

There is a second, smaller question in the same code: when the dialog route is skipped, the slider
fallback mutates the playhead. A route that cannot deliver the requested position arguably should not
move it at all. That is a contract question, not a bug report, and I am not deciding it here.

## Environment

Logic Pro 12.3 (English), release build of the current `main`, driven over the raw MCP protocol.
`nav.goto_bar` throughout; no coordinate or mouse events involved in any measurement above.

MongLong0214:
Closes #529.

## The defect

`gotoPositionViaDialog` gated the dialog route on `enabled of mi`, read on a **closed** menu. That
value is a cache, and a reading taken while the item is disabled sticks — outliving both the
condition and the process that took it.

Isolated cleanly on Logic Pro 12.3, each reading a separate `osascript` invocation:

```
reset (open the Navigate menu, Escape)                       → enabled=true
open the modal, cancel it, without reading while it is open   → enabled=true    (not poisoned)
reset                                                         → enabled=true
open the modal, read once while it is open, cancel it         → enabled=false   (persists)
```

**So the route poisons itself.** If a previous call leaves the `Go To Position` modal open — which
happens; I found one wedged after an interrupted run — the next call's guard reads `enabled` while
that modal is up, gets `false`, and answers `MENU_DISABLED`. The router falls through to the slider
route, which nudges the playhead **one bar per call** instead of going where it was asked. And that
read has now cached `false` for every later reader, so it keeps happening after the modal is gone.

That last part is why it looked intermittent and unconnected to the modal.

A fresh Logic launch has never opened that menu either, so the same guard reads `false` there too.

## The change

Open the menu chain before reading. That re-validates the item, which both makes the read meaningful
and replaces the poisoned value.

The guard itself stays — this repository's recorded decision is that an unreadable `AXEnabled` must
not authorise a pick — so an unreadable state becomes its own refusal (`MENU_STATE_UNREADABLE`)
rather than being conflated with "disabled". Every refusal path now dismisses the menu it opened, so
a refusal cannot leave the application wedged behind an open menu.

The slider fallback is untouched. Whether a route that cannot deliver the requested position should
refuse to move the playhead at all is a separate contract question and is not decided here.

## Live evidence

Two release artifacts, same machine, the poisoned precondition re-established immediately before each
run:

| | requested | observed | state | via |
| --- | --- | --- | --- | --- |
| parent commit | `53.1.1.1` | `46.1.1.1` | B | slider |
| this branch | `61.1.1.1` | `61.1.1.1` | **A**, verified | **dialog** |

The evidence document bound to this head records the precondition it created
(`read_after_modal_dismissed: "false"` — genuinely poisoned, not assumed), three mutation-backed
checks, three settled captures within one display, a visual assertion that the transport's BAR
readout changed, the playhead restored to where the run found it, and a screen recording.

The visual region was verified by cropping it and looking at it, so that it holds the BAR/BEAT digits
and excludes the LCD's MIDI-activity dot — which blinks on its own and stops any capture settling.

## Tests

`Issue529MenuValidationTests` asserts on the generated AppleScript by **string ordering**: the menu
bar item and the submenu parent must be opened before the first `enabled of` read, in both the Korean
and the English chain, and the disabled path must dismiss the menu before returning.

Mutation-tested rather than assumed: removing the four menu-opening clicks fails two of the three
tests. The third guards the dismissal, which that mutation does not touch.

Full suite: 3476 tests pass.

### DIFF
--- a/Sources/LogicProMCP/Channels/AccessibilityChannel+Transport.swift
+++ b/Sources/LogicProMCP/Channels/AccessibilityChannel+Transport.swift
@@ -971,43 +971,100 @@ extension AccessibilityChannel {
         """
     }
 
+    /// The successful AppleScript channel payload is a JSON object containing the
+    /// script's return value in `result`. Only an exact `OK` means this route
+    /// reached the dialog and submitted it; any sentinel or malformed payload
+    /// must fall through to the slider route.
+    enum GotoPositionDialogResultClassification: Equatable {
+        enum Failure: Equatable {
+            case menuNotFound
+            case menuStateUnreadable
+            case menuDisabled
+            case menuPickFailed
+            case dialogNotReady
+            case malformedPayload
+            case unexpectedResult
+        }
+
+        case driven
+        case failure(Failure)
+    }
+
+    private struct GotoPositionDialogScriptPayload: Decodable {
+        let result: String
+    }
+
+    /// Kept internal for the menu-validation regression tests. This consumes
+    /// the JSON envelope produced by `AppleScriptChannel.channelResult` rather
+    /// than inspecting its serialized representation.
+    static func classifyGotoPositionDialogResult(
+        _ output: String
+    ) -> GotoPositionDialogResultClassification {
+        guard let data = output.data(using: .utf8),
+              let payload = try? JSONDecoder().decode(GotoPositionDialogScriptPayload.self, from: data)
+        else {
+            return .failure(.malformedPayload)
+        }
+
+        switch payload.result {
+        case "OK":
+            return .driven
+        case let value where value.hasPrefix("MENU_NOT_FOUND"):
+            return .failure(.menuNotFound)
+        case "MENU_STATE_UNREADABLE":
+            return .failure(.menuStateUnreadable)
+        case "MENU_DISABLED":
+            return .failure(.menuDisabled)
+        case let value where value.hasPrefix("MENU_PICK_FAILED"):
+            return .failure(.menuPickFailed)
+        case "DIALOG_NOT_READY":
+            return .failure(.dialogNotReady)
+        default:
+            return .failure(.unexpectedResult)
+        }
+    }
+
     private static func gotoPositionViaDialog(bar: Int) async -> ChannelResult {
         let script = gotoPositionViaDialogAppleScript(bar: bar)
-        let result = await AppleScriptChannel.executeAppleScript(script)
+        // Fixed waits consume 0.8 s and the dialog poll can consume 3.0 s.
+        // Eight seconds leaves 4.2 s for osascript startup and menu traversal.
+        let result = await AppleScriptChannel.executeAppleScript(script, timeout: 8.0)
         switch result {
         case .success(let output):
-            if output.contains("MENU_DISABLED") {
-                return .error("goto-position dialog disabled (project has no regions yet)")
-            }
-            if output.contains("MENU_STATE_UNREADABLE") {
+            switch classifyGotoPositionDialogResult(output) {
+            case .driven:
+                // #105: this State B reports only that the Go-To-Position dialog
+                // keystroke was sent; the playhead is verified independently by
+                // `TransportDispatcher.finalizeGotoPositionResult` via a transport-
+                // state read-back. Earlier this carried a `note` claiming the
+                // playhead was "not read back" — which the finalize step then
+                // contradicted by reading it back and gating `verified` on it, so a
+                // verified State A shipped a self-contradictory note. The provenance
+                // (`via:"dialog"`) plus finalize's `verification_source` /
+                // `observed` / `verified` fields describe the outcome honestly
+                // without it.
+                return .success(HonestContract.encodeStateB(
+                    reason: .readbackUnavailable,
+                    extras: [
+                        "requested": "\(bar).1.1.1",
+                        "via": "dialog"
+                    ]
+                ))
+            case .failure(.menuDisabled):
+                return .error("goto-position dialog menu state did not authorize the pick")
+            case .failure(.menuStateUnreadable):
                 return .error("goto-position dialog menu enabled state unavailable")
-            }
-            if output.hasPrefix("MENU_NOT_FOUND") {
-                return .error("goto-position menu not found: \(output)")
-            }
-            if output.hasPrefix("MENU_PICK_FAILED") {
-                return .error("goto-position dialog menu pick failed: \(output)")
-            }
-            if output.contains("DIALOG_NOT_READY") {
+            case .failure(.menuNotFound):
+                return .error("goto-position menu not found")
+            case .failure(.menuPickFailed):
+                return .error("goto-position dialog menu pick failed")
+            case .failure(.dialogNotReady):
                 return .error("goto-position dialog did not appear within timeout")
+            case .failure(.malformedPayload):
+                return .error("goto-position dialog returned an invalid result payload")
+            case .failure(.unexpectedResult):
+                return .error("goto-position dialog returned an unexpected result")
             }
-            // #105: this State B reports only that the Go-To-Position dialog
-            // keystroke was sent; the playhead is verified independently by
-            // `TransportDispatcher.finalizeGotoPositionResult` via a transport-
-            // state read-back. Earlier this carried a `note` claiming the
-            // playhead was "not read back" — which the finalize step then
-            // contradicted by reading it back and gating `verified` on it, so a
-            // verified State A shipped a self-contradictory note. The provenance
-            // (`via:"dialog"`) plus finalize's `verification_source` /
-            // `observed` / `verified` fields describe the outcome honestly
-            // without it.
-            return .success(HonestContract.encodeStateB(
-                reason: .readbackUnavailable,
-                extras: [
-                    "requested": "\(bar).1.1.1",
-                    "via": "dialog"
-                ]
-            ))
         case .error(let msg):
             return .error("goto-position dialog failed: \(msg)")
         }
--- a/Sources/LogicProMCP/Channels/AccessibilityChannel+Transport.swift
+++ b/Sources/LogicProMCP/Channels/AccessibilityChannel+Transport.swift
@@ -889,29 +889,54 @@ extension AccessibilityChannel {
         return text
     }
 
-    private static func gotoPositionViaDialog(bar: Int) async -> ChannelResult {
+    /// The script is internal so the menu-validation regression tests can assert the
+    /// exact generated AppleScript ordering without invoking Logic Pro.
+    static func gotoPositionViaDialogAppleScript(bar: Int) -> String {
         // Poll for the dialog's presence instead of relying on a fixed delay.
         // Without this guard, a slow machine (>500ms to render the dialog) would
         // send Cmd+A to the arrange area, selecting all regions unexpectedly.
         let logicProAppleScript = LogicProTarget.appleScriptTarget()
-        let script = """
+        return """
         \(logicProAppleScript.activateByBundleID)
         delay 0.2
         tell application "System Events"
             tell \(logicProAppleScript.systemEventsProcessTarget)
                 try
+                    -- Opening each menu makes Logic validate the child item before
+                    -- AXEnabled is read. A closed menu can expose a stale value.
+                    click menu bar item "탐색" of menu bar 1
+                    click menu item "이동" of menu 1 of menu bar item "탐색" of menu bar 1
                     set mi to menu item "위치…" of menu 1 of menu item "이동" of menu 1 of menu bar item "탐색" of menu bar 1
                 on error errMsg
                     try
+                        -- The Korean lookup may have opened a menu before failing.
+                        -- Dismiss it before opening the English fallback chain.
+                        key code 53
+                        click menu bar item "Navigate" of menu bar 1
+                        click menu item "Go To" of menu 1 of menu bar item "Navigate" of menu bar 1
                         set mi to menu item "Position…" of menu 1 of menu item "Go To" of menu 1 of menu bar item "Navigate" of menu bar 1
                     on error errMsg2
+                        key code 53
                         return "MENU_NOT_FOUND: " & errMsg2
                     end try
                 end try
-                if not (enabled of mi) then
+                try
+                    set menuItemEnabled to enabled of mi
+                on error
+                    -- An unreadable AXEnabled must not authorise the pick.
+                    key code 53
+                    return "MENU_STATE_UNREADABLE"
+                end try
+                if not menuItemEnabled then
+                    key code 53
                     return "MENU_DISABLED"
                 end if
-                click mi
+                try
+                    click mi
+                on error errMsg
+                    key code 53
+                    return "MENU_PICK_FAILED: " & errMsg
+                end try
                 -- Wait up to 3s for the dialog window to appear before typing,
                 -- otherwise keystrokes would go to the arrange area and click
                 -- Cmd+A there — silently "Select All Regions".
@@ -930,6 +955,7 @@ extension AccessibilityChannel {
                     end try
                 end repeat
                 if not dialogReady then
+                    key code 53
                     return "DIALOG_NOT_READY"
                 end if
             end tell
@@ -943,14 +969,24 @@ extension AccessibilityChannel {
         end tell
         return "OK"
         """
+    }
+
+    private static func gotoPositionViaDialog(bar: Int) async -> ChannelResult {
+        let script = gotoPositionViaDialogAppleScript(bar: bar)
         let result = await AppleScriptChannel.executeAppleScript(script)
         switch result {
         case .success(let output):
             if output.contains("MENU_DISABLED") {
                 return .error("goto-position dialog disabled (project has no regions yet)")
             }
+            if output.contains("MENU_STATE_UNREADABLE") {
+                return .error("goto-position dialog menu enabled state unavailable")
+            }
             if output.hasPrefix("MENU_NOT_FOUND") {
                 return .error("goto-position menu not found: \(output)")
+            }
+            if output.hasPrefix("MENU_PICK_FAILED") {
+                return .error("goto-position dialog menu pick failed: \(output)")
             }
             if output.contains("DIALOG_NOT_READY") {
                 return .error("goto-position dialog did not appear within timeout")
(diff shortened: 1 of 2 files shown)


MongLong0214:
Independent review blocked this, and the sharpest finding is one my own change made worse. Recording
it before revising.

## The sentinel checks cannot match

`AppleScriptChannel.channelResult` wraps every successful osascript run:

```swift
return .success("{\"result\":\"\(AppleScriptChannel.escapeJSON(result))\"}")
```

so the output always starts with `{"result":"`, and neither of these ever fires:

```swift
if output.hasPrefix("MENU_NOT_FOUND")   { ... }   // pre-existing
if output.hasPrefix("MENU_PICK_FAILED") { ... }   // added on this branch
```

`MENU_DISABLED` and `DIALOG_NOT_READY` use `contains`, which is the only reason they work.

**What that costs:** when the menu is not found or the pick fails, control falls to the final
`return .success(encodeStateB(... "via": "dialog"))` — claiming the dialog route was driven when it
was not. The finalize step then reads the transport, and if the playhead already happens to sit at
the requested bar, it emits **State A, verified**, for an operation that did nothing.

One of the two was pre-existing; the other I added, by copying the shape next to it without checking
what the wrapper does. Being fixed by parsing the payload and matching on its `result` value, for
every sentinel, so there is one way of doing this rather than two.

## The script no longer fits its timeout

`ServerConfig.appleScriptTimeout` is 5.0s. Measured against this branch's script: fixed delays 0.8s,
dialog-wait loop 30 × 0.1s = 3.0s, so 3.8s worst case — leaving about 1.2s for osascript startup
**and** the four menu clicks this branch added. A timeout kills the script mid-flight with menus or a
modal open, which is the state this branch exists to stop producing.

Being given an explicit timeout for this call, with the budget written down. Not raising the global
config value.

## A refusal that names a cause it did not observe

`"goto-position dialog disabled (project has no regions yet)"` asserts a cause that is now known to
be wrong at least some of the time — the value can be a stale `false` cached from a read taken while
a modal was up, which is this issue's entire subject. It will say only that the menu state did not
authorise the pick.

## Raised, and deliberately not changed here

- **Escape is unverified** on every early return, and the typing sequence after the dialog is
  detected is unguarded. This branch improved cleanup from "none" to "one Escape"; making cleanup
  *observed* is a larger change and belongs on its own.
- **Japanese is absent** from the chain. I will measure the labels from a Japanese Logic rather than
  translate them; folding a guessed locale string in here would be worse than leaving the gap
  visible.

MongLong0214:
Revised. All three findings addressed.

**1 — the sentinels are decoded, not prefix-matched.** The payload is parsed and its `result` value
classified, with exactly one value — `OK` — meaning the dialog was driven. A payload that does not
decode, or carries something unrecognised, is a **failure**, not a success. That closes the path
where a missing menu produced a State B claiming `via: "dialog"` and, if the playhead already sat at
the requested bar, a verified State A for an operation that did nothing.

Mutation-backed: breaking the classification so an unrecognised result counts as driven fails
`JSON-wrapped menu-not-found result refuses the dialog route`.

**2 — the timeout fits the budget.** 8 s for this call, with the arithmetic written at the call site
(0.8 s fixed waits + 3.0 s dialog poll leaves 4.2 s for startup and menu traversal). The global
`ServerConfig.appleScriptTimeout` is untouched.

**3 — the disabled refusal no longer names a cause.** It says the menu state did not authorise the
pick, which is what was observed.

Live evidence re-recorded against this head, with the poisoned precondition established and recorded
rather than assumed (`read_after_modal_dismissed: "false"`): `via: dialog`, State A verified,
requested `37.1.1.1` observed `37.1.1.1`, no dialog left behind, playhead restored, screen recording.
Full suite 3481.

## Japanese, now measured rather than deferred

I switched a real Logic to Japanese and read the labels off it. The dialog this route opens is
**`位置の移動`** — the script currently recognises only `위치로 이동` and `Go to Position`, so on
Japanese Logic it will hit `DIALOG_NOT_READY` and refuse to the slider. That is fail-closed and, after
this change, correctly classified rather than reported as a driven dialog.

The full chain measured there: `移動 ▸ 移動 ▸ 位置…` — the menu bar item and the submenu are the
**same word**. Deliberately still not added here: adding a third locale to this script is a separate
change with its own live verification, and folding it into a blocker fix would hide it. Measurements
are posted on #528 and #519 so whoever does it does not have to guess.

MongLong0214:
Blocked again, and both findings are mine — not #534's pre-existing ones.

**1. I turned a lookup into an actuation.** The locale fallback used to resolve elements and only
click once it knew which chain existed. My version clicks the Korean menu bar item first and lets the
AX error select the English path. Since an AX call can report failure after the click took effect,
that is a return-code-driven second actuator — the exact rule this repository has for not trusting AX
results. Fixing it by deciding the locale from a **read** and clicking only the chain that resolved.

**2. The menu opening I added has no observed cleanup.** The error branches assume one Escape closed
the menu, and the 8-second process timeout can bypass those handlers entirely — leaving a menu open
while the router proceeds to the slider. The previous version never opened a menu, so this exposure
is introduced here, not inherited.

**3. `Issue529MenuValidationTests.swift:76` cannot fail.** `dismissal` is found by searching from
`disabledGuard` onward, so once the `#require` passes the comparison is necessarily true. Line 77 is
the assertion doing the work.

Fixing all three here rather than filing them. Two of the three exist because I added menu clicks to
a function that previously only read.

MongLong0214:
Revised, and the live run caught a regression **my own fix** introduced — worth writing down because
a single run could not have found it.

The entry-time cleanup I added refused on any state that was not `CLOSED`, which included
`UNREADABLE`. Measured against the running application:

| | reached the dialog route |
| --- | --- |
| with the strict entry refusal | **3 of 8** fresh server processes |
| after the fix below | **8 of 8**, State A, `observed == requested` every time |

The first menu-bar read on a cold System Events connection is frequently unreadable, and a client
gets exactly **one** first navigation per process. So in practice five runs in eight would have
silently taken the slider route — the very degradation this PR exists to stop. Every individual call
looked fine; the failure only existed as a rate.

The fix follows from what the observation means: nothing has been opened at entry, so an unreadable
menu bar is an **absent observation**, not evidence that a menu is up. Entry now refuses only on an
observed `OPEN` it could not close. Every refusal *after* this run opens a menu keeps the strict
form, because there the run knows something is open. A test asserts both halves of that asymmetry.

The two findings from the previous review are also fixed:

- **the locale is decided by reads that actuate nothing** (`exists` tests), and exactly one chain is
  then clicked. Previously the Korean click ran first and an AX error selected the English click —
  and an AX call can report failure after the click landed, so a false negative clicked both.
- **every cleanup is observed**, polling `AXSelected` on the menu bar with a bound, and a failed read
  returns `UNREADABLE` rather than being counted as clean.

The 8-second process timeout can still kill the script mid-flight and no in-script handler can
survive that — which is why entry-time clearing exists at all: a predecessor killed by the timeout is
exactly the state the next run has to survive.

The assertion that could not fail is gone (it compared an index against one found by searching from
that same index onward). The three ordering tests were replaced rather than amended, because what
they asserted — a menu bar item clicked before `AXEnabled` is read — is no longer the design.

Live evidence re-recorded against this head, twice, clean both times. Full suite 3482.

MongLong0214:
Fixed. The refusal no longer feeds the actuator it was protecting against.

The dialog result is now classified into failures the slider may legitimately answer — menu not
found, item disabled, state unreadable, dialog never appeared, malformed payload — and one it may
not: **the menu could not be closed**. That case returns State C naming what happened, with
`write_attempted: false`, and never reaches the slider.

Both flagged assertions are rewritten. They compared an index against one found by searching from
that same index, so each held as soon as its helper returned. The second was worse than tautological:
it looked for a cleanup *somewhere before* each refusal, so one branch's cleanup satisfied every later
path — the test would have passed with a refusal path that had no cleanup at all. It now requires a
cleanup **between consecutive refusal returns**.

Mutation-backed: making `isUnsafeToActuateAgain` always false fails
`"a menu that could not be closed returns State C without touching the slider"`.

Measured after the change, eight fresh server processes: **8 of 8 reached the dialog route with
State A**, so the added gate costs nothing on the healthy path — which is the thing worth checking,
since the previous round's regression was exactly a safety guard degrading the healthy path.

Live evidence re-recorded against this head. Full suite 3484.

MongLong0214:
Blocked. Two findings accepted, one I am pushing back on with reasoning.

**Accepted — the entry guard has a hole.** `dismissOpenMenu` can observe `OPEN`, send Escape, then
read `UNREADABLE`, and the guard refuses only on `OPEN` — so that path actuates without ever
observing the menu closed. The distinction that makes this right is about what is known when:

- `UNREADABLE` **before any dismissal attempt** is an absent observation; nothing has been opened by
  this run, so it proceeds. (That is the fix that took fresh-process dialog success from 3/8 to 8/8.)
- `UNREADABLE` **after this run observed OPEN and tried to dismiss** means something WAS open and we
  cannot prove it closed. That must refuse.

Two outcomes, not one.

**Accepted — a State C that claims nothing was attempted, after a click.** Cleanup failures raised
after `click mi` report `write_attempted: false`. A click was attempted. That is the contract this
branch exists to defend, broken by the branch itself.

**Pushing back — the Japanese slider path.** The review is right that corrected `MENU_NOT_FOUND`
parsing makes the slider reachable on Japanese Logic, but that is an improvement rather than a newly
introduced defect. Before this branch, the same Japanese run produced a State B claiming
`via: "dialog"` for a dialog that never opened — and the finalize step could then emit **State A,
verified**, for an operation that did nothing. A slider that honestly reports what it did is strictly
better than a fabricated verified result. Leaving it, and not adding Japanese strings here: that is
#519, with the labels already measured.

MongLong0214:
Three more, all mine, all real.

**1. State C with `write_attempted: true` is a contradiction.** I asked for the refusal to stop lying
about what had been done, and it kept State C while flipping the flag. Attempted-but-unverifiable is
**State B**. Fixing the state, not the flag.

**2. `.axWriteFailed` is non-terminal, so the router keeps going.** When the menu is known to be
unclosed, `ChannelRouter` walks on to MCU / CoreMIDI / CGEvent — which is *exactly* the defect #526
exists for, in a place I did not look. The mechanism to stop it is the one #530 adds
(`fallback_unsafe: true` on a State C the router must return verbatim). So this branch will take
#530 first and mark that refusal, rather than inventing a second way to say the same thing.

**3. `click selectedMenuBarItem` then `click selectedSubmenuItem` with nothing observed between
them.** The first click's return selects the second actuator. Same rule as everywhere else in this
branch: observe that the menu opened before clicking into it.

**Two assertions that cannot fail** (`Issue529MenuValidationTests.swift:231-232`): each asserts the
inequality that the `.first { $0 > ... }` predicate used to build its operand already guarantees.
Only the preceding `#require` can fail. That is the third time in this branch I have written an
ordering assertion whose operand was found by the ordering — worth naming as a pattern rather than
fixing quietly: **if you locate a position by searching from X, you cannot then assert it is after X.**

MongLong0214:
Fixed, and the contradiction I flagged resolved in the direction I did not expect.

I asked for the State C / State B question to be worked out rather than bent to fit
`fallback_unsafe`'s State-C-only predicate. The answer: **State C is correct and my "attempted but
unverifiable is State B" instinct was wrong here.** Every unsafe-close path happens *before* the
target position is typed and submitted, so no position write occurred. Switching to State B would let
the dispatcher read the transport back and **promote it to State A** for an operation that never
wrote anything — the exact failure this branch has spent its rounds removing.

So it stays State C with `write_attempted: false`, and a separate `menu_actuation_attempted` says
truthfully that menu clicks happened. Two different facts, two fields, neither lying.

That envelope now also carries **`fallback_unsafe: true`**. Without it, `ax_write_failed` is
non-terminal and `ChannelRouter` walked on to MCU / CoreMIDI / CGEvent while the menu was known to be
unclosed — #526's defect in a place I had not looked. Using the mechanism #530 landed rather than
inventing a second way to say the same thing.

The submenu is no longer clicked on the strength of the menu-bar click's return; an observed open
menu comes in between.

Two more assertions could not fail, each comparing an index against one found by searching from that
same index. **Third occurrence of that shape on this branch** — worth stating as a rule rather than
fixing quietly: *if you locate a position by searching from X, you cannot then assert it is after X.*

Mutation-backed: setting `fallback_unsafe: false` fails
`"a post-click menu close failure records menu navigation, not a position write"`.

Full suite 3502. Live evidence re-recorded against this head.

MongLong0214:
One finding accepted, one pushed back with the repository's own definition.

**Accepted — the submenu click is still authorised by a call return, and the observation is not
identity-bound.** `enabled of mi` is not evidence the submenu opened (it was readable before the
click), and `menuOpenedAfterClick` accepts *any* open menu rather than the one that was clicked.
Both are the same mistake this branch has been fixing everywhere else, one level deeper. Being fixed
so the observation is bound to that element's own menu.

**Pushed back — State C with `menu_actuation_attempted: true`.** The review's premise is that
"State C requires nothing attempted". The repository defines it differently:

```swift
/// State C — hard failure: the write itself didn't succeed. `error` is
/// mandatory per the contract.
```

The axis is **the write**, not any actuation. Opening a menu mutates no project state, so
`write_attempted: false` is accurate and State C is the right state. The alternative — State B —
would let the dispatcher read the transport back and promote this to **State A** for an operation
that never wrote anything, which is a worse and far more likely failure than a debatable field on a
rare error path.

If the intended contract really is "no actuation of any kind", that is a change to the contract's
definition and to `HonestContract`'s doc comment, affecting every channel — not something to settle
inside a `goto_bar` fix. I would rather leave one field arguable than redefine a shared contract to
close out a PR.

MongLong0214:
Fixed. The submenu observation is now bound to the clicked element's own menu rather than to a global "some menu is open" state, and `enabled of mi` no longer stands in for evidence that the submenu opened — it was readable before the click.

Mutation-backed: replacing the element-bound read with an unconditional true fails `"an unrelated open menu cannot satisfy the submenu observation"`.

Full suite 3505. Live evidence re-recorded against this head.

MongLong0214:
Second finding fixed. `dismissOpenMenu` returned on an unreadable first read without sending Escape — right *before* this run clicks anything (nothing is open, and refusing there sent 5 of 8 fresh processes to the slider), wrong *after* it has clicked a menu open (a menu IS up, and an unreadable read is not permission to skip the dismissal).

The caller's context is now a parameter: the entry cleanup passes `false` — the one call that may skip — and all nine post-click cleanups pass `true`.

Mutation-backed: making the skip unconditional again fails `"an unreadable read skips Escape only before this run opened anything"`.

Full suite 3506. Live evidence re-recorded.

MongLong0214:
Both remaining findings fixed.

The post-click `UNREADABLE` hole was already closed in the previous push (the review ran against the diff before it): `dismissOpenMenu` now takes the caller's context, so only the entry call may skip on an unreadable read and all nine post-click cleanups always send Escape.

The new one is real and is fixed here: **every AppleScript error or timeout was classified fallback-safe.** This script clicks menus open, so its death leaves the menu state *unknown* — and the slider would actuate into whatever is on screen. The death path now runs an independent observation (is any menu bar item selected → Escape → look again). Only an observed CLOSED authorises the fallback; open or unreadable classifies as `menuCouldNotBeClosed`, because a failed reading is not evidence that nothing is open.

Mutation-backed: swapping the branches so an unknown state authorises the slider fails `"a dead script does not authorise the slider until the menu is observed closed"`.

Full suite 3507.

PR #536, opened 2026-08-11
fix(#533): teach the modal classifier the Japanese New Track sheet

### LINKED ISSUE #533: project.new fails closed on Japanese Logic: the ordinary New Tracks dialog is classified as unknownSheet
`project.new` fails closed on Japanese Logic because it does not recognise the ordinary New Tracks
dialog.

## Measured

Logic Pro 12.3 switched to Japanese (`defaults write com.apple.logic10 AppleLanguages -array ja`,
restarted). Driving `logic_project new` over the raw MCP protocol:

```json
{"error":"channels_exhausted",
 "hint":"{\"method\":\"accessibility\",\"operation\":\"project.new\",
          \"phase\":\"unexpected_blocking_sheet\",\"reason\":\"readback_unavailable\",
          \"safe_to_retry\":false,\"selection\":\"direct\",\"sheet_kind\":\"unknownSheet\",
          \"state\":\"B\",\"success\":true,\"verified\":false,\"write_attempted\":true}"}
```

**The project was created.** The window `名称未設定 - トラック` existed immediately afterwards.

The sheet it could not identify is the ordinary **New Tracks** dialog that Logic shows for every new
project:

```
sheet 1 of window "名称未設定 - トラック"
  AXStaticText  新規トラックを作成          <- title
  AXButton      作成                        <- Create
  AXButton      キャンセル                  <- Cancel
  AXStaticText  作成するトラック数:
```

Clicking `作成` by label dismissed it and the project was normal.

## Why it matters

This is not an exotic state. It is what a Japanese user sees on **every** new project, so `project.new`
is unusable there — and it reports failure for something that worked, which is the worse half: a
caller that retries gets a second project.

Failing closed on an unrecognised sheet is the right instinct. The defect is that a routine,
expected dialog is unrecognised, because the classifier's labels are English and Korean only.

## Fix direction

The Japanese labels above are measured, not translated, and can go straight into the sheet
classifier. Whatever else it recognises should be audited the same way rather than extended one
string at a time — this dialog was found by accident, and the honest assumption is that its
neighbours are missing too.

Related: the same locale gap in the marker-focus keyword set is noted on #530, and the measured
Japanese label set is on #528 and #519.

MongLong0214:
Closes #533.

## The defect

`project.new` failed closed on Japanese Logic. The mandatory New Track sheet that every new project
presents was classified `.unknownSheet`, so the operation reported failure — **for a project it had
in fact created**.

Same precondition (Logic running, no project open), same release artifact, minutes apart:

```
English    isError false   mandatory_track_created: true, "Untitled 56 - Tracks"
Japanese   isError true    unexpected_blocking_sheet, sheet_kind: unknownSheet
```

`ModalReconciliation.classify` identifies that sheet by its `AXDescription` or by its button shape,
and all three inputs carried English and Korean only.

## The fix, and why it is the description and not the buttons

Measured on the live Japanese sheet rather than translated:

| | measured |
| --- | --- |
| sheet `AXDescription` | `新規トラック` |
| Create button | `作成` |
| Cancel button | `キャンセル` |
| Cancel enabled? | **yes** |

That last row decides it. With Cancel enabled the button-shape disjunct
`(createButtonPresent && cancelButtonPresent && !cancelButtonEnabled)` cannot fire, so the
description match is the only path — which is exactly how English already passes. Adding the
description variant makes Japanese behave as English does; it does not widen the classifier, and a
test asserts an unrelated sheet with an enabled Cancel is still `.unknownSheet`.

The button variants are still needed, because the reconciler clicks the real localized title rather
than a hardcoded English one.

**Nothing else Japanese is added.** Other label sets in that file have the same gap. A guessed string
is worse than a visibly missing one — a test asserts `newTrackSheetDescription` does not contain the
plausible-but-wrong `新しいトラック`.

## Live evidence

Bound to this head, on a Japanese Logic Pro 12.3 with no project open:

```
{"mandatory_track_created": true, "phase": "created_project_window_observed",
 "window_title": "名称未設定 - トラック"}     isError: false
```

Three mutation-backed checks — the sheet is recognised, no sheet is left blocking the project, and
the menu bar really is Japanese so the Japanese path was the one exercised. Plus a visual assertion
that the created project **accepts an edit**: a marker made through Logic's own `移動 ▸ マーカーを作成`
appears in that window's ruler. I looked at the two crops rather than trusting the hash — `マーカー1`
is there in the second and absent in the first.

The visual is driven by Logic's own menu, not by this product, and that is deliberate. On Japanese
Logic almost nothing in the product is localized yet, so it cannot supply its own proof here:

- `nav.goto_bar` finds no `移動 ▸ 移動 ▸ 位置…` chain, falls to the slider, and does not move
  (documented on #531);
- `track.create_instrument` routes to a MIDI key command, fires CC 21 on a binding nothing listens
  to, and answers `state B, success: true` while the screen does not change — the same class of
  defect as #526, observed during this run and recorded rather than used as proof of anything.

Full suite: 3477 tests pass.

### DIFF
--- a/Sources/LogicProMCP/Accessibility/AXLocalePolicy.swift
+++ b/Sources/LogicProMCP/Accessibility/AXLocalePolicy.swift
@@ -244,27 +244,28 @@ enum AXLocalePolicy {
 
     static let cancelButton = LabelSet(
         canonical: "Cancel",
-        variants: ["취소"],
-        rationale: "Dialog dismissal fallback; no success state is inferred from this click."
+        variants: ["취소", "キャンセル"],
+        rationale: "Dialog dismissal fallback; no success state is inferred from this click. JA live-confirmed (Logic 12.3: `キャンセル`)."
     )
 
     /// #346/#350: the mandatory New Track sheet's only exit ("Create"). The modal
     /// reconciler clicks it to un-wedge Logic, then verifies via track-count
     /// readback — the click itself gates no State-A success. KO live-confirmed
-    /// (Logic 12.3: `생성`).
+    /// (Logic 12.3: `생성`); JA live-confirmed (Logic 12.3: `作成`).
     static let createButton = LabelSet(
         canonical: "Create",
-        variants: ["생성"],
-        rationale: "Mandatory New Track sheet's only exit; reconciler-clicked, then verified by track-count readback. KO live-confirmed (Logic 12.3)."
+        variants: ["생성", "作成"],
+        rationale: "Mandatory New Track sheet's only exit; reconciler-clicked, then verified by track-count readback. KO live-confirmed (Logic 12.3); JA live-confirmed (Logic 12.3: `作成`)."
     )
 
     /// #346/#350: `AXDescription` that identifies the mandatory New Track sheet.
     /// One of two signals (with disabled-Cancel) the reconciler uses to classify
-    /// the sheet; read-only classifier. KO live-confirmed (Logic 12.3: `새로운 트랙`).
+    /// the sheet; read-only classifier. KO live-confirmed (Logic 12.3: `새로운 트랙`);
+    /// JA live-confirmed (Logic 12.3: `新規トラック`).
     static let newTrackSheetDescription = LabelSet(
         canonical: "New Track",
-        variants: ["새로운 트랙"],
-        rationale: "Identifies the mandatory New Track sheet by AXDescription (with disabled-Cancel); read-only classifier. KO live-confirmed (Logic 12.3)."
+        variants: ["새로운 트랙", "新規トラック"],
+        rationale: "Identifies the mandatory New Track sheet by AXDescription (with disabled-Cancel); read-only classifier. KO live-confirmed (Logic 12.3); JA live-confirmed (Logic 12.3: `新規トラック`)."
     )
 
     /// #346/#350: primary destructive button on the "delete channel strips that
--- a/Tests/LogicProMCPTests/Issue533JapaneseNewTrackSheetTests.swift
+++ b/Tests/LogicProMCPTests/Issue533JapaneseNewTrackSheetTests.swift
@@ -1,0 +1,79 @@
+import Testing
+@testable import LogicProMCP
+
+@Suite("Issue #533 Japanese New Track sheet")
+struct Issue533JapaneseNewTrackSheetTests {
+    private func sheetSignals(
+        description: String,
+        createTitle: String,
+        cancelTitle: String,
+        cancelEnabled: Bool
+    ) -> ModalReconciliation.ModalSignals {
+        ModalReconciliation.ModalSignals(
+            sheetPresent: true,
+            sheetDescription: description,
+            createButtonPresent: true,
+            cancelButtonPresent: true,
+            cancelButtonEnabled: cancelEnabled,
+            deleteConfirmButtonPresent: false,
+            strayMenuOpen: false,
+            topLevelAlertPresent: false,
+            topLevelAlertButtonCount: 0,
+            topLevelAlertPrimaryButton: "",
+            createButtonTitle: createTitle,
+            cancelButtonTitle: cancelTitle,
+            deletePrimaryTitle: ""
+        )
+    }
+
+    @Test("Issue533 classifies the measured Japanese sheet with an enabled Cancel")
+    func measuredJapaneseSheetClassifiesViaDescription() {
+        let signals = sheetSignals(
+            description: "新規トラック",
+            createTitle: "作成",
+            cancelTitle: "キャンセル",
+            cancelEnabled: true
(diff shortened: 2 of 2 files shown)


MongLong0214:
Revised. The review's finding was right and is fixed at the actuator.

The reconciler reported `performed` from the AppleScript call's success — which describes an
attempted IPC request, not what Logic did with it. That is the same rule this repository already
applies to AX return codes, and downstream it matters: last-track deletion marks the reconciliation
confirmed and can return **State A** from its deletion count while Logic is still wedged behind the
sheet.

Each actuator now discards the call's result and **observes the outcome** — the main window's sheet
is gone, the classified alert window is gone, no menu bar item is selected — bounded at five polls of
50 ms. No observation, no `performed`.

The part worth pointing at: each observation distinguishes **"gone" from "could not read"**. A
missing AX window, an unreadable menu bar, or an unreadable window list all return false, because an
unavailable readback is not proof that anything closed. The alert check walks the full window list
looking for the classified element rather than re-reading "the first blocking window", which another
dialog moving in front would have answered wrongly.

Mutation-backed: restoring `return await ...executeAppleScript(script).isSuccess` fails
`"Issue533 does not report Create performed while the sheet remains"`.

The two weak assertions the review named are also addressed — the fixture now resolves Create/Cancel
presence from the label sets instead of hardcoding both booleans, and the "only measured labels"
claim was narrowed to what it actually checks.

The mechanism predates this branch for English and Korean; this branch turns it on for Japanese,
which is why it is fixed here rather than filed.

Live evidence re-recorded against this head on a Japanese Logic: three mutation-backed checks, and
the visual assertion that the created project accepts an edit (`マーカー1` appears in its ruler after
Logic's own `移動 ▸ マーカーを作成`). Full suite 3480.

MongLong0214:
Blocked again, and the sharpest finding is the same defect one level down from where I fixed it.

**1. The witnesses I added treat a failed read as absence.** I guarded the outer reads —
`mainWindow`, the app root, the menu bar — and was pleased with myself for distinguishing "gone" from
"could not read". But `firstSheet` and `AXHelpers.getChildren` **flatten an AX failure to `nil` /
`[]`**, so a transient error inside the witness reads as "the sheet is gone" and reports
`performed == true` with the sheet still up. That is exactly the failure this PR exists to remove,
reintroduced underneath the guard I was proud of.

Fixing it so absence is reported only when the read **succeeded and was empty** — an unreadable
window, child list, or window list all yield "not performed", the same as the outer guards.

**2. A call result still selects a second actuator.** `confirmDeleteTracksSheet` uses
`executeAppleScript(...).isSuccess` to decide whether to send Return as a fallback, so a false
negative actuates twice. The fallback will be chosen from an observation — the sheet is still
present — rather than from the first call's return value. If the sheet is gone, no Return is sent.

**3. An assertion that cannot fail:** `#expect(fixture.state.isSheetPresent)` where the fixture is
built with `dismissAfterClick: false`, so it has no path to become false.

All three are being fixed here. The first two are mine; the second is pre-existing code in the
function family this PR changes, and leaving it while claiming to have fixed the unverified-actuation
problem would be a half-truth.

MongLong0214:
**Reset to the measured locale fix. The observation work is split out as #538.**

I spent four rounds trying to fold the observed-effect contract into this PR. It produced two
regressions and no working witness, so it does not belong here.

What happened, because the shape of it matters:

1. The review was right that reporting `performed` from a call's return value is wrong.
2. I implemented the observed version — poll until the sheet is gone — and it **turned `project.new`
   from working into failing** on a real Japanese Logic: `mandatory_track_create_unconfirmed` for a
   sheet that did close.
3. That inverted the earlier result. The first version passed *because* of the bug: `firstSheet`
   flattened an AX read failure to `nil`, and the witness read that as "gone". Removing the
   flattening is correct, and it revealed that the disappearance **is not observable through the path
   the reconciler uses**. Raising the bound from 250 ms to 2 s changed nothing, so it is not timing:

   ```
   t=3.88s   project window appears WITH the sheet   (windows=1, sheets=1)
   t=5.90s   sheet gone                              (windows=1, sheets=0)
   t=5.94s   project.new returns "unconfirmed"
   ```

   An outside observer sees the close plainly. The in-process witness does not.

So this PR is back to `986352aa` — the locale fix alone, which passed its live gate on a Japanese
Logic with three mutation-backed checks and a visual assertion that the created project accepts an
edit.

That fix is worth landing on its own: **`project.new` currently fails 100% of the time on Japanese
Logic**, reporting failure for a project it did in fact create. This makes it behave as it does in
English.

The observation contract is #538, with the measurements and the two failed approaches written down so
the next attempt starts from what is known rather than from where I started. It explicitly keeps
"unreadable is not gone" — that part is right, and the temptation to walk it back to make a test pass
is exactly what #538 has to resist.

Issue #538, opened 2026-08-11
Modal reconciler reports performed from a call result, and the sheet's disappearance is not observable through the path it uses


MongLong0214:
The modal reconciler reports `performed` from a call's return value rather than from an observed
effect, and the obvious fix does not work — the disappearance is not observable through the path the
reconciler uses.

## The reporting defect

`AccessibilityChannel+ModalReconcile.perform` sets `performed` from `executeAppleScript(...).isSuccess`
and from `AXPress` results. Those describe an attempted IPC request, not what Logic did. Downstream,
last-track deletion marks the reconciliation confirmed and can return **State A** from its deletion
count while Logic is still wedged behind the sheet.

## Why the straightforward fix fails, measured

I implemented the observed-effect version — poll until the main window has no sheet — and it turned
`project.new` from working into failing on a real Japanese Logic:

```
project.new → phase: mandatory_track_create_unconfirmed      (the sheet DID close)
```

Sampling the window tree every 100 ms across the same operation, from outside the process:

```
t=3.88s   the project window appears WITH the New Track sheet   (windows=1, sheets=1)
t=5.90s   the sheet is gone                                     (windows=1, sheets=0)
t=5.94s   project.new returns "unconfirmed"
```

An outside observer sees the close plainly. The in-process witness does not, and **raising the bound
from 250 ms to 2 s did not change that** — so it is not a timing problem.

Worth stating because it inverts the usual reading: the previous version passed *because* of a bug.
`firstSheet` flattened an AX read failure to `nil`, which the witness read as "the sheet is gone".
Removing that flattening is correct, and it revealed that the question being asked is unanswerable
on this path.

The likely reason — **not confirmed, and the next attempt should confirm it before building on it** —
is that during `project.new` the project window is created and replaced, so a main-window reference
does not survive the operation being observed. The observable fact may be "a readable main window
exists and has no sheet", re-resolved from the application root each poll; an attempt at that still
failed its own replacement-window test, so the mechanism is not yet understood.

## Why this is filed rather than fixed in #536

#536 is a measured locale fix that makes `project.new` work on Japanese Logic at all — it currently
fails there 100% of the time. Four attempts at folding the observation contract into it produced two
regressions and no working witness. Holding a user-facing fix hostage to a harder contract problem
serves nobody, and shipping a half-working witness would be worse than shipping neither.

## What a fix needs

- Confirm what the witness actually observes per poll during a live `project.new` (instrument it;
  do not reason about it).
[...]

PR #539, opened 2026-08-12
fix(#523): delete through the Marker List's own menu, never on an AX return code

### LINKED ISSUE #523: navigate.delete_marker selects the row and returns State B without deleting it

MongLong0214:
Closes #523.

`navigate.delete_marker` selected the row and returned State B without deleting it. It now deletes through the Marker List's own Edit menu, and reaches **State A live** — Logic's own count readout 20 → 19 with the Marker List holding no keyboard focus, and the surviving marker set proving which row went.

## The keyboard fallback is gone

Three revisions answered "make the fallback Delete key safe" and each closed a narrower hole than the last: a frontmost sample, then an identity-bound focus guard, then `CGEvent.postToPid`. PID binding stops the key leaving Logic; it does not stop it leaving the marker table. If focus moves to a region, track or dialog inside Logic between the guard and delivery, Logic stays frontmost and the pid stays correct while Delete reaches the new key responder — and marker readback then reports State B while an unrelated object may already be gone. `postToPid` returns nothing to inspect.

Before removing it, measured live whether the row could be deleted by actuating an element instead:

| element | actions |
|---|---|
| marker table (`AXTable`) | none |
| marker row (`AXTableRow`) | `Delete` with `Target: 0x0`, `Selector: (null)` — not addressable by name |

There is no element actuator. So the operation refuses when the Edit-menu route is unavailable, and says which of three things happened: the menu was absent, the menu was unreadable, or the exact Delete entry was missing or disabled. That costs the nine Logic locales whose Edit and Delete labels this route does not yet recognise. The answer there is to add measured labels, not to post an untargeted key.

## What the live runs taught, in order

**A menu opened by AXShowMenu can close on its own inside a 250 ms poll.** Requiring two agreeing sightings made every delete refuse while the menu was open. Presence is acted on at first sighting; absence needs two readings; and a failed reading retires that poll rather than the observation — taking the first failure as final made deletes refuse intermittently.

**`attributeUnsupported` and `noValue` are answers, not failures.** A childless node and a table without `AXRows` report exactly those, so classifying them as unreadable made the whole readback fail. That distinction now lives on `AXStatusError` itself, because it had been rediscovered independently in two channels and got it wrong the second time.

**A search is not a readback.** Discovery walks twelve levels for the Edit control; one node answering `-25200` mid-walk is not proof the search failed — marker enumeration in the same call succeeded, so the API was plainly not disabled. Discovery skips what cannot answer and remembers that something could not, so "there is no Edit control" stays distinct from "none found and part of the tree was unreadable".

**But an incompletely read list is not an empty one.** Treating "no children" as an answer is right for asking whether a node has children and wrong for asking whether every row was enumerated. A row whose cells will not answer is a marker that could not be read; skipping it silently turned an incomplete enumeration into a claim of emptiness, and State A could be certified for a delete that never happened.

## Contract

State A requires a settled survivor readback, a unique target position, and canonical position evidence. A target position shared by more than one marker cannot reach State A, because the surviving multiset is identical either way.

Ship gate: 3528 tests, live evidence bound to HEAD with a screen recording and a region-bound visual assertion.


MongLong0214:
Four findings, all mine, all being fixed here.

**The menu is not bound to the control that opened it.** Falling back to "the first `AXMenu` anywhere
in the window" means a stale or unrelated menu containing a Delete entry could be actuated. That is
the same identity failure this repository keeps producing — verifying the *kind* of thing rather than
the *identity* of the one targeted. It will bind to the Edit control's own menu, and an unbindable
menu becomes `menuUnavailable` rather than something to pick from.

**Refusal branches leave the menu open.** The missing-entry and disabled-entry paths neither dismiss
nor verify dismissal, so they can press Delete into a still-open menu or return State C with
`safe_to_retry: true` while Logic is wedged. A refusal that leaves the app wedged is not a clean
refusal — the same discipline #529 landed applies here.

**Korean is missing.** I added Japanese and not Korean, so this PR would take the route *away* from
Korean users. Measured on Korean Logic 12.3: the toolbar Edit menu button is **`편집`** and its Delete
entry is **`삭제`**. Adding both, with the same exact-match requirement — Korean has the
Delete-Undo-History sibling too.

**Two checks that cannot fail** (`:184`, `:212`) require `menuEntryID` while every corresponding loop
passes a non-`nil` title. And there is **no test at all** asserting an opened menu was dismissed —
which is how finding 2 got through.

MongLong0214:
All four fixed. The menu is read from the Edit control's own children (an unbindable menu is `menuUnavailable`, not something to pick from); every path that opens a menu now dismisses it and observes the dismissal; Korean `편집`/`삭제` are in, measured on Logic 12.3, so this no longer takes the route away from Korean users; and the two always-true `menuEntryID` checks are rewritten with a new test asserting an opened menu was dismissed — the absence of that test is how the second defect got through.

Full suite 3496. Live evidence re-recorded against this head.

MongLong0214:
Both fixed.

The closure observation used `getChildren`, which turns an AX read failure into an empty array — so "no menu child" could not tell *the menu is gone* from *I could not read*. That could authorise the Delete-key second actuator with the menu still open. It now uses a status-preserving read, and anything other than an observed-empty result counts as still open.

A menu that ignored `AXCancel` was reported and left open — reporting is not clearing. The dismissal escalates (a menu that ignores AXCancel may still close on Escape), bounded, observing after each attempt. Only a menu still observed open, or unobservable, ends the operation.

Mutation-backed: making a failed read return "closed" fails `testIssue523UnreadableMenuClosureForbidsFocusFallback`.

Full suite 3498.

MongLong0214:
First finding is a real hole I left: I made the menu **closure** observation status-preserving and left the **discovery** one flattening. A failed read there becomes "menu unavailable" and sends the operation to the Delete key while the menu may be open. Fixing it with three outcomes — present, genuinely absent, and unknown — where unknown refuses after attempting and observing a dismissal rather than falling through.

Second finding, partly accepted. The operation cannot do more than exhaust its dismissals (`AXCancel`, then Escape, bounded, observing after each) and report — there is no further mechanism, and an unbounded retry is worse. What is wrong is the **test framing**: `#expect(fixture.menuIsOpen)` on a fixture that disables every close mechanism reads as though a wedged app were the desired outcome. The tests will assert what is actually under test — every dismissal was attempted, the Delete key was not posted, and the envelope reports the unclosed menu plainly — rather than asserting Logic's state as an expectation.

MongLong0214:
Both fixed.

I had made the menu **closure** observation status-preserving and left **discovery** flattening — fixing one of two reads in the same file. A failed discovery read became "menu unavailable" and sent the operation to the Delete key while the menu might be open. Discovery now has three outcomes: a bound menu is present; the read succeeded and there is none (may fall through); or the read failed, which is unknown and refuses after attempting and observing a dismissal.

The wedged-menu tests asserted `fixture.menuIsOpen` on fixtures with every close mechanism disabled — reading as though a wedged application were the expected outcome. They now assert what is under test: every dismissal was attempted, the Delete key was not posted, and the envelope reports the unclosed menu plainly. The operation genuinely cannot do more than exhaust its dismissals and say so; an unbounded retry would be worse.

Mutation-backed: making the discovery read return "no menu" on failure fails `testIssue523UnreadableMenuDiscoveryRefusesAfterObservedDismissal`.

Full suite 3518.

MongLong0214:
Accepted — and it is the same shape as what #538 just taught me.

After `AXShowMenu`, a single immediate child read finding no menu is taken as "no menu" and permits the Delete fallback. Logic can expose the menu **asynchronously**, so it may appear after that read, and the key then lands in the open menu.

Replacing the single read with a **settled observation** — poll with a bound, require consecutive agreement before believing either answer — which is the idiom this file already uses for the marker-list readback. Not settled, or a genuine read failure, means unknown: clean up and refuse rather than fall through.

The test fixture exposes the menu synchronously inside the action handler, so it cannot show this race. A fixture where the menu appears only on a later read is being added.

Carrying #538's finding across too: AX `-25205` (`attributeUnsupported`) and `-25212` (`noValue`) are **answers, not read failures** — classifying them as failures is what made that witness never settle. Checking whether these reads can return them.

Issue #537, opened 2026-08-11
goto_marker routes to channels that cannot carry its index or name


MongLong0214:
`nav.goto_marker` routes to `[.midiKeyCommands, .cgEvent]`, and neither channel can carry the target
the caller specified.

The command takes `{ index: Int }` or `{ name: String }`. A MIDI key command is a global trigger — it
carries no argument at all — and the CGEvent rung is a keystroke with the same limitation. So the
operation cannot honour its own parameters through either route; it can only fire a "go to next/some
marker" command and report whatever happens.

This is the same class as #526/#530, which was fixed for `nav.delete_marker` by removing the channel
that cannot express its target:

> a channel that cannot express the operation's target must not be in that operation's chain

`nav.delete_marker` is now accessibility-only for that reason, and `nav.rename_marker` already was.
`nav.goto_marker` is the remaining indexed marker operation still routed to channels that cannot
address a marker.

Found during the #530 review by auditing the indexed marker operations; reported rather than changed
there, because widening a blocker fix to a second operation would hide the second change inside the
first.

**What this needs before a fix:** whether an accessibility route exists for it at all. `goto_marker`
by name or index should be reachable through the Marker List the same way `delete_marker` is, but
that has not been measured. If no AX route exists, the honest outcome is a State C refusal rather
than a key command that fires at an unrelated marker.

MongLong0214:
**An accessibility route exists.** I measured it, so this is fixable rather than a refusal.

`Navigate ▸ Go To ▸ Marker Number…` opens a floating dialog that takes the marker's **number** —
which is exactly the argument the key-command rung cannot carry:

```
window "Go to Marker"  subrole=AXFloatingWindow
  text field 1  value="1"
  button "Cancel"
  button "OK"
```

Driven by label and AX value-set, no coordinates:

```applescript
set value of text field 1 of window "Go to Marker" to "15"
click button "OK" of window "Go to Marker"
```

Result, read off the transport LCD: the playhead moved to **bar 5**, which is where marker 15 sits in
that project (`logic://markers` reports `Marker 15 @ 5.1.1.1`; markers 1–14 are all at bar 1, and
jumping to 14 correctly left the playhead at bar 1 — a control that would have looked like "nothing
happened" if I had only tested that one).

## What the fix needs to get right

- **The dialog's number is the 1-based ordinal in the marker list; the MCP `index` is 0-based**
[...]

PR #540, opened 2026-08-12
fix(#535,#532): the focus guard must name its target, and a refusal must disclose the selection

### LINKED ISSUE #532: delete_marker answers State C with write_attempted:false after it has already written a row selection
`nav.delete_marker` can answer State C with `write_attempted: false` after it has already written a
selection to the Marker List.

## What happens

The operation selects the target row before checking whether the list holds keyboard focus:

```swift
guard selectMarkerRowForDeletion(index, in: window, runtime: runtime.ax) else { ... }
guard markerTableHasKeyboardFocus(runtime: runtime) else {
    extras["write_attempted"] = false      // <- a selection has already been written
    return .error(HonestContract.encodeStateC(error: .axWriteFailed, ...))
}
```

When the second guard refuses, the envelope says State C, `write_attempted: false` — but the
selection write landed.

## How it was observed

Not by reading the code. A live run asserted that the Marker List's rows would be pixel-identical
across a refusal, and the assertion failed. Both captures show all sixteen markers — nothing was
deleted — and the difference is that the row highlight moved.

## Why it matters

State C means "refused, nothing was attempted". A caller that retries elsewhere, or that reasons
about what the app looks like after a refusal, is being told the application was not touched. It was:
its marker selection is now different from what the user left, and Logic's marker selection is
visible and is an input to other operations.

Both readings of the fix are defensible and I am not choosing one here:

- narrow the claim — `write_attempted` refers to the delete, so say so, and report the selection
  write separately (for example a `selection_changed` field), or
- restore the previous selection before refusing, so the envelope's claim becomes true.

The second is the stronger contract and the more invasive change.

## Scope note

This predates #526/#530 and is not caused by it. #530 deliberately did not widen its change to cover
this — fixing a router-level defect and quietly redefining what `write_attempted` means in the same
commit would hide the second change inside the first.

### LINKED ISSUE #535: delete_marker's focus guard accepts any marker table, not the one the row was selected in
`nav.delete_marker`'s focus guard checks that *a* marker-labelled table has keyboard focus, not that
it is **the** table in the window where the requested row was selected.

`AccessibilityChannel+MarkerDelete.swift` selects the row against a specific window binding, then
verifies focus with `markerTableHasKeyboardFocus`, which matches on the marker container labels. With
two projects open, each with a Marker List, the guard can be satisfied by the other project's list —
and the Delete keystroke then goes there.

The operation already holds the binding it selected against. The check should compare the focused
table and its window against that binding with `CFEqual`, so "focused" means "the one I targeted"
rather than "one that looks like the right kind".

This is the same class as the repository's recorded rule that a fail-closed check which passes for
the wrong target is still a false pass: verifying the *kind* of thing rather than its *identity*.

Found while reviewing #530. Pre-existing — #530 does not introduce or touch this guard, and marking
the refusals `fallback_unsafe` does not affect which table is checked.

MongLong0214:
Closes #535. Closes #532.

## #535 — the guard verified a kind, not an identity

`markerTableHasKeyboardFocus` checked that *a* marker-labelled table had focus, not that it was **the**
table the row was selected in. With two projects open, each with a Marker List, the other project's
list satisfied it and the Delete keystroke went there.

`selectMarkerRowForDeletion` now returns the table it selected in, and the guard compares that table
and its window against the binding with `CFEqual`.

## #532 — a refusal that said the application was untouched

Selecting a row is an AX write: the highlighted row moves, it is visible on screen, and it is an
input to other operations. The focus refusal reported `write_attempted: false` and nothing else, so a
caller was told nothing had happened when its selection had changed.

`write_attempted` stays scoped to the destructive Delete — widening it would blur a field other paths
depend on — and the refusal now carries `selection_changed: true` as well.

## Two tests, because one could not prove what it claimed

The two-project test passes with **either** `CFEqual` removed: the table check and the window check
cover for each other there, so neither was individually necessary. I found that by mutating them one
at a time rather than trusting the test's own mutation note.

The second test puts a sibling table in the **same window**, where only the table comparison can
reject it. Removing that comparison posts the Delete key — `deleteKeyPosts.value == 1` — and the test
catches it.

## One fixture fix worth naming

`FakeAXRuntimeBuilder.setChildren` did not wire `AXParent`. A real AX tree always has it, and code
that walks upward from a focused element reads it — so every such walk failed silently in fixtures
and made the identity guard look broken. It now wires both directions.

## #537 is not part of this and is being corrected separately

I filed #537 claiming `goto_marker` routes to channels that cannot carry its index or name. **That is
wrong.** `NavigateDispatcher` has resolved the marker from cache and navigated via
`transport.goto_position` since v3.1.10, and its comment documents the exact defect I "found" as
already fixed. An attempt to add an accessibility route broke three existing tests that assert the
translation, which is how I learned it. Those changes are reverted from this branch.

Full suite: 3490 tests pass. Live evidence recorded against this head.

### DIFF
--- a/Sources/LogicProMCP/Accessibility/AXHelpers.swift
+++ b/Sources/LogicProMCP/Accessibility/AXHelpers.swift
@@ -13,6 +13,9 @@ enum AXHelpers {
         let childCount: @Sendable (AXUIElement) -> Int?
         /// Injectable typed-children seam. `nil` means "use the production read".
         let childrenResult: (@Sendable (AXUIElement) -> Result<[AXUIElement], AXStatusError>)?
+        /// Injectable status-preserving attribute seam. `nil` falls back to the ordinary
+        /// attribute reader, whose historical `nil` result has no status information.
+        let attributeValueResult: (@Sendable (AXUIElement, String) -> Result<AnyObject?, AXStatusError>)?
 
         /// Explicit memberwise initializer, replacing the compiler-synthesized one so
         /// `childrenResult` can default to `nil`. Every existing construction that omits
@@ -24,7 +27,8 @@ enum AXHelpers {
             children: @escaping @Sendable (AXUIElement) -> [AXUIElement],
             performAction: @escaping @Sendable (AXUIElement, String) -> Bool,
             childCount: @escaping @Sendable (AXUIElement) -> Int?,
-            childrenResult: (@Sendable (AXUIElement) -> Result<[AXUIElement], AXStatusError>)? = nil
+            childrenResult: (@Sendable (AXUIElement) -> Result<[AXUIElement], AXStatusError>)? = nil,
+            attributeValueResult: (@Sendable (AXUIElement, String) -> Result<AnyObject?, AXStatusError>)? = nil
         ) {
             self.axApp = axApp
             self.attributeValue = attributeValue
@@ -33,6 +37,7 @@ enum AXHelpers {
             self.performAction = performAction
             self.childCount = childCount
             self.childrenResult = childrenResult
+            self.attributeValueResult = attributeValueResult
         }
 
         static let production = Runtime(
@@ -70,6 +75,14 @@ enum AXHelpers {
                 let result = AXUIElementGetAttributeValueCount(element, kAXChildrenAttribute as CFString, &count)
                 guard result == .success else { return nil }
                 return count
+            },
+            attributeValueResult: { element, attribute in
+                var value: AnyObject?
+                let status = AXUIElementCopyAttributeValue(element, attribute as CFString, &value)
+                guard status == .success else {
+                    return .failure(AXStatusError(raw: status.rawValue))
+                }
+                return .success(value)
             }
         )
     }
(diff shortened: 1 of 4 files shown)


MongLong0214:
Fixed. `selection_changed` was set unconditionally, but the post-write `AXSelectedRows` read only proves the target is selected **now** — not that this call changed it. A row that was already the sole selection is not a change, and claiming one asserts an effect that did not happen.

The selection is read before the write and compared. Two fields, two facts: `selection_write_attempted` is what the call did, `selection_changed` is what it altered.

Mutation-backed: making it unconditional again fails `"an already-selected row is not reported as a selection change"`.

On the cross-project test — you are right that it does not isolate the new guard, and I found the same thing by mutating each `CFEqual` separately: it passes with either removed. That is why the sibling-table test exists, where only the table comparison can reject. I am keeping both, the weaker one as a scenario record, with the sibling test carrying the proof.

Full suite 3491.

MongLong0214:
Blocked on the same defect class the whole session keeps producing, this time in the fix for it.

The pre-write `AXSelectedRows` read yields `[]` on failure, which is indistinguishable from "nothing was selected". So when that read fails and the target was **already** selected, the code concludes the selection changed and reports `selection_changed: true` for a write that altered nothing — the exact over-claim this field was added to remove.

Being fixed with a status-preserving read, and a third answer: read succeeded and it was already the sole selection → false; succeeded and it was not → true; **read failed → unknown**, reported as unknown rather than guessed either way.

Second finding accepted: `#expect(deleteKeyPosts.value == 0)` at line 138 cannot fail for the regression it names — the focused table carries no marker-matching `AXDescription`, so the old guard refuses too. I found the same thing myself by mutating each `CFEqual` separately; that is why the sibling-table test exists. Either the fixture gets a focused table the old guard would have accepted, or the assertion goes and the sibling test carries the proof.

Also restored `Package.resolved` in that worktree — the review's `--skip-build` run rewrote it and its sandbox could not put it back.

MongLong0214:
Fixed. The pre-write `AXSelectedRows` read now preserves its status, and the answer has three outcomes instead of two: read succeeded and the target was already the sole selection → false; succeeded and it was not → true; **read failed → unknown**, omitted rather than guessed either way. `selection_write_attempted` still records what the call did.

Worth naming: this was the same over-claim reproduced *inside the fix for it*. The field exists because reporting an effect that did not happen is wrong, and the implementation then treated a failed read as evidence of that effect.

Mutation-backed: collapsing unknown back into "changed" fails `"an unreadable prior selection leaves selection change unknown"`.

Full suite 3492.

Issue #543, opened 2026-08-15
logic_mixer.set_volume / set_pan: element_not_found for any track, in Mixer and Tracks view


edgarperatla-create:
## Summary

`logic_mixer.set_volume` and `logic_mixer.set_pan` fail with `element_not_found` for **any** track index, in both the Mixer view and the Tracks/Arrange view.

## Environment

- `LogicProMCP` v3.13.0 (installed binary, SHA256-verified release)
- Logic Pro 12.3
- macOS, arm64

## Repro

1. Open a project with at least a few tracks (confirmed with a 9-track project).
2. Call `logic_mixer` with `{"command": "set_volume", "params": {"track": 0, "value": 0.75}}`.
3. Observe:
   ```json
   {"error":"element_not_found","hint":"Cannot locate volume control for track 0","operation":"mixer.set_volume","recovery_hint":"Ensure track 0 exists and the Tracks area is shown.","requested":0.75,"state":"C","success":false,"target_identity":{"control":"volume","track_index":0},"track":0}
   ```
4. Tried track indices 0, 1, and 8 (last track) — same error every time.
5. Tried with the Mixer window explicitly open (`logic_navigate.toggle_view` → `mixer`) and with it closed (Tracks/Arrange view only, matching the recovery hint) — same error in both states.
6. `set_pan` fails identically (`Cannot locate pan control for track <n>`).

Meanwhile `mute`/`solo` on the same tracks work fine (`keyboard-mute`/`keyboard-solo`, AX-verified), so this looks specific to the volume/pan control lookup path, not a general AX connectivity issue.

## Why I'm filing this

Found while doing due-diligence testing before bifurcating the project (disclosure: I'm forking this for a personal-permissions-model project). Happy to share more detail (traces, project file) if useful for reproducing.

MongLong0214:
Thanks — this is a well-scoped report, and the mute/solo-still-work detail turned out to be the key.

**I could not reproduce it here**, so before anything else, what I actually ran, against a build from
current `main` (not the v3.13.0 release binary you used), Logic Pro 12.3:

- grew the project to 9 tracks, then `set_volume` on indices 0, 4 and 8
- both cold (fresh server process, no prior call) and after a `logic_tracks.select`
- both in the Tracks/Arrange view and with the Mixer open via `logic_navigate.toggle_view`

Every one returned `state: "A"`, `verified: true`, `verify_source: "ax_slider"`. `set_pan` likewise.

[...]

PR #558, opened 2026-08-17
feat(#543): make the mixer refusal say which lookup step failed

MongLong0214:
**Observability, not a fix — and it says so.** #543 does not reproduce here: `set_volume` and `set_pan`
both return State A on this machine, at several track counts, in both views. There is nothing to fix
against yet.

`element_not_found` answered "Cannot locate <label> control for track <n>" and nothing else. Three
different failures produce that identical message and each needs a different fix, so neither the reporter
nor anyone reading the issue can tell them apart. The receipt now names the step:

```
track_header_list_not_found           the track-header list was not found at all
no_header_at_index                    headers exist, none at this index — a count mismatch
header_has_no_slider_within_depth_4   header found, no AXSlider inside it
sliders_present_but_none_selected
```

with `header_count`, `requested_index`, and `sliders_in_header` when a header was found.

**The middle step is the one I would look at first.** `allTrackHeaders` keeps only `AXLayoutItem` direct
children when any exist, so a header of another role is dropped and every index after it shifts — which
presents exactly as this issue describes, every index failing, and is invisible in the current receipt.
That enumerator was also flagged as a deferred MAJOR during #542's review.

**Structural facts only.** Counts, an index, a step name — no track name or other user content. A test
builds a sentinel track name and asserts it never appears in the serialized receipt, because a diagnostic
carrying the user's project cannot be pasted into the report it exists to serve.

| | |
|---|---|
| Suite | **3765 passed** |
| Ship gate | **PASS** |
| Live | 5/5, mutation-backed, visual passing, captures settled |

Additive only: `error`, `state`, `operation`, `track`, `requested`, `target_identity` and `recovery_hint`
keep their existing names and meanings, asserted by a test rather than assumed.

Also here, from fighting this branch's own live runs: the harness now raises the arrange WINDOW rather
than just activating the app (the menu drive needs it frontmost, and the Marker List was taking that
position), and moves the Marker List off the region it photographs instead of photographing somewhere
else — the previous region never settled, because Logic animates indicators there.

### DIFF
--- a/Scripts/livekit/live_543_lookup_names_the_step.py
+++ b/Scripts/livekit/live_543_lookup_names_the_step.py
@@ -1,0 +1,168 @@
+#!/usr/bin/env python3
+"""Live proof for #543's mixer control-lookup diagnostics against the running Logic Pro.
+
+Usage:  LPM_EVIDENCE_ROOT=/abs/path/outside/repo \
+        python3 live_543_lookup_names_the_step.py <worktree> <full-40-char-head-sha>
+
+What this run is actually for, stated plainly rather than implied.
+
+The #543 fix replaced `allTrackHeaders` with `allTrackHeadersRead` in the refusal path. Those two are
+NOT the same reader. The old one is tolerant: any traversal failure becomes an empty array. The new one
+is strict and status-preserving: it distinguishes "rail absent", "rail found but not enumerable", and
+"rail read, zero rows". That is the whole point of the change — but it is also the risk the change
+introduces, and it is a risk only the real application can settle. If Logic's live track-header rail does
+not satisfy the strict reader, then every real-world lookup failure would now report
+`track_header_list_unreadable` instead of naming the step, and the fix would have made the diagnostic
+worse than the bug it replaced. No unit fixture can answer that: the fake tree is built to be readable.
+
+So the load-bearing check is: drive a genuinely out-of-range index at the REAL rail and confirm the
+receipt says `no_header_at_index` with a truthful `header_count` — which is possible only if the strict
+reader successfully enumerated the live rail.
+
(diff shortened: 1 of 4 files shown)


MongLong0214:
Review fixes pushed @ 5f05b37d.

**[11] `allTrackHeaders` conflates three answers** — fixed. The refusal path now reads through `allTrackHeadersRead` and reports the three separately: `track_header_list_not_found`, `track_header_list_unreadable`, `track_header_list_empty`. `header_count` is emitted **only** when a read actually succeeded; publishing `0` for an unreadable rail asserts "this project has no tracks" from a read that never happened.

**[12] re-read race** — fixed. The header is indexed out of the array already read, so the receipt can no longer contradict itself with `header_count: 3` beside `no_header_at_index` for index 1.

Evidence:
- Unit: 8/8, and **mutation-tested** — collapsing the three answers back into "not found, count 0" fails all three new/changed tests and leaves both controls passing.
- Live: new `live_543_lookup_names_the_step.py`, 5/5 against running Logic. The load-bearing check drives an out-of-range index at the **real** rail and asserts `no_header_at_index`, which only the `.read` branch can produce — so it witnesses that the *stricter* reader this fix switches to can actually enumerate Logic's live rail. That is the risk the fix introduces and no fake tree can settle it, since the fake tree is built to be readable.
- The harness was itself mutation-tested against the running app: forcing the reader to refuse turns exactly 2 of the 5 checks red.

Not proven live, and said so in the docstring rather than implied: the `.unreadable` branch (cannot force a real `kAXErrorCannotComplete` on demand) and the re-read race (cannot force the timing window). Both are locked by fixtures.

Ship gate PASS — 3768 tests, live evidence bound to head.

Issue #554, opened 2026-08-16
mergeReconcileExtras: conditionally-written keys can survive a second merge, and only some of them should


MongLong0214:
Found while fixing a BLOCKER in #553, where the new `reconciled_modal_unreadable_node` key could carry a
node from an earlier poll's sheet scan into a receipt whose reason came from a later, unrelated failure.
That key is now written totally (every branch sets or removes it). **The mechanism is not confined to it.**

### The mechanism

`mergeReconcileExtras` writes keys conditionally and never removes them, and `verifyTrackCreation` merges
**twice into one dictionary** — the post-menu outcome, then `var merged = extras` plus the final poll's
outcome (`AccessibilityChannel+Tracks.swift`, the merge at ~:1383 and the merge at ~:1472). Any key set by
the first merge and not set by the second survives into the final receipt.

`reconciled_modal_unreadable_reason` is immune because both merges write it whenever a reason exists. Every
other key is conditional:

| key | written only when |
|---|---|
| `reconciled_modal_kind`, `reconciled_action` | `kind != .none` |
| `new_track_dialog_auto_confirmed` | that flag is true |
| `reconciled_modal_observation` | the `kind == .none` branch |
| `modal_reconciliation_witness` | a witness summary exists |
| `reconcile_refused` | a refusal exists |
| `reconcile_action_error` | an action failure exists |
| `reconciled_modal_unreadable_ax_status` / `_ax_status_name` | the reason carries a status |

There is also an early `guard kind != .none || unreadableReason != nil else { return }`. When the second
merge's outcome is clean and complete, it returns immediately and **every** key from the first merge
survives untouched.

The `_ax_status` pair is the clearest defect of the set: a `stray_menu_read_failed` reason carries no
status, so a `-25200` from an earlier sheet-scan failure can remain beside it, describing a different
failure entirely.

### Why a blanket fix would be wrong

The first merge is deliberate history, not an accident. From `+Tracks.swift`:

> `// #348: note any pre-create reconciliation (alert acknowledged / stray menu escaped).`

So `reconciled_modal_kind` and `reconciled_action` surviving from the first merge is the intended
behaviour — the receipt is meant to record that an alert was acknowledged before the write, even though
the final poll saw nothing. Making every key total would erase that and regress #348.

The fields therefore split into two groups, and the split has to be decided per field:

- **latest-wins** — anything that annotates `reconciled_modal_unreadable_reason`, because it describes
[...]

Issue #549, opened 2026-08-16
Track operations withhold State A when the Marker List holds a row: one unreadable AXCell vetoes the blocker scan


MongLong0214:
> **Read this first.** This issue originally claimed a specific cause — an unreadable `AXCell` in the
> Marker List window vetoing the blocker scan — and called it proven by intervention. **That was
> retracted**; the intervention had a confound and the reproduction has since been lost. The original
> claim and the retraction are kept in the comments rather than edited away. What follows is what is
> actually known.

## Observed

`track.create` and `track.delete` sometimes answer Honest Contract State B `retry_exhausted` where State
A was earned. The write itself succeeds every time this has been seen — the track is created or deleted,
the count is correct, and no blocker is left on screen. Only the certification is withheld.

The receipt in the failing runs:

```
"reconciled_modal_observation": "incomplete",
"reconciled_modal_unreadable_reason": "window_sheet_read_failed",
"reconciled_modal_unreadable_ax_status": -25200
```

Seen repeatedly across a stretch of one session, on both operations, against Logic Pro 12.3.

## Not currently reproducible

Later in the same session, with no code change, every condition that had produced State B produced State
A instead — three consecutive creates, with the auxiliary window open, with it raised, with Logic in the
foreground and in the background. **The trigger is unknown and intermittent.** It was present for a
period covering many operations and is absent now.

This matters for planning: **this is not an issue that can be closed by writing a fix today.** Nothing can
be verified fixed while the failure cannot be produced on demand.

## Severity: wrong in the safe direction

State B where State A was earned. The write succeeded and the caller is told it is unverified — never the
reverse. A caller that reads `observed_delta` gets the truth. This is a usability defect, not an honesty
one, and it does not have the property that makes a false State A dangerous.

## What was ruled out, with measurements

- **Stale window element** — the arrange window element was held across a create; its `AXChildren` read
  stayed `success/11` throughout. Not invalidated.
- **A transient during the post-mutation rebuild** — a 30 Hz walk across a create, 581 passes over 22
  seconds, found one `-25204` and no `-25200`.
- **The Marker List `AXCell`** — that node does return `-25200` on every census, reproducibly. But
  operations certify normally while it is present, so it is not sufficient to cause the failure.

Also measured: Logic answers `AXSheets` with `-25205` on every window, sheet or no sheet, and briefly
`-25204` while busy. It does not return `-25200` there.

## First task: make the failure name itself

[...]

Issue #348, opened 2026-07-12
Strict-live fresh bootstrap flakes on single-OK top-level alert (audio-interface warning) during first-track creation; extend modal reconciliation to the create path


MongLong0214:
## Summary

The strict-live E2E fresh bootstrap intermittently fails with `fresh_track_ui_not_ready` ("Logic kept the project picker or blocking dialog open before first track creation") when Logic pops a **single-OK top-level alert during first-track creation**. Observed live (2026-07-12): Logic's audio-interface warning —

> **The previously selected audio interface is not available.**
> The built-in audio inputs and outputs of the Mac mini will be used instead for this session. **[OK]**

— a top-level `AXWindow subrole=AXDialog desc="alert"` with a single `OK` button. It appears when Logic initializes audio for a freshly-created document, which is **after** the bootstrap's dialog-clear loop, so the bootstrap's `clear_hidden_blocking_dialog` (Return) / `dismiss_safe_blocking_dialog` (Cancel/Escape) never fire against it, and the subsequent `create_instrument` is blocked.

It is intermittent (depends on audio-hardware state): the same fixed binary passed §17b Fresh Record Modal Guard in one run and hit this alert in the next.

## Impact
- Flaky §17b Fresh Record Modal Guard (7 dependent checks fail together) under strict-live E2E.
- Not a product regression — it is an environmental Logic startup alert; but a real MCP client whose Logic pops such an alert while the server is mid-flow would be blocked until it is dismissed.

## Proposed fix (two complementary options)
1. **Server-side (preferred, fundamental):** wire a **narrow** `reconcilePreflight` into the `create_instrument` / first-track path — acknowledge single-button top-level alerts (added in #347) and Escape a stray menu, but **NOT** click-Create on a mandatory New Track sheet from preflight (that would double-create alongside `create_instrument`'s own New Track handling). This makes the create path self-heal a stray blocking alert, and the E2E bootstrap (which calls `create_instrument`) inherits it.
2. **Harness-side:** in `Scripts/logic_session_bootstrap.py`, re-check for a blocking dialog at the first-track-create retry step and click a single-`OK` alert's button (not just Return/Cancel/Escape) before giving up with `fresh_track_ui_not_ready`.

Related: #346 / #347 added the server-side reconciler that already classifies + acknowledges this exact single-button top-level alert on the **delete** path; this issue is about extending that reconciliation to the **create / bootstrap** path.

PR #559, opened 2026-08-17
fix(#554): the AX status pair moves with the reason it annotates

MongLong0214:
Fixes the unambiguous half of #554. The rest is deliberately left open.

`verifyTrackCreation` merges twice into one dictionary, and the status keys were written only when the
reason carried a status. A later reason that carries none — `stray_menu_read_failed`,
`descendant_traversal_depth_cap` — left the earlier reason's status standing beside it, so a receipt could
read `stray_menu_read_failed` with a `-25200` that belonged to a sheet scan in an earlier poll.

**A number describing a different failure is worse than no number**, because it is specific and therefore
believed. That is the property the #549 node field exists to protect, and this was the same defect one
field over.

`ModalReadFailure.apply(to:)` now writes the reason and its status pair totally — set when there is one,
removed when there is not — and every site that copied `envelopeExtras` key by key calls it.

### What is NOT fixed here, and why

#554 lists the other conditionally-written keys in that function. They are untouched on purpose: some of
that survival is **intended**. #348 records a pre-create reconciliation deliberately, so
`reconciled_modal_kind` surviving the first merge is documented behaviour, not a bug. Deciding which of the
remaining fields are latest-wins and which are historical changes published receipt meaning — wider than
this defect needed. Two of them I could not classify from the code at all, and said so in the issue rather
than guessing.

These two are unambiguous because they annotate whichever reason is currently reported.

| | |
|---|---|
| Suite | **3763 passed** |
| Ship gate | **PASS** |
| Live | 5/5, mutation-backed, captures settled, visual passing |

Mutation: removing the clearing branch reproduces the stale pair exactly — `-25200` and `failure` sitting
beside `stray_menu_read_failed`.

### DIFF
--- a/Scripts/livekit/live_549_cell_does_not_veto.py
+++ b/Scripts/livekit/live_549_cell_does_not_veto.py
@@ -137,12 +137,16 @@ if present is not True:
 
 rec = ev.record_screen(seconds=100)
 win = E.logic_window()
-# NOT the track rail: the Marker List window floats over the left of the arrange window, and it has to
-# stay open because it carries the trigger. A crop there shows the Marker List, which does not change when
-# tracks are added, so the assertion compared two identical wrong pictures and read as "nothing happened"
-# on a run that plainly worked. Sample the arrange lanes to the right of it instead.
-RAIL = (int(win["w"] * 0.45), int(win["h"] * 0.15),
-        int(win["w"] * 0.45), int(win["h"] * 0.60)) if win else None
+# Move the Marker List OFF the track rail rather than photographing somewhere else. It has to stay OPEN —
+# it carries the trigger — but it does not have to sit on top of the subject. Two earlier region choices
+# failed for opposite reasons: cropping the rail while it was covered compared two identical pictures of
+# the Marker List, and cropping the arrange lanes instead never SETTLED, because Logic animates indicators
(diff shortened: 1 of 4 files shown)


MongLong0214:
Review fixes pushed @ 5d8e3d9e.

**[9] `apply(to:)` destroys a real status** — fixed, and the finding was right. The pair was gated on `status.symbolicName`, which is `nil` for the two production sentinels (`malformedAttribute`, `malformedChildren`) that sit deliberately outside `AXError`. For those failures the receipt dropped a status it **had**, asserting no unreadable status existed — for exactly the reads those sentinels were introduced to make visible.

Now gated on the status itself. One wrinkle worth naming: publishing `status.diagnosticLabel` unconditionally, as suggested, would have regressed the common case, because `diagnosticLabel` renders an ordinary status as its bare number (`cannotComplete` → `-25204`). So it is `symbolicName ?? diagnosticLabel`, and there is a control test holding that line.

**[10] `envelopeExtras` has no readers** — removed.

Evidence:
- Unit: 5/5, **mutation-tested** — restoring the `symbolicName` gate fails the sentinel test and leaves the other four passing, confirming the mutation is narrow and detected.
- Live: `live_538_modal_reconcile.py` 6/6 against running Logic at this head.

Ship gate PASS — 3765 tests, live evidence bound to head.

Issue #552, opened 2026-08-16
Make the preflight impossible to walk past: stamp the tree, refuse the push without it


MongLong0214:
`Scripts/lpm-ship.sh` exists because of this failure, and names it in its own header:

> running the preflight, seeing FAIL, and pushing anyway because the commands were chained with `&&` so
> the failure scrolled past (2 times)

It happened a third time today:

```sh
bash lpm-public-surface-preflight.sh ... | tail -2 && git push ...
```

`tail` exits 0, so the pipeline exits 0, so `&&` proceeds. The preflight printed `PREFLIGHT FAIL` on
screen and the push went out anyway. The rejection was harmless that time — a `Package.resolved` dirty in
the worktree but absent from the commit — but the mechanism that allowed it does not know that.

**Two occurrences read as human error; three is a property of the arrangement.** The script that prevents
this only prevents it when it is used, and hand-chaining recurs exactly when someone is in a hurry, which
is when the gate matters most.

### Proposal

Move the check from a place that must be *invoked* to a place that must be *passed*:

1. A passing preflight writes a stamp for the state it validated.
2. A `pre-push` hook refuses to push a state with no valid stamp.

Then how the preflight was invoked stops mattering — `&&`, a lost exit code through `tail`, or not running
it at all all end the same way.

### Key the stamp to the TREE hash, not the commit SHA

Today `8895070b` was amended to `d45f6bc3` to reword a commit message. Same tree: `c820f0860b8d` for both.
Keyed to the SHA, that amend would have invalidated a stamp backed by a 400-second suite run, forcing a
re-run for a change that altered no bytes of content — and a gate that demands pointless work is a gate
people learn to bypass. Keyed to the tree, a stamp is invalidated exactly when the content changes, which
is exactly when it should be.

### Implementation note

`.git/hooks/pre-push` is currently a CommitLore shim that overwrites itself on reinstall, so editing it
directly will not survive. It has a documented chaining seam — measured in the installed v1.0.2 shim:

```sh
chained="$hook_dir/pre-push.commitlore-chained"
if [ -x "$chained" ]; then
  "$chained" "$@" || exit $?
fi
```

So the stamp check belongs in `pre-push.commitlore-chained`, which the shim runs first and preserves
across reinstalls. Note the execute bit is load-bearing: the shim only runs it if `-x`.

Not urgent, and deliberately not being done during the current release work. Filed because the next
occurrence may not be a harmless one.

[...]

PR #560, opened 2026-08-17
feat(#552): make the preflight impossible to walk past

MongLong0214:
`lpm-ship.sh` exists because the preflight kept being walked past, and names the failure in its own header.
It happened a third time today:

```sh
bash preflight ... | tail -2 && git push ...
```

`tail` exits 0, so the pipeline exits 0, so `&&` proceeds. `PREFLIGHT FAIL` printed to the screen and the
push went out anyway. **Two occurrences read as human error; three is a property of the arrangement** — a
script that prevents this only prevents it when it is used, and hand-chaining recurs exactly when someone
is in a hurry, which is when the gate matters most.

So the check moves from something that must be **invoked** to something that must be **passed**: a passing
preflight records a stamp, and a pre-push hook refuses a tree that has none.

### Keyed to the tree, not the commit

`8895070b` was amended to `d45f6bc3` today to reword a message — same tree, `c820f0860b8d`, both times.
Keyed to the SHA, that amend would have invalidated a stamp backed by a 400-second suite run, forcing a
re-run for a change that altered no bytes. **A gate that demands pointless work is one people learn to
route around.** Keyed to the tree, a stamp dies exactly when content changes.

### Two deliberate limits

`record` does **not** run the preflight itself. A recorder that also verified would be trusted to do both
and could then certify its own work; this only writes down a result something else produced.

The hook installs in the CommitLore **chained** slot, because `.git/hooks/pre-push` is a shim that
rewrites itself on reinstall and runs `pre-push.commitlore-chained` first — measured in the installed
v1.0.2 shim, where the execute bit is load-bearing. Install is one symlink, documented in
`Scripts/README-hooks.md`.

### Watched to refuse

`Scripts/test-preflight-stamp.sh`, registered in CI beside the other guards:

```
unstamped tree            → 1
stamped tree             → 0
unresolvable tree-ish    → 2      distinct, so a rule tolerating one cannot silently tolerate the others
hook, unstamped commit   → refused
hook, stamped commit     → allowed
hook, branch deletion    → allowed      no tree to have verified
```

Scripts and CI only — no Sources change, so the ship gate correctly reports live evidence as `n/a`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,6 +28,11 @@ jobs:
         # prove nothing. Fail fast (no Xcode needed) before the build if any reappear.
         run: bash Scripts/ci-forbid-dead-expect.sh
 
+      - name: Self-test the preflight stamp gate (#552)
+        # The stamp only helps if it actually refuses. This asserts NOT-STAMPED is 1, stamped is 0, an
+        # unresolvable tree-ish is 2, and the hook refuses an unstamped commit while allowing a deletion.
+        run: bash Scripts/test-preflight-stamp.sh
+
       - name: Self-test the menu-literal guard (#519)
         # A guard nobody has watched fail is decoration. This asserts each known evasion is caught, that
         # legitimate code is not flagged, and that a tree the guard cannot examine is never reported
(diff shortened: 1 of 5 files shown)


MongLong0214:
Review fixes pushed @ b83046aa.

**[1] BLOCKER — the stamp survives `--amend`** — fixed, and this was the worst finding in the set because it was mine at its proudest point: the gate certified exactly the class it exists to refuse, via the one operation I had documented as safe.

`stamp_key()` hashed the tree alone. Preflight check C1a greps **commit messages**, so `git commit --amend -m "<forbidden metadata>"` left the tree identical, kept the stamp valid, and pushed content the preflight would have refused. The key is now over `tree + commit + base`.

Reproduced the reviewer's exploit and confirmed closure:
```
check after forbidden-message amend:  exit=1
HOOK  after forbidden amend:          exit=1   (1 = push REFUSED)
```
The exploit is now a case in `Scripts/test-preflight-stamp.sh` — **7/7** self-test cases pass.

**[2] the hook was never installed** — fixed, twice over. Installed at `.git/hooks/pre-push.commitlore-chained` (confirmed the CommitLore shim invokes it), and the ship gate now has a `pre-push stamp hook installed` step that prints `NOT ENFORCED` with the install command when it is missing. CI could only ever prove the script *would* refuse; it could not prove it *is* invoked. An uninstalled gate is decoration, so the gate now says so out loud — and it already does, correctly, in the two worktrees that do not yet have this script.

First real use: the gate ran all checks, passed, stamped, and its push then failed on an unrelated non-fast-forward. The subsequent hand `git push --force-with-lease` went through the hook, which found a valid stamp and allowed it. Behaved exactly as designed.

PR #561, opened 2026-08-17
fix(#545): route the top-level delete-confirm dialog instead of failing closed

### LINKED ISSUE #545: logic_tracks.delete: 'Delete Track and Regions?' dialog not recognized, fails closed
## Summary

Deleting a track that has regions on it shows Logic's native "Delete Track and Regions?" confirmation dialog, which the MCP server doesn't dismiss — the operation ends up fail-closed instead of completing.

## Environment

- `LogicProMCP` v3.13.0
- Logic Pro 12.3

## Repro

1. Create a track with at least one region on it (audio or MIDI).
2. Call `logic_tracks` `{"command": "delete", "params": {"index": <n>, "expected_name": "<name>"}}`.
3. Logic shows the native "Delete Track and Regions?" sheet.
4. The call returns fail-closed: `"reason":"retry_exhausted"`, `"reconciled_modal_kind":"unknown_sheet"`, `"reconciled_action":"fail_closed"`.
5. Only workaround found: have a human click "Delete" on the actual dialog, then re-check `track_count` via `logic_system.health`.

## Why this isn't the same as #346/#347

I checked — `defaultDeleteTrack`'s `reconcileAfterMutation(isDeleteContext: true)` (from #347) actively dismisses the **"delete channel strips that are assigned to tracks!"** sheet (fires when the track has an assigned instrument/channel strip). It does not recognize **"Delete Track and Regions?"** (fires when the track has regions on it) — different trigger condition, different dialog title, and the modal classifier falls through to `unknown_sheet` for it. The fail-closed behavior for unrecognized sheets is clearly intentional (`"unknown sheets are never blindly dismissed to prevent data loss"`) — this is a request to add this specific, well-known sheet to the recognized set, not a request to relax that guard.

## Side note (separate, less certain observation)

In one earlier session, a `delete` call returned `"error":"operation_timeout"` but the deletion had actually completed on Logic's side by the time we checked — so on a timeout it's worth re-verifying real state (e.g. `track_count`) rather than assuming failure. Didn't manage to reproduce this one reliably enough to give clean repro steps, flagging in case it rings a bell / matches something already tracked.

MongLong0214:
Closes #545.

## The first fix was wrong and every test agreed with it

The original theory: Logic shows "Delete Track and Regions?" as an `AXSheet` whose button labels were unlocalised, so the fix added label variants. Six unit tests passed.

Logic never vends `AXSheets` at all — `-25205` for that attribute, always — and this confirmation is a top-level `AXWindow`/`AXDialog`. **The label change altered nothing on the path that actually runs.** Only a run against the application caught it, which is why this PR ships a live harness alongside the fix.

## What the bug actually cost

Not the failed delete. The dialog was classified `unknown_sheet` and **left standing**, after which every later operation in every surface returned State C until a human clicked it.

## The fix

Add `.deleteConfirm(ModalSignalRead)` and route it beside `.mandatoryNewTrack` in the top-level dialog switch. Acting on it stays gated downstream — `decide` confirms only when `isDeleteContext` is true, and preflight never acts on `.deleteConfirm` — so it cannot press Delete on a dialog the server did not ask for. It also joins the alert-acknowledger's refusal group: that path acknowledges by pressing the single button, and this dialog's buttons are a **choice**.

## Live evidence — 7/7, and the harness is built around one trap

A check that only asks *"is a dialog on screen afterwards?"* passes just as happily **when the dialog never appeared** — an empty track deletes silently and the run would certify the fix while exercising nothing. So the appearance is witnessed too, by polling System Events on a background thread for the whole delete. Both are required.

```
the-confirmation-dialog-actually-appeared   max_dialogs_seen=1  samples=25
the-dialog-was-recognised-not-refused       reconciled_modal_kind='delete_confirm'
no-dialog-is-left-on-screen                 dialog_count_after=0
the-track-was-actually-deleted              7 -> 6
the-delete-is-certified-not-abandoned       state='A'
```

Three defects in my own instruments were found and fixed getting here, each of which would have produced a false green:

- `dialogs()` read dialog **names**. Logic's auto-save recovery alert has an empty name, so the helper reported "no dialogs" with a modal alert plainly on screen. It counts now.
- `the-dialog-was-recognised-not-refused-as-unknown` was `reason != "unknown_sheet"`, which `reason=None` satisfies — it went **green on a run where the delete never executed**. Now gated on the dialog having appeared.
- The visual region sampled the leftmost fifth of the window, which is the Library browser when it is open, not the track rail. It failed while the delete had plainly worked.

Two refusals along the way were the **product being correct**, not obstacles: `track.delete` refuses a bare ordinal ("the index is an unproven ordinal"), and refuses an `expected_name` matching two live tracks ("same-named tracks can swap and stay self-consistent"). The harness now picks a uniquely-named track.

Ship gate PASS — 3768 tests, live evidence bound to head.


MongLong0214:
Updated @ 4a0d1918 — three review rounds landed since the last green. **The earlier CLEAN was for `4aafc8b7`, which still contained the hand-translations and the vacuous tests**; the PR was briefly drafted so that stale green could not read as merge-ready.

## Peer review — hand-translated locale variants removed

The set carried `삭제` and `削除`, and its own rationale gave them away: *"both measured live on 12.3"* covered only the two English forms, then the KO/JA were asserted as *"equivalents"*. They were translated by hand.

`AXLocalePolicy`'s header forbids exactly that, and says why — a hand translation was already wrong in this app (New is `신규`, not the `새로 만들기` a translator reaches for). Eleven other sets carry empty variants for the same reason. Note what is **not** sufficient: `markerListDeleteMenuItem` does carry a live-confirmed `삭제`/`削除`, so these are real Logic strings *somewhere*. That raises the prior; it is not a measurement of this button.

The gap is now stated in the rationale and pinned by tests that assert the variants do **not** match, so restoring them requires naming a measurement.

## Adversarial review — two BLOCKERs, both real

**1. The unit tests could not fail if the fix were removed.** Every fixture built `AXSheets`, so `readModalSignals(runtime:)` returned through the sheet branch and `topLevelDialogRead` — the code path under test — was never entered. Measured:

```
delete the .deleteConfirm case, then run the suite:
  ✔ a bare English "Delete" ... classifies as deleteConfirm
  ✔ the channel-strip sheet's full canonical label ...
  ✔ unmeasured locale labels do NOT classify
  ✔ a deleteConfirm sheet outside a delete context ...
  ✘ a top-level modal delete-confirm is routed and pressed   <- only the new one
```

**This is the same defect that let #545's FIRST fix pass six tests while changing nothing** — re-enacted in the suite meant to prove the second one. `Issue538` already had the correct shape. Added a fixture that builds what Logic presents: top-level `AXWindow`, `AXModal == true`, `AXDialog` subrole, no sheets. Plus a test pinning that it is **not** pressed outside a delete context.

**2. Every livekit harness exited 0 on failure.** `sys.exit(0 if out["passed"] else 1)` reads a COUNT as a boolean — `5` of 7 is truthy, so only a run where *nothing* passed could fail. Fixed at source with `evidence.is_clean()`, which also requires visuals, capture settling, restorations and provenance to be clean.

Two MAJORs of my own making, same class:

- **A restoration record was hardcoded `True`** — in the record type whose entire purpose is proving cleanup. Now compares the bundle's ProjectData mtime across the run; an unreadable stat is a failed measurement, not a clean restore.
- **The recognition check asserted an absence** (`reason != "unknown_sheet"`, satisfied by `nil`) — it went green on a run where the delete never executed. Now asserts `reconciled_modal_kind == "delete_confirm"`.

Also documented `ModalSignals.sheetPresent`: the element-scoped reader sets it unconditionally, so it means *"a modal container was read here"*, not *"an AXSheet exists"*. Reading that name literally is what produced the original wrong root cause.

Live 7/7 at this head, restoration measured. Ship gate PASS — 3783 tests.

PR #593, opened 2026-08-18
fix(#590): the project chooser is not an open document

### LINKED ISSUE #590: project.new cannot run from the state Logic lands in at launch: the project chooser is counted as an open document
## Reproduction, measured today on Logic Pro 12.3 (both English and Korean)

Launch Logic with no document. It shows **Choose a Project**. Then:

```
logic_project.new
-> {"error":"unsupported_state","failure_stage":"precondition_open_document",
    "observed_window_count":1,
    "hint":"project.new requires Logic to have no open document; 1 window(s) are open. …",
    "recovery_action":"Close the open project with logic_project(\"close\", {confirmed: true}) …"}
```

The one window is the chooser. There is no open project.

I hit a second refusal before this one, and it is part of the same first-contact story: a fresh
launch also raises a single-button informational alert —

```
[] subrole=AXDialog buttons=[OK]
   "The previously selected audio interface is not available. The built-in audio input and output
    of this Mac mini will be used for this session instead."
```

— and `project.new` refuses on it with `failure_stage: preflight_blocking_dialog`. That one is a
defensible fail-closed, though this repository already carries a reconciler for single-button
top-level informational alerts (#453); the preflight refuses before it runs.

Dismiss the alert and the chooser refusal above is what remains.

## The cause is one line, and the classifier it needs already exists

`AccessibilityChannel+Project.swift:30` gates on `blankApplicationCanRevealChooser(windowCount:)`,
which counts raw windows. `AXLogicProElements.isProjectPickerWindow` already exists — `getTrackHeaders`
uses it at `AXLogicProElements+Tracks.swift:36` for exactly this reason — and the precondition does
not.

The reasoning in the comment above that guard is sound: with a document already open, a newly
created project's window cannot be told apart from the ones already on screen. **A chooser is not a
document**, and it disappears when the new project appears, so it does not create the ambiguity the
guard exists to prevent.

## Confirmed by removing the chooser and nothing else

```
click the chooser window's own AX close button   (its close/zoom/minimize ARE exposed)
logic_project.new
-> {"state":"B","success":true,"phase":"created_project_window_observed",
    "mandatory_track_created":true,"method":"accessibility"}
   window: "Untitled 56 - Tracks"
```

Same Logic, same launch, one window removed.

## The recovery_action is also wrong for this state

It says to run `logic_project("close", {confirmed: true})`. There is no project to close — the
recovery that actually worked was closing the chooser window itself. A caller following the hint from
this state is sent somewhere else.

## Why this is worth more than its size

This is the state every first-time user is in. #516 is an outside integrator who ended their Logic
Pro integration over `project.new`; #519 records that they were on a machine configured differently
from the maintainer's. This refusal needs no unusual configuration at all — only a Logic that was
launched and not yet given a project.

## The chooser is an Accessibility wall, so the caller cannot work around it either

```
Choose a Project  buttons: [close button][zoom button][minimize button]
                  rows: (none)   static texts: "Choose a Project"
```

Nothing in its content is exposed. A caller that wanted to pick a template from there cannot, and a
caller that wanted to get past it has only the window's own close button — which is not something the
current hint mentions.

## Suggested shape

Exclude chooser windows from the count using the existing `isProjectPickerWindow` classifier, and
make the recovery action name the chooser when the chooser is what is in the way. Both halves want a
live proof from a cold launch, since that is the only state where this reproduces.

MongLong0214:
Closes #590

## The defect

A freshly launched Logic shows **Choose a Project**. `project.new`'s open-document precondition counted
raw AX windows, so that chooser *was* an open document:

```
{"error":"unsupported_state","failure_stage":"precondition_open_document",
 "observed_window_count":1,
 "hint":"project.new requires Logic to have no open document; 1 window(s) are open. …",
 "recovery_action":"Close the open project with logic_project(\"close\", {confirmed: true}) …"}
```

There was no open document. This is the state **every first-time caller is in**, and it needs no
unusual configuration — only a Logic that was launched and not yet given a project. Reproduced in
English and in Korean, so it is not a locale defect.

The recovery hint made it worse: from this state there is nothing to close.

## What did not change

The precondition's reasoning, which is right: with a real document open, a newly created project's
window cannot be told apart from the ones already on screen.

A chooser does not create that ambiguity. It is titled, it is not a project, and **driving File ▸ New
with it still on screen was measured to create the project anyway**. So the count now excludes chooser
windows using the classifier the codebase already had — `isProjectPickerWindow`, which
`getTrackHeaders` uses for exactly this reason and which already carries both the English and the
Korean window titles. The envelope now reports both numbers, so a reader can see a chooser on screen
was not the reason for a refusal.

## Measured before and after, on the same host

With the fix — 6 checks, 6 passed, 2 mutation-backed, visual 1/1, restoration verified.

With `documentWindowCount` restored to raw windows and the release binary rebuilt:

```
failure_stage 'precondition_open_document'   all_windows 1   documents 1
before ['Choose a Project']   after ['Choose a Project']
```

No project created. That is the defect, reproduced live from the same script.

The harness quits Logic and waits for the chooser before it begins — one that started from an open
project would exercise a different branch and pass while measuring nothing.

## Three things the run had to handle, recorded rather than hidden

**A fresh launch raises Logic's own single-button audio-interface alert**, and `project.new` refuses on
it at a different stage (`preflight_blocking_dialog`). That refusal is defensible and untouched here,
so the alert is acknowledged before the test and the run says which one it was.

**The alert does not arrive with the chooser.** An earlier version looked once, saw nothing, declined —
and the operation then refused on the alert that arrived a moment later, failing the run for a reason
that was not the one under test. It now waits. The gate is the *button count*, not the text: the same
dialog reported `["OK"]` to the product while `name of every button` came back empty through System
Events, so a name-based count would have declined on a dialog that was plainly there.

**Quitting can be refused by a sheet, silently.** A freshly created project leaves its "New Track"
chooser up and `quit` then does nothing while the process stays alive. The shutdown escapes a sheet
once and asks again, and the precondition judges whether Logic actually stopped rather than whether a
quit was sent.

## Gate

`lpm-ship.sh` PASS — 3825 tests, preflight clean, `Package.resolved` untouched, branch contains

### DIFF
--- a/Scripts/livekit/live_590_project_new_from_cold_launch.py
+++ b/Scripts/livekit/live_590_project_new_from_cold_launch.py
@@ -1,0 +1,236 @@
+#!/usr/bin/env python3
+"""Live proof that `project.new` works from the state Logic lands in at launch.
+
+Usage:  LPM_EVIDENCE_ROOT=/abs/path/outside/repo \
+        python3 live_590_project_new_from_cold_launch.py <worktree> <full-40-char-head-sha>
+
+WHAT WAS WRONG
+--------------
+A freshly launched Logic shows "Choose a Project". `project.new`'s open-document precondition counted
+raw AX windows, so that chooser was an open document and the operation refused:
+
+    {"error":"unsupported_state","failure_stage":"precondition_open_document",
+     "observed_window_count":1,
+     "hint":"project.new requires Logic to have no open document; 1 window(s) are open. …"}
+
+There was no open document. This is the state every first-time caller is in, and it needs no unusual
+configuration — only a Logic that was launched and not yet given a project. Measured in English and
+in Korean; it is not a locale defect.
+
+The precondition's reasoning is untouched and still right: with a real document open, a newly created
+project's window cannot be told apart from the ones already on screen. A chooser does not create that
+ambiguity, and driving File > New with the chooser still on screen was measured to create the project
+anyway. So the count now excludes chooser windows, using the classifier `getTrackHeaders` already
+uses for the same purpose.
+
+WHY THIS RUN RELAUNCHES LOGIC
+-----------------------------
+The defect exists only at a cold launch. A harness that started from an already-open project would
+exercise a different branch and pass while measuring nothing, so this one quits Logic and waits for
+the chooser before it begins. It says so here rather than leaving a reader to infer why Logic
+restarted.
+
+A fresh launch also raises Logic's own single-button audio-interface alert on this host. That is
+acknowledged before the test — it is Logic's, informational, and its own preflight refusal
+(`preflight_blocking_dialog`) is a separate and defensible one that would otherwise mask the branch
+under test.
+"""
+
+import json
+import os
+import subprocess
+import sys
+import time
+
+sys.path.insert(0, os.path.dirname(os.path.abspath(__file__)))
+import evidence as E  # noqa: E402
+
+WT = sys.argv[1] if len(sys.argv) > 1 else ""
+HEAD = sys.argv[2] if len(sys.argv) > 2 else ""
+if not WT or not HEAD:
+    sys.exit(__doc__)
+
+E.REPO = WT
+E.BIN = f"{WT}/.build/release/LogicProMCP"
+missing = E.have_tools()
+if missing:
+    sys.exit(f"cannot run: missing {missing}")
+
+ev = E.Evidence(HEAD, os.environ["LPM_EVIDENCE_ROOT"])
+LAUNCH_TIMEOUT = 90.0
+CHOOSER_TITLES = ("Choose a Project", "프로젝트 선택")
+
+
+def osa(script):
+    r = subprocess.run(["osascript", "-e", script], capture_output=True, text=True)
+    return (r.stdout or "").strip()
+
+
+def window_titles():
+    raw = osa('tell application "System Events" to tell process "Logic Pro" to '
+              'return name of every window')
+    return [t.strip() for t in raw.split(",")] if raw else []
+
+
+def logic_running():
+    return osa('tell application "System Events" to return (count of (every process whose '
+               'name is "Logic Pro"))') == "1"
+
+
+def dismiss_single_button_alert():
+    """Acknowledge Logic's own informational alert, but ONLY when it has exactly one button.
+
+    The count is the gate, not the text: measured, the same audio-interface alert read its static
+    texts on one launch and returned nothing on the next, and an earlier version of this function
+    keyed the click on the text being readable — so it silently declined to dismiss anything and the
+    run failed at a later stage for a reason that was not the one under test.
+
+    One button means there is no choice to make and nothing this run could decide on the operator's
+    behalf. Two or more, and it declines and says so, which is the same rule the product's own modal
+    reconciler follows.
+    """
+    # The alert does not appear the instant the chooser does. An earlier version looked once, saw
+    # nothing, and declined — and the operation then refused on the alert that arrived a moment
+    # later, failing the run for a reason that was not the one under test. So this waits for it.
+    #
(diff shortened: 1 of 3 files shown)


MongLong0214:
Updated with two commits since the description above.

## The look-alike project — the direction this fix opened

A blind review and my own check landed on the same hole. `isProjectPickerWindow` matches the window
title by **containment**, so a project the user names `Choose a Project` produces the arrange window
`Choose a Project - Tracks` — which contains the phrase. On a title-only rule that window stops being
counted and `project.new` proceeds with a genuine document open.

That is worse than the bug being fixed. The old defect refused too often; accepting too often creates
the ambiguity the precondition exists to prevent.

The title is now paired with a structural signal, measured on Logic Pro 12.3:

```
[Untitled 56 - Tracks]  AXDocument = file:///…/Untitled%2056.logicx/
[Choose a Project]      AXDocument = missing value
```

A window leaves the document count only when **both** signals say chooser. Every uncertain case —
unreadable title, unreadable `AXDocument`, no match — counts as a document, so failing to identify the
chooser costs a refusal rather than an ambiguous creation. The new test is the look-alike project
rather than the chooser, because the chooser was already covered and the look-alike is what a
title-only rule gets wrong; restoring title-only reddens that test and nothing else.

## The harness was assuming the state it needed

It assumed a cold launch lands on the chooser. Logic reopens the document that was open when it quit,
so a relaunch after quitting with a project open came back straight to `Untitled 56 - Tracks` and the
branch under test was never reached — the run failed honestly and told me why.

The shutdown now **establishes** the state: document windows are closed before the quit. A save prompt
is three buttons, a real choice, so it is answered only for a document the run can identify as
untitled scratch. A named project is the operator's — the run cancels and stops, and the precondition
reports that as the reason.

## Gate

`lpm-ship.sh` PASS at `d5e8035f` — 3826 tests, preflight clean, `Package.resolved` untouched, branch
contains `origin/main`, live evidence bound to this head: 6 checks, 6 passed, 2 mutation-backed,
visual 1/1, restoration verified.

MongLong0214:
Updated after #588 and #589 merged. Branch rebuilt on `origin/main` and re-gated from a single clean
run — an earlier attempt had two gate processes overlapping in the same worktree, which produced a
log with both `SHIP GATE FAIL` and a successful push in it. That was my error, not a flaky gate.

```
full suite    ok — 3828 tests
live evidence 6 checks · 6 passed · 2 mutation-backed · visual 1/1 · restoration verified
SHIP GATE PASS · pushed @ 42e02604
```

## A second commit is on this branch now

`fix(#590): identifying the chooser by title alone can hide a real document` closes the direction the
first commit opened, and it is the more important half.

`isProjectPickerWindow` matches the window title by **containment**. A project the user names
"Choose a Project" produces the arrange window `Choose a Project - Tracks`, which contains the
phrase — so on a title-only rule that window stops being counted and `project.new` proceeds with a
genuine document open. The original defect refused too often; accepting too often is worse.

The title is now paired with a structural signal, measured on Logic Pro 12.3:

```
[Untitled 56 - Tracks]  AXDocument = file:///…/Untitled%2056.logicx/
[Choose a Project]      AXDocument = missing value
```

A window leaves the document count only when **both** signals say chooser. Every uncertain case —
unreadable title, unreadable `AXDocument`, non-matching title — counts as a document, so failing to
identify the chooser costs a refusal rather than an ambiguous creation.

The new test is the look-alike project rather than the chooser: the chooser was already covered, and
the look-alike is the case a title-only rule gets wrong. Restoring title-only classification reddens
that test and nothing else.

## One note on the gate output

`live evidence for this head — n/a — no Sources/ change in this branch` appears because this head is
a merge commit, whose own diff carries no `Sources/` change. The evidence was produced anyway and is
clean at `3907fcc7`; the pushed SHA differs because the pre-push stamp hook rewrites the commit.

Issue #575, opened 2026-08-17
Seven region.* operations are implemented and routed but reachable from no tool, and absent from the OperationRegistry


MongLong0214:
## The census

`RoutingTable` routes eight `region.*` operations. Only one is reachable from a tool:

| operation | channel implementation | dispatcher | OperationRegistry |
|---|---|---|---|
| `region.get_regions` | yes | `logic_project.get_regions` | — |
| `region.select` | yes | **none** | **absent** |
| `region.select_last` | yes | **none** | **absent** |
| `region.move_to_playhead` | yes | **none** | **absent** |
| `region.move` | yes | **none** | **absent** |
| `region.resize` | yes | **none** | **absent** |
| `region.loop` | yes | **none** | **absent** |
| `region.set_name` | yes | **none** | **absent** |

Verified against the running server, not only by grep — no tool accepts them under any spelling:

```
logic_project.select_region   -> invalid_params  "Command 'select_region' is not registered for ..."
logic_edit.move_to_playhead   -> invalid_params
logic_project.region_select   -> invalid_params
```

`AccessibilityChannel+Regions.swift` implements all seven, including the menu-drive work that #519
converted to `AXLocalePolicy` LabelSets (`Edit > Move > To Playhead`). That code runs for nobody.

## Two consequences worth separating

**No MCP client can move, resize, loop, rename or select a region.** The region surface is
read-only in practice, and `logic://regions` plus `logic_project.get_regions` are the whole of it.

**The exhaustiveness gates cannot see them.** They are absent from `OperationRegistry`, so
`OperationRegistryCoverageTests`, `IndexBindingCensusTests` and `SemanticOracleTests` — which exist
to make an operation's contract impossible to forget — have nothing to check. An operation that is
routed but unregistered sits outside the very machinery built to keep operations honest.

## How it surfaced

#519's acceptance requires "a region operation succeeds on a Korean Logic, verified live". There is
no region operation to drive. That acceptance criterion cannot be met as written — not because the
locale work is incomplete, but because the surface it names is not exposed.

## Not proposing the fix here

Whether these should be exposed, and with what index-binding and verification contract, is an
ADR-shaped decision (#302 already carves out region selection as its own piece of work, and
`AccessibilityChannel+Regions.swift:479` shows `region.select_last`'s State-A story is unsettled).
Filing the census so the gap is visible and #519's acceptance can be restated against what exists.

MongLong0214:
[...]

PR #597, opened 2026-08-18
fix(#592): retire six more table rows that stood in front of no implementation

### LINKED ISSUE #592: Six channel-table rows stand in front of no implementation (the non-region half of the #575 census)
#587 retired five `region.*` rows whose only behaviour was a refusal string. I said at the time that
"eight more operations share this shape" and deliberately left them alone as out of scope.

**Two of the eight do not share it.** Correcting that before it becomes the premise of someone else's
work. Audited against the source at `ed58020c`, chain by chain:

| operation | channel chain | who implements it | verdict |
|---|---|---|---|
| `mixer.set_send` | `[.mcu]` | `MCUChannel.swift:259` | **implemented** — the accessibility arm's refusal is unreachable, since AX is not in its chain |
| `automation.set_mode` | `[.mcu, .midiKeyCommands, .cgEvent]` | `MIDIKeyCommandsChannel.swift:104` (key command 84) | **implemented** — accessibility is not in its chain either |
| `mixer.set_input` | `[.accessibility]` | nobody | row in front of a refusal |
| `mixer.set_output` | `[.accessibility]` | nobody | row in front of a refusal |
| `mixer.toggle_eq` | `[.mcu, .accessibility]` | nobody — MCU has no arm for it | row in front of a refusal |
| `mixer.reset_strip` | `[.mcu, .accessibility]` | nobody — MCU has no arm for it | row in front of a refusal |
| `plugin.list` | `[.accessibility]` | nobody | row in front of a refusal |
| `automation.get_mode` | `[.accessibility]` | nobody | row in front of a refusal |

So it is **six**, not eight. None of the six is registered for any tool, so no caller can reach any
of them today.

## Why the rows still matter

A row in the channel table is a statement that an operation is real and a declaration of which
surfaces may carry it. For these six it declares a channel order for a refusal — and the sentence it
refuses with reads as a promise ("not yet implemented via AX") rather than as an answer. That is the
same reasoning #587 applied to the region five.

Two of them are worse than merely empty: `mixer.toggle_eq` and `mixer.reset_strip` name `.mcu` first,
and `MCUChannel` has no arm for either. The table promises a fallback that does not exist.

## Two rows to leave alone, and why they are different

`mixer.set_send` and `automation.set_mode` should keep their rows — they work. What they have is a
smaller, separate problem: an `AccessibilityChannel` arm that returns "not yet implemented via AX"
for an operation whose chain never routes to accessibility. That code is unreachable, and a reader
who greps for the operation finds a refusal that is not what the operation does. Removing those two
arms is a tidy-up, not a contract change.

## What a fix wants

The same shape as #587: remove the six rows and their refusal arms, remove the two unreachable arms,
and prove live that nothing reachable moved — the reachable read-only surface across every tool, plus
the specific neighbours that share a prefix with each removed row (`mixer.set_volume`,
`mixer.set_pan`, `mixer.set_master_volume`, `plugin.get_inventory`, `tracks.set_automation`).

The neighbour check is the load-bearing one. A removal's risk is not in what it deletes but in what
it takes with it, and `plugin.list` sits next to a `plugin.*` family that is very much alive.

MongLong0214:
Closes #592. Same shape as #587, which retired five `region.*` rows; these are the rest.

| operation | chain | who implements it |
|---|---|---|
| `mixer.set_input` | `[.accessibility]` | nobody |
| `mixer.set_output` | `[.accessibility]` | nobody |
| `mixer.toggle_eq` | `[.mcu, .accessibility]` | **nobody — MCU has no arm either** |
| `mixer.reset_strip` | `[.mcu, .accessibility]` | **nobody — MCU has no arm either** |
| `plugin.list` | `[.accessibility]` | nobody |
| `automation.get_mode` | `[.accessibility]` | nobody |

A row in the channel table states that an operation is real and declares which surfaces may carry it.
For these six it declared a channel order for a **refusal**, and the sentence it refused with reads as
a promise rather than an answer. The two that name `.mcu` first are worse than empty: the table
advertised a fallback that does not exist.

## Two that look identical and are not — correcting my own claim

I wrote on #575 that *eight* operations shared this shape. Two are implemented:

```
mixer.set_send        [.mcu]                              MCUChannel.swift:259
automation.set_mode   [.mcu, .midiKeyCommands, .cgEvent]  key command 84
```

Neither chain contains `.accessibility`, so the refusal arm I had been reading is **unreachable** for
both. I was looking at a refusal the operation never reaches and calling the operation a stub. Their
rows stay; the dead arms go, because a reader grepping for the operation should not find a sentence
that is not what it does.

## What the live run shows

None of the six was reachable from a tool, so no live call can demonstrate a behaviour change *in
them*. The load-bearing evidence is the **neighbours** — a removal's risk is not in what it deletes
but in what it takes with it, and `plugin.list` sat beside a `plugin.*` family that is very much
alive.

```
6 checks · 6 passed · 3 mutation-backed · negative visual 1/1
```

Neighbours are driven with a parameter each one **rejects**: proving one survives does not require
moving the user's mixer, and the discriminator is the *hint*, because a live command and a retired one
both answer `invalid_params`. A live one says `Unknown parameters: … Allowed parameters: …` — it
validated input. A retired one says `is not registered for MCP tool`.

## Three assumptions the harness gave up, all because the product was clearer than the check

**`plugin.get_inventory` requires parameters**, so calling it bare answers `invalid_params` for a
reason unrelated to this removal. Probed the same way as the mixer neighbours now.

**The retired mixer names answer `command_not_exposed`**, not `invalid_params` — the mixer dispatcher
carries an explicit not-exposed list and says so. Demanding one shape would have failed the run for
the product being *more* specific than the check expected, so it accepts either and records which.

**The visual band was a constant** measured on a window that has since changed shape, and it now
landed on the Step Sequencer, which animates on its own — so a check that exists to show nothing
moved reported movement in a part of the screen the run never touched. It is read from the rail's own
AX frame at run time now: `366,707 235x1182` against this window, nowhere near the `603,162 325x406`
it replaces. Two AppleScript hazards fell out of that — `by` is a reserved word, and
`first window whose name is "…"` answered *invalid index* for a window `name of every window` had just
listed, so the window is chosen by elimination.

## Gate

`lpm-ship.sh` PASS — 3828 tests, preflight clean, `Package.resolved` untouched, branch contains

### DIFF
--- a/Scripts/livekit/evidence.py
+++ b/Scripts/livekit/evidence.py
@@ -14,6 +14,10 @@ What it refuses to let you record:
   resulting image is not what the user sees. Every capture records which display it fell wholly within.
 - **A whole-window diff as a visual assertion.** A full-window comparison changes when the clock ticks.
   `visual()` requires an explicit region and states what it expected to happen there.
+- **A region measured once and carried as a constant.** A rectangle taken on one window shape silently
+  watches something else after a resize — measured, a stale band landed on the Step Sequencer and made
+  a negative assertion report movement in a part of the screen the run never touched. Derive the band
+  at run time; `track_header_band()` does it for the arrange rail.
 - **A check that cannot fail.** `check()` requires you to name the mutation you applied to the product
   that flipped it. A check nobody has seen fail is not evidence.
 - **A cached read presented as live.** `provenance()` records the source and age of any value that did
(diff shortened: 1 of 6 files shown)


MongLong0214:
Two commits added since the PR was opened, both from the same defect and both worth reading as one
thing.

## The visual band was a constant, and it had drifted onto something that moves

The negative visual assertion failed on a run in which nothing was written. The band —
`(603, 162, 325, 406)` — was measured once on a 1920x1050 window with the Library closed. The window
has since changed shape, and that rectangle now lands on the **Step Sequencer**, which animates on its
own. So a check whose whole job is to show that nothing moved reported movement, in a part of the
screen the run never touched.

Derived from the rail's own AX frame at run time it is `366,707 235x1182`. Nowhere near the constant.

## The same defect is latent in twenty-one other harnesses

Every `settle_region` in `Scripts/livekit/` is a hardcoded rectangle. Rather than fix the one that
broke, the derivation moved into the recorder as `evidence.track_header_band()`, and the module
docstring gains the rule beside the other things it refuses to let an operator record. The #592
harness calls the shared helper now, so there is one implementation rather than three.

The other twenty-one keep their constants. Rewriting runs that are not being executed would be a
large unreviewed change; the next harness written gets this for free and is told why.

## Three AppleScript facts fell out of building it, all measured

- **`by` is a reserved word** — a variable named that fails to *parse*, not to run.
- **`first window whose name is "…"` answered "invalid index"** for a window `name of every window`
  had listed a moment earlier, with a plain-ASCII title matching exactly.
- **Excluding the chooser by title was not enough.** A run that overlapped an `Import` dialog picked
  that dialog as the project window and got nothing. The window is now chosen by *what it contains* —
  every window is tried until one yields a track rail.

## Gate

`lpm-ship.sh` PASS at `86d26cb7` — 3828 tests, live evidence 6/6 with 3 mutation-backed checks and the
negative visual passing.

MongLong0214:
Superseded, not rebased — the conflict is not worth resolving because the change is already in.

Re-counted on current main: all six rows this PR retires are gone, and so is every dead switch arm it removes. Main also carries a later, more accurate version of this branch's harness prose — the correction that `mixer.set_send` and `automation.set_mode` lost their accessibility arms *without* losing their rows.

The one thing here that main lacked is `track_header_band()`. That is recovered on a fresh branch from main in #611, with the docstring's unverifiable measurement replaced by one taken today. The four conflicting files need no resolution.

PR #598, opened 2026-08-18
docs(#369): the export surface CAN do per-track stems — ticket, not a scope decision

MongLong0214:
Part of #369. Docs only — `docs/tickets/issue-369-per-track-stem-export/`.

## The roadmap's expectation for this issue was wrong, and it is measured wrong

Wave 6 says #369 should close with a measurement:

> the export panel's per-track filename fields report `settable=true` and do not accept writes.
> **If that holds**, per-track stems cannot be driven from this surface and the issue closes with the
> measurement.

Driven live on Logic Pro 12.3, Accessibility only, no coordinates:

```
~/Music/Logic/Studio Grand_1.aif    394,922 bytes

logic_audio.analyze_file
  2 ch · 48000 Hz · 2.048 s · peak −13.8 dBFS · 0.744 s non-silent · status: pass
```

Real audio, from the surface the plan said could not make it. So #369 is **implementable work**, not a
scope decision, and this PR turns the measurements into the ticket that implements it.

## What the earlier reading missed

`One File per Track` is a popup **on** the export panel. The previous measurement saw 148 settable
text fields and concluded the per-track surface sat behind the panel — that was the file-browser
half. The option half was on the same panel the whole time:

```
popups      Logic · Trim Silence at File End · AIFF · One File per Track · 16-bit · Overload Protection Only
checkboxes  Bypass Effect Plug-ins · Include Audio Tail · Include Volume/Pan Automation ·
            Add resulting files to Project Browser · Include Tempo Information
```

## Four things a naive drive gets wrong, all measured

**Menu enablement read from a closed menu is meaningless.** The same items read `enabled=false`
closed and `enabled=true` open — macOS validates on open. I spent a cycle believing export was
unavailable.

**The leaf title is rewritten by Logic**: `Tracks as Audio Files…` → `1 Track as Audio File…` with one
track selected. Name resolution must survive a title the application *edits*, not only one it
localizes.

**Typing a destination path dismisses the panel** — reproduced twice, nothing written. The folder has
to be picked as a browser element (`AXOutline`/`AXRow`) and the destination popup re-read to confirm
it changed *before* Export.

**The `Logic Pro` window after Export is the progress dialog, not an error.** It is gone by the time a
follow-up read arrives; a probe that sees a window appear and vanish around a write will file it as an
unread failure. Its disappearance is the completion signal — the Export click returning is not.

## What the ticket is blocked on, and why it is not decided here

Logic writes one file per populated track, so a run can partly succeed. If three of four verify and
one is silence, the run produced something real and something useless.

The two coherent answers are State C for the whole run with every file named, or State A per file
with a run-level summary that is not a State. This project's contract has no shape for "mostly
worked", and deciding that inside an implementation ticket is how a contract gets set by whoever
happened to be writing code that week.

Everything else in the ticket is measured and mechanical.

## Gate

`lpm-ship.sh` PASS — preflight clean, `Package.resolved` untouched, branch contains `origin/main`.

### DIFF
--- a/docs/tickets/issue-369-per-track-stem-export/STATUS.md
+++ b/docs/tickets/issue-369-per-track-stem-export/STATUS.md
@@ -54,3 +54,16 @@ contract: what an artifact plan promises when the names arrive late.
 
 Two corrections in two revisions, both from the same habit — answering about a file without opening
 it. Recorded rather than tidied away, because the shape of the mistake is the useful part.
+
+## Third correction, and the last one
+
+"T1 stands alone and is buildable" was true of the code and false of shipping it. The public export
+surface is `project.export_plan` and `project.export_run` and nothing else, so a standalone stem drive
+would have to be a new public operation — a surface decision — or it would sit unrouted, which is the
+shape #587 and #592 retired eleven rows for on the same day this ticket was written.
+
+The order is therefore fixed rather than parallel: answer what an artifact plan promises when the
+names arrive late, wire `export_run artifacts:[stem]`, and let T1's drive land inside it.
+
+Three corrections in three revisions. The first two came from answering about a file without opening
+it. This one came from treating "buildable" and "shippable" as the same word.
--- a/docs/tickets/issue-369-per-track-stem-export/STATUS.md
+++ b/docs/tickets/issue-369-per-track-stem-export/STATUS.md
@@ -37,3 +37,20 @@ flows through.
 
 The claim was an absence asserted without reading the file the answer was in. Corrected in place
 rather than left standing, because a ticket that says "blocked" is read as a reason not to start.
+
+## Second correction
+
+The first correction removed the "blocked" flag on the strength of `ProjectExportExecutor` carrying
+per-artifact State. That finding is right and it stands.
+
+It was also incomplete. `ProjectExportPlanner` computes one artifact as one known path
+(`<project>-<kind>.wav`) with `exists`, the collision policy and containment all resolved at plan
+time — and a stem run produces N files whose names Logic assigns after the fact. So `export_plan`,
+which is a dry run whose job is to say what will be written, cannot enumerate a stem run.
+
+T1 is therefore unblocked as an AX drive and can be built and proven on its own. Wiring it into
+`export_run artifacts:[stem]` waits on one question that is a property of the published dry-run
+contract: what an artifact plan promises when the names arrive late.
+
+Two corrections in two revisions, both from the same habit — answering about a file without opening
+it. Recorded rather than tidied away, because the shape of the mistake is the useful part.
--- a/docs/tickets/issue-369-per-track-stem-export/STATUS.md
+++ b/docs/tickets/issue-369-per-track-stem-export/STATUS.md
@@ -2,13 +2,13 @@
 
 **Issue**: #369
 **Size**: M
-**Current Phase**: measured, ticket written, blocked on one contract decision
+**Current Phase**: measured, ticket written, ready to implement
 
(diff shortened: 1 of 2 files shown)


MongLong0214:
## Correction: the ticket was not blocked, and I was wrong the way this repo hunts for

The first revision declared T1 blocked on a contract decision — *"what State does a partially
successful stem run report"* — on the claim that this project has no shape for "mostly worked".

**It has one, and it is in the executor `export_run artifacts:[stem]` already flows through.**
`ProjectExportExecutor`:

```
Honest Contract per artifact:
  State A — bounce fired AND on-disk verification PASSED
  State B — bounce fired but the artifact could not be verified — success uncertain
  State C — a hard failure (open/identity/route error, or the plan was degraded for this artifact)
```

A stem run is exactly that list, one artifact per populated track, each verified on disk with
`AudioAnalyzer` — which is what AC-5 already asked for.

So the blocker was not a gap in the contract. It was **an absence I asserted without aiming anything
at the place the answer lives** — the defect class I spent today finding in the product, in the ship
gate's own tooling, and in three of my own harnesses. Doing it in a ticket that says "blocked", which
is read as a reason not to start, is the version that costs the most.

Corrected in place rather than left standing with a footnote.

## What the correction adds

Two behaviours the stem run should **inherit rather than reinvent**, both already in that executor:

- an artifact the plan flagged `would_overwrite` **fails closed** instead of bouncing over an existing
  file under `fail_if_exists`
- already-present-and-verified artifacts are **skipped**, which is what makes `export_resume`
  idempotent — a stem run gets resume for free if it produces its file list the same way

And the one thing that genuinely is new: **Logic names the files, not the plan.** The panel writes
`<track name>_1.aif`, so the executor cannot pre-compute the paths it polls for. It has to enumerate
the destination after the progress window closes and bind each file to a track by name. That is
implementation, not a contract question.

T1 is now `Depends On: None`.

MongLong0214:
## Second correction — and the same habit produced it

The correction above removed the "blocked" flag because `ProjectExportExecutor` carries per-artifact
State. **That finding is right and it stands.** It was also incomplete, because I answered a question
about `ProjectExportPlanner` without opening it either.

```swift
let url = outputRoot.appendingPathComponent("\(safeProject)-\(kind).wav").standardizedFileURL
let exists = existingPath != nil
```

One artifact is **one known path**, computed at plan time, with `exists`, the collision policy and the
containment check all resolved before anything runs. A stem run breaks every one of those:

- N files, not one
- **Logic names them**, not the plan — `<track name>_1.aif`
- `.aif`, not the `.wav` the path model assumes
- `would_overwrite` cannot be evaluated for names that do not exist yet

`export_plan` is a **dry run** whose job is to tell the caller what will be written. For stems it
cannot — and that is a property of a published contract, not an implementation detail. I called it
one in the previous comment.

## What the ticket looks like now

**T1 stands alone and is buildable.** It is the AX drive: open the panel, choose a destination as an
element, confirm the popup changed, set `One File per Track`, press Export, wait for the progress
window to go, enumerate what appeared, verify each file with `logic_audio.analyze_file`. None of that
touches the planner, and all of it is measured.

**Wiring it into `export_run artifacts:[stem]` waits on one question**: what an artifact plan promises
when the names arrive late, and what `fail_if_exists` and `export_resume`'s idempotency mean then.
That is a separate ticket.

## On the two corrections

Both came from the same habit — answering about a file without opening it — which is the defect class
this repository spends most of its guards on and the one I spent the day finding elsewhere. They are
left in the ticket's record rather than tidied away, because the shape of the mistake is the part
worth keeping.

Issue #604, opened 2026-08-18
project.save_as cannot recognise Logic's Save panel whenever its file browser is expanded, and leaves the panel open


MongLong0214:
## Measured on Logic Pro 12.3

`project.save_as` returned:

```
{"error":"channels_exhausted","state":"C",
 "hint":"Exact Save As dialog did not appear within 3 seconds"}
```

The panel **did** appear. Timed from the menu click:

```
Save window appeared after: 0.75 s
```

So the 3-second budget is not the problem. `exactSaveAsDialog` never classifies the panel that is
sitting on screen.

## Exactly one clause of the classifier fails

`saveAsDialogSelectionIsUnambiguous` requires all of:

```swift
exactContainer && windowTitle == "Save"
  && filenameFieldCount == 1 && saveButtonCount == 1 && saveEnabled
  && cancelButtonCount == 1 && packageRadioCount == 1 && folderRadioCount == 1
```

Measured against the live panel:

| clause | required | observed |
|---|---|---|
| window title | `"Save"` | `Save` ✅ |
| save button | 1 | 1 ✅ |
| cancel button | 1 | 1 ✅ |
| Package radio | 1 | 1 ✅ |
| Folder radio | 1 | 1 ✅ |
| **filename fields** | **1** | **156** ❌ |

`filenameFields` counts every `AXTextField` whose `AXDescription` is `"text field"`. With the file
browser expanded, the browser's own name cells carry that same description — 155 of them here. The
rule holds only when the panel is collapsed to the bare name field, and macOS remembers the expanded
state, so once a user has expanded it the operation cannot succeed again.

## It leaves the panel open, which is the worse half

The timeout path returns before `dismissDialog()` is even defined:

```swift
guard let saveDialog = dialog else {
    return .error("Exact Save As dialog did not appear within 3 seconds")   // panel still on screen
}
func dismissDialog() { … }                                                  // defined after the guard
```

Every following operation then refuses on `preflight_blocking_dialog`. Measured — two subsequent
`save_as` calls and, in the same session, `project.new` and `plugins.*` all came back:

```
{"blocking_dialog_present":true,"dialog_title":"Save","failure_stage":"preflight_blocking_dialog"}
```

The panel has no AX-reachable buttons of its own to answer with (`dialog_buttons: []`), so a caller
cannot clear it either. Escape does.

## Why the message makes it hard to find

*"did not appear within 3 seconds"* names a cause that was not measured — the same shape as #594's
[...]

Issue #594, opened 2026-08-18
record_sequence fails on the first import after a project is created, and succeeds on every retry


MongLong0214:
## Measured, 3 for 3

Every `record_sequence` I have seen fail today failed the same way, and every one of those was the
**first import into a freshly created project**:

```
{"error":"ax_write_failed","file_open_dialog_seen":false,
 "hint":"midi.import_file menu/button click failed:
         IMPORT_BTN_ERROR: Import button never became enabled (file not selected)"}
```

| # | locale | context | result |
|---|---|---|---|
| 1 | ko | first call after `project.new` | FAIL, retry seconds later → State A |
| 2 | en | first call after `project.new` | FAIL, retry seconds later → State A |
| 3 | en | first call after a cold-launch `project.new` | FAIL, then 3 consecutive calls → State A, State A, State A |

Three failures, three recoveries on retry, and no failure yet on a project that had already imported
once. That is a pattern rather than a flake, though it is not yet a root cause.

## Where it gives up

`AccessibilityChannel+MIDIImport.swift`, after the go-to-folder field accepts the path and `return` is
pressed:

```applescript
-- ~4s (20 x 200ms)
set importClicked to false
repeat 20 times
    … if (enabled of ib) then click ib …
    delay 0.2
end repeat
```

The path was accepted — a rejected path returns `DIALOG_NOT_FOUND` from the step above, and that is
not what these runs returned. So the panel opened and took the path; the **Import button never became
enabled within four seconds**.

## Hypothesis, stated as one

Four seconds is calibrated on a warm Open panel. The first panel in a freshly created document is
slower to construct and to resolve the navigation, so the budget expires before the file is selected;
by the second call the panel is warm and the same budget is ample.

I have not confirmed this — confirming it means measuring how long the first selection actually takes,
which needs the budget raised to observe. Filing the pattern rather than the conclusion.

## Why it matters more than its frequency suggests

It is the *first* import that fails, which is the one an agent makes right after creating a project —
the most common opening move there is. And the operation reports `ax_write_failed`, which reads as
"Logic refused the write" rather than "we stopped waiting". A caller that trusts the error stops;
a caller that retries succeeds, which is the wrong lesson to teach.

## Suggested shape

Two separate changes, and the second matters more than the first:

1. Poll to a deadline generous enough for a cold panel, or better, poll on the panel's own state
   (a selected file) rather than on elapsed attempts.
[...]

PR #605, opened 2026-08-18
fix(#604): a save_as refusal must say what it saw and must not leave the panel up

MongLong0214:
Two changes, both about **how it fails**. The classifier is untouched — see below for why that is
the honest outcome rather than an unfinished one.

## The first diagnosis was wrong, and the message caused it

`project.save_as` returned *"Exact Save As dialog did not appear within 3 seconds"*. Timed from the
menu click:

```
Save window appeared after: 0.75 s
```

The budget was never the failure. The message asserted a cause nobody measured — the same shape #594
fixed — and it sent me to raise a timeout that was already generous. That is exactly the wrong fix,
and I nearly shipped it.

## With the refusal reporting what it saw, one clause of seven fails

```
title "Save"   save_buttons 1   save_enabled true   cancel_buttons 1
               package_radios 1  folder_radios 1
               filename_fields 0        ← the rule requires exactly 1
```

A rule with seven conjuncts that reports one bit cannot be diagnosed from its own output. It now
reports every candidate window's observed shape.

## The refusal used to block everything after it

The timeout path returned **before** `dismissDialog` is even defined, so the panel stayed up.
Measured in one run afterwards: two further `save_as` calls, `project.new`, and the plugin operations
all came back `preflight_blocking_dialog` on a `"Save"` window that exposes **no buttons a caller can
answer with** (`dialog_buttons: []`). Escape is the only thing that clears it, and the refusal does
that now.

## Why the classifier is unchanged — the second wrong diagnosis

Two readers disagree about that panel:

| reader | text fields described `"text field"` |
|---|---|
| AppleScript `entire contents` | **156** at depth 8, and 1 at depth 2 |
| this code's `findAllDescendants` | **0** at depth 12 |

I tried a shallow search and an ancestor-based filter that excludes browser/outline/table subtrees.
Both were driven against the live panel; **both still counted zero**. Neither is the rule.

Whatever the panel's real shape is through this reader, it is not what either candidate assumed, and
a fix that "works" without explaining that disagreement would be a coincidence. So the classifier
ships unchanged and the measurement is filed on #604 — the next attempt starts from data instead of
from a timing hypothesis.

## Live proof

```
7 checks · 7 passed · 4 mutation-backed · visual 1/1 · restoration measured
```

The run drives the failing operation **on purpose** and checks how it fails: that the refusal names
the shape it saw, that it no longer claims a timing cause, that no panel is left on screen, and —
the decisive one — that an unrelated operation immediately afterwards is **not** refused by something
this one left behind.

The harness also had to stop picking the wrong window: a plugin window carries the *track* name
(`Absolute Zero`), and taking the first document title gave one `logic_window` could not locate. It
now picks the largest window it can actually find on screen, which needs no knowledge of the title's
language.

## Gate

`lpm-ship.sh` PASS at `6ad01599` — preflight clean, `Package.resolved` untouched, branch contains

### DIFF
--- a/Scripts/livekit/live_604_refusal_says_what_it_saw.py
+++ b/Scripts/livekit/live_604_refusal_says_what_it_saw.py
@@ -1,0 +1,172 @@
+#!/usr/bin/env python3
+"""Live proof that a save_as refusal is legible and leaves nothing blocking behind it.
+
+Usage:  LPM_EVIDENCE_ROOT=/abs/path/outside/repo \
+        python3 live_604_refusal_says_what_it_saw.py <worktree> <full-40-char-head-sha>
+
+WHAT THIS RUN IS ABOUT
+----------------------
+`project.save_as` does not work on this host, and this change does not fix that. What it changes is
+the refusal.
+
+The old message was "Exact Save As dialog did not appear within 3 seconds". Timed from the menu
+click, the panel appears in 0.75s — so the budget was never the failure, and the message sent the
+reader to the wrong place. It also returned before the dismissal helper is even defined, leaving the
+panel on screen; every later operation then refused with `preflight_blocking_dialog` on a "Save"
+window that exposes no buttons a caller can answer with.
+
+So the run drives the failing operation on purpose and checks two things about how it fails: that the
+refusal names the shape it actually saw, and that the next operation is not blocked by a leftover
+panel.
+
+The classifier is untouched. Its filename-field clause is what fails, and two readers disagree about
+that panel — AppleScript's `entire contents` finds 156 text fields described "text field" while this
+code's own reader finds zero at the same depth. Until that is explained, any replacement rule is a
+guess, and the honest deliverable is a failure that can be read.
+"""
+
+import json
+import os
+import subprocess
+import sys
+import time
+
+sys.path.insert(0, os.path.dirname(os.path.abspath(__file__)))
+import evidence as E  # noqa: E402
+
+WT = sys.argv[1] if len(sys.argv) > 1 else ""
+HEAD = sys.argv[2] if len(sys.argv) > 2 else ""
+if not WT or not HEAD:
+    sys.exit(__doc__)
+
+E.REPO = WT
+E.BIN = f"{WT}/.build/release/LogicProMCP"
+missing = E.have_tools()
+if missing:
+    sys.exit(f"cannot run: missing {missing}")
+
+ev = E.Evidence(HEAD, os.environ["LPM_EVIDENCE_ROOT"])
+CHOOSER_TITLES = ("Choose a Project", "프로젝트 선택")
+TARGET = "/Users/isaac/Music/Logic/lpm-604-refusal-proof.logicx"
+
+
+def osa(script):
+    r = subprocess.run(["osascript", "-e", script], capture_output=True, text=True)
+    return (r.stdout or "").strip()
+
+
+def window_titles():
+    raw = osa('tell application "System Events" to tell process "Logic Pro" to '
+              'return name of every window')
+    return [t.strip() for t in raw.split(",") if t.strip()] if raw else []
+
+
+def document_titles():
+    return [t for t in window_titles() if not any(c in t for c in CHOOSER_TITLES)]
+
+
+def inner(result):
+    return result if isinstance(result, dict) else {}
+
+
+titles = document_titles()
+ev.check("604/precondition-a-project-is-open", bool(titles),
+         "Logic has a project window, so File > Save As has something to save",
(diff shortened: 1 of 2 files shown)


MongLong0214:
## An independent blind review blocked the first cut. Every finding held.

I put this branch in a throwaway worktree at the exact merge commit and asked a reviewer from a
different vendor to try to **refute** the claim, with the target list named and `filesRead`/`notRead`
required in the response. It came back `BLOCK` with three dead live checks and four places where the
product published something it never looked at. I verified each one against the source before acting
on it — none were wrong.

### Three live checks could not fail

| check | why it could not go red |
|---|---|
| `no-panel-is-left-blocking-the-next-operation` | compared `len(after_titles)` to `len(document_titles())`, and `document_titles()` only drops the project-chooser titles. A leftover window named `Save` sits in **both** lists, so the comparison only ever said "no chooser is visible". The mutation it named could not flip it. |
| `the-next-operation-is-not-blocked-by-what-this-one-left` — the one the file calls decisive | used `system.health`, which builds its report from channel health, cache, permissions and process metrics and never consults `blockingDialogInfo()`. It cannot report a blocking dialog whatever is on screen. |
| the visual, `the-panel-is-gone-from-the-screen` | `shot()` is `screencapture -l <arrange window id>`. A top-level `Save` AXDialog is not in that frame, so `expect_change=False` was true whether the panel was up or gone — the assertion was blind to its own subject. |

Fixed by asking the questions directly: a `save_panels()` predicate that enumerates Save windows and
sheets; `list_library`, which is read-only **and** runs the same preflight the mutating operations do
(`blockingLogicDialogResult(operation: "library.list")`); and a visual that now claims only the
narrower true thing — the arrange window behind was undisturbed. It runs after the visual, because
listing the Library can open Logic's Library panel and move the window the visual guards.

### Four places the product published an absence

- The dismissal said *"The panel, if one opened, has been dismissed."* without looking. The click that
  opens the panel is AX; the Escape is System Events. Deny Automation and the panel stays while the
  sentence still says it is gone. It now counts panel-shaped windows before and after and reports
  whether the keystroke was delivered at all.
- The candidate dump **dropped every window whose title would not read**, then printed
  `Candidates seen: none` — the same defect the message exists to cure.
- `subrole` was omitted while being one of the classifier's seven conjuncts, so a failing clause was
  invisible in the message meant to diagnose it. And `enabled ?? false` reported an *unread* Save
  button as a *disabled* one.
- The refusal was a bare sentence. That is not a State C envelope, so a single-channel chain could not
  surface it and relabelled this deliberate refusal `channels_exhausted`. It is now a real State C
  carrying `write_attempted: false`.

### And the review sent me to the actual root cause

Chasing the leftover-panel check meant asking what a Save panel looks like. Measured today, four
trials, deterministic both ways:

```
Logic backgrounded → click File ▸ Save As…  →  0 windows named "Save"   (0/2)
Logic frontmost    → click File ▸ Save As…  →  1 window  named "Save"   (2/2)
                                                AXDialog, 157 text fields, exactly ONE focused
                                                focused field value = "Untitled"   ← the filename field
```

`clickMenuItem` is an `AXPress` and never activates Logic, so in production conditions **the panel
never opens at all**. That is why this file's own story said one reader saw 156 text fields and the
product's reader saw zero at the same depth: I had activated Logic before probing and the product
never does. The docstring said that disagreement was unexplained; it is explained now, and it was not
about depth.

So the refusal now also reports whether Logic owned the keyboard, and the harness activates Logic
before driving the failure — otherwise no panel opens, the dismissal under test never runs, and the
checks cannot tell a working dismissal from a missing one.

With the panel actually present, the refusal reads:

```
error: element_not_found
hint:  Save As panel was not recognised by the structural classifier. Candidates enumerated: 3.
       Shapes: [title: "Save", role: AXWindow, subrole: AXDialog, filename_fields: 0,
                save_buttons: 1, save_enabled: true, cancel_buttons: 1,
                package_radios: 1, folder_radios: 1] | [title: "Absolute Zero", …]
       Logic owned the keyboard when the menu was pressed: true.
       Escape cleared the 1 panel-shaped window(s) that were open.
```

Six of seven conjuncts pass and the message names the one that fails. That is the whole point of the
change, and it now demonstrably does it.

**A working `save_as` is still not in this PR** — `filename_fields: 0` through the product's reader
while AppleScript counts 157 on the same panel is a second, separate wall. It gets its own change and
its own proof.

### Gate

```
full suite                    3843 tests passed
live evidence for this head   9 checks, 9 passed, 5 mutation-backed, 1 visual, 0 failed
SHIP GATE PASS                f70b4e40
```

The new unit test was mutation-tested: forcing `summary` to always claim the panel was cleared turns
it red with 6 issues, and reverting turns it green.

PR #611, opened 2026-08-19
test(livekit): derive the track-header band from AX instead of writing a rectangle

MongLong0214:
Draft until an independent blind review passes — see the convention note at the bottom.

Recovered from the branch behind #597. **#597 itself is superseded**: I re-counted the table on
current `main` and every row it retires is already gone, along with every dead switch arm it removes.

```
mixer.set_output / set_input / toggle_eq / reset_strip
    RoutingTable rows                    0   (removed; the table even carries a #592 comment saying so)
    AccessibilityChannel refusal arms    0   (removed)
    anything routing them                0
    what remains                         an explicit notExposedCommandResult in the dispatcher

plugin.list / automation.get_mode
    absent everywhere outside the table, and absent from the table too
```

Main also carries a **later and more accurate** version of that harness's prose — the correction that
`mixer.set_send` and `automation.set_mode` lost their accessibility arms *without* losing their rows,
because those two operations work on the channels their chains actually name. The branch has the
earlier wording.

So the only thing on it that `main` lacks is this helper, and it is worth keeping.

### What it is for

A band written into a harness is right until someone resizes a window. The #592 harness carried
`(603, 162, 325, 406)`, measured on a 1920x1050 window with the Library closed. After the window
changed shape that rectangle landed on the **Step Sequencer**, which animates on its own — and a
NEGATIVE visual assertion over it then reported movement in a part of the screen the run had never
touched, failing a run in which nothing was written.

`track_header_band()` asks AX for the rail at run time. It returns `None` when the rail cannot be
found, which a caller should treat as a failed precondition rather than a reason to guess.

Two AppleScript hazards are handled inside it so a caller does not rediscover them:

- `by` is a reserved word, so a variable named that fails to **parse** rather than to run.
- `first window whose name is "…"` has been measured answering *invalid index* for a window that
  `name of every window` listed a moment earlier, with a plain-ASCII title that matched exactly. So
  the window is chosen by **what it contains** — a track rail — not by its name. Excluding the
  project chooser by title was not enough either: a run that overlapped an "Import" dialog picked
  that dialog and got nothing.

### One correction to the recovered text

The docstring claimed a specific past measurement — `366,707 235x1182` — that I had no way to
reproduce. Inherited text keeps asserting old facts, so I replaced it with one taken here today
against the window on screen now:

```
track_header_band() -> (366, 527, 235, 522)
logic_window('Tracks') -> 1920x1050 at 0,30
```

`main`'s `live_592` harness is **not** rewired to use it in this PR. Its band is already derived from
the window frame rather than hard-coded, so there is no stale-constant defect left there to fix, and
switching it would change what that harness asserts without a run to show the new assertion still
holds.

### Gate

```
full suite                    3846 tests passed
live evidence                 n/a — no Sources/ change
SHIP GATE PASS                dc736912
```

### Convention

Two of my PRs this week were merged one commit before the blind review's revision existed — not a
race, just me opening them ready before the review had run. From now on mine open as **draft** and
become ready only after a cross-provider blind review passes. A non-draft PR from me means the review
is done.

### DIFF
--- a/Scripts/livekit/evidence.py
+++ b/Scripts/livekit/evidence.py
@@ -14,6 +14,10 @@ What it refuses to let you record:
   resulting image is not what the user sees. Every capture records which display it fell wholly within.
 - **A whole-window diff as a visual assertion.** A full-window comparison changes when the clock ticks.
   `visual()` requires an explicit region and states what it expected to happen there.
+- **A region measured once and carried as a constant.** A rectangle taken on one window shape silently
+  watches something else after a resize — measured, a stale band landed on the Step Sequencer and made
+  a negative assertion report movement in a part of the screen the run never touched. Derive the band
+  at run time; `track_header_band()` does it for the arrange rail.
 - **A check that cannot fail.** `check()` requires you to name the mutation you applied to the product
   that flipped it. A check nobody has seen fail is not evidence.
 - **A cached read presented as live.** `provenance()` records the source and age of any value that did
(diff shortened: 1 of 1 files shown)


MongLong0214:
Closing rather than fixing. A blind review blocked this, I verified the finding directly, and the
verification showed the helper is worse than "imprecise" — it returns a confident wrong band.

### The helper returned the Inspector, not the rail

Its rule was "first `AXLayoutItem` with a text-field child, return its grandparent". I replicated
exactly that rule through the raw AX API against the live window:

```
HELPER PICKS: AXGroup  window-relative 366,527  235x522
              its parent: AXLayoutArea  desc = "Mixer"
```

That is the Inspector channel strip, which carries animating level meters — the same class of
false-positive the helper exists to prevent, except now arriving as a *successful* rail read rather
than as `None`.

**And I had published that number as verification.** The docstring I wrote said "Measured here on
2026-08-19 … 366,527 235x522 — the rail". I ran it, got numbers, and treated "it returned a number"
as "it returned the right thing". The reviewer spotted it from the number alone: `x=366, w=235` is
the Inspector column, and this repository had already measured that layout in
`live_576_completeness_is_measured.py`.

### What the rail actually is

Enumerating every container of `AXLayoutItem` rows in the same window:

```
AXLayoutArea "Mixer"               366,527   235x522     <- what the helper picked
AXLayoutArea "Tracks time ruler"   928,124   977x38
AXGroup      "Tracks header"       603,116   325x295     <- the rail
AXLayoutArea "Track 26 …"          837,246   …
AXLayoutArea "Mixer"               603,367   1317x683
```

Note `x=603, w=325` on the rail — **the constant this PR set out to replace, `(603, 162, 325, 406)`,
was aimed at the right column all along.** Its hazard is that it goes stale, not that it was ever
wrong; and `main`'s `live_592` band is already derived from the window frame rather than hard-coded,
so there is no stale-constant defect left there to fix.

Also worth recording: `AXSelectedChildren` does **not** separate these — all five expose it. The
product's `isTrackHeadersGroup` works because it is applied to candidates already narrowed by
`getTrackHeaders`, not as a standalone discriminator.

### And the description rule cannot be built in AppleScript here

I rewrote it to match `AXDescription == "Tracks header"` exactly, failing closed to `None` on anything
else. It returns `None` — both via System Events' `description` property and via
`value of attribute "AXDescription"`. `entire contents` never reaches that element, while a Swift walk
to depth 14 finds it immediately. That is the second time today the same trap has cost me: a rule
prototyped through System Events does not necessarily hold through the API the product uses, and here
System Events cannot even see the element.

So a correct version of this helper is a Swift witness, not an osascript string — which means
duplicating `getTrackHeaders`, logic that already exists and is already hard to keep right.

The stale-band hazard is real and stays worth guarding. This is not the guard.

Issue #628, opened 2026-08-19
Identity from tree order: findDescendant returns the first match and nothing records that there was only one


MongLong0214:
## The class

> **Identity came from tree order rather than from a discriminator, and because the result happened to be right, nothing said it was luck.**

This is a step worse than the class #626 addresses. There, a check could not *see* what it had to answer. Here a check *does* see something, treats it as established, and **is silent on success** — so a correct answer and a lucky one are indistinguishable afterwards.

## Where it lives — measured, not assumed

Both shared locators return the first match in traversal order, with no signal that other candidates existed:

```swift
// AXLocalePolicy.swift:1090
AXHelpers.findAllDescendants(of: element, role: role, …).first { elementMatches($0, labels, …) }

// AXHelpers.swift:479-481
if roleMatch && titleMatch && idMatch { return child }
```

Census across the tree:

```
findDescendant( / findAllDescendants( call sites in Sources/     110   in 19 files
.first(where:                                                     68
firstIndex(where:                                                 11
livekit AX tools                                                   4 files
```

So this is **not** a `Scripts/livekit/` problem. That was checked rather than assumed, the same way #626's scope was, and the answer again came back wider than the starting guess.

## How it was found

`ax_control_bar_band.swift` looked an element up by `AXDescription` and walked depth-first. Measured on one arrange window, `"Control Bar"` matches **two** elements, `"Library"` four, `"Event"` three. It returned the right Control Bar — **which is worse than returning the wrong one, because nothing showed that it was luck.**

`live_291`'s own comment already records the same trap from the other side: matching anything described `"Mixer"` hits the Inspector's strip area, and *"this detector has been wrong twice."* The harness knew what the tool did not.

## What a fix has to do, and the half that is easy to miss

Refusing when the name is ambiguous is **half**. #627 does that for the livekit tool.

The other half: **the fact that it was NOT ambiguous has to survive into the record.** A locator that emits `candidates: 1` leaves evidence a later reader can weigh. One that stays silent on success leaves exactly what we have now — a result with no way to tell a discriminator from tree order.

So both of these must be demonstrable:

```
two candidates exist   →  refused, and both are named
one candidate exists   →  recorded as one, not merely accepted
```

## Suggested decomposition

Rewriting 110 call sites is not the first move and probably not the right one. Nearer term:

1. an ambiguity-aware locator in the AX layer that reports its candidate count, alongside the existing one
[...]

PR #650, opened 2026-08-21
feat(#628): the transport scan says when it chose among several

MongLong0214:
> **Updated at `8f8c5f5b`** after a second adversarial review returned **BLOCK** on four defects. Three are fixed, the fourth is measured and recorded rather than guessed at; details in the comments. Two claims below have changed and are corrected inline.

**Rewritten after an outside review found three defects in it.** The original version of this PR
added a log line and claimed that made an ambiguity visible. That claim was false on this project's
own observation channel, and two more defects were in the work around it. All three were verified
against the tree before being accepted.

### 1. Composes the discriminated accessor instead of guessing

`getTransportBar` gathers groups, narrows with `looksLikeTransportContainer`, and took the first
survivor. Measured on Logic 12.3: **37 gathered, 4 survive.**

I originally declined to fix this, calling it a contract decision — *is a "transport bar" a "control
bar"?* **That was refuted by this repository.** `getControlBar` (`Transport.swift:53-84`) already
implements the discriminator: label match → holds-a-checkbox → exactly-one-or-nil. And
`AccessibilityChannel+Transport.swift:11` already writes `getControlBar() ?? getTransportBar()` at a
call site. A caller had made the decision, in merged code, while I called the question open.

Measured **before** changing anything:

```
getControlBar() resolves          true
scan returns something            true
same element                      TRUE
```

So the composition is observationally a no-op on the measured machine, and the difference appears
only on a tree where the scan's first survivor is not the control bar — the tree nobody has seen,
and the reason guessing was wrong.

**Correction (review 2).** Composing it *unconditionally* was wrong. `getControlBar`'s last branch
returns a lone labelled group holding **zero** checkboxes, so a labelled shell could shadow a
container that works, and `findTransportButton` would then search only the shell and return nil.
The composition now also requires the identified bar to actually **hold a transport control**.

My first attempt validated with `looksLikeTransportContainer` and **did not work** — that function's
first branch matches on *metadata*, so anything described "Control Bar" satisfies it whether or not
a control is inside. The new test caught it returning the shell anyway. The control-hit computation
is extracted (`transportControlKeywordHits`) and reused, not copied.

**Still not the fail-closed variant, and that half of the refusal holds.** `currentPlayheadBar`
(`AccessibilityChannel+Transport.swift:620`) and the cycle-range reader (`Regions.swift:330`) take
`getTransportBar` with no fallback; exactly-one-or-nil there would regress them to nil. Composition
needs no contract answer. Refusal does.

### 2. Removes a predicate this repo now had two copies of

The probe added in #649 contained a **verbatim copy** of `findPanControlInHeader`'s closure. I
refused predicate reimplementation four times in #628's thread and wrote comments about why it is
the failure being censused — then shipped a copy. `headerPanSliderCandidates` is the single
definition now; the product and the probe both call it. Live measurement unchanged: 2 → 1.

### 3. Teaches the census the spelling that hid the defect

The detector required `findAllDescendants(…)` and `.first` to be **adjacent**. This codebase's
dominant spelling binds the result to a name first:

```swift
let groups = AXHelpers.findAllDescendants(…)
groups.first(where: { … })
```

So the thirty-seven-site list contained **none** of `Transport.swift:24`, `Mixer.swift:309`,
`Mixer.swift:322`, `Tracks.swift:54` — including the one site measured live as genuinely ambiguous.
**It reported "at budget" while that site shipped.**

Budget **37 → 59**, with the reason recorded beside the constant, matching the one prior legitimate
rise (27 → 39, also a spelling it had been blind to). The twenty-two are not new code; only the
instrument changed.

## What the log line actually reaches — corrected

`Log.info` writes to **stderr**, and the live harness runs the server with `stderr=DEVNULL`
(`Scripts/livekit/evidence.py:184`). **It cannot reach the evidence document.** The original PR body
said this made the ambiguity visible; it makes it visible to an operator reading a server log, and
to nothing the ship gate reads. That limit is now stated at the call site rather than left implied.

The log is also narrower now, and more interesting: it fires only when `getControlBar` could **not**
identify one *and* the scan found several — both routes failing, rather than merely the scan
guessing.

## Verification

```
9 unit tests green, including the pre-existing Issue107 mixer tests (the extraction moved nothing)
harness 17/17 · 14 mutation-backed · live evidence bound to this head
SHIP GATE PASS · 3891 tests
```

Mutation control retained from the original: truncating the candidate list to one turns two cases
red, and `splitPreservesTheChoice` pins that the refactor did not move which element is returned.

## What this still does NOT do

- **It does not record uniqueness.** #628's contract asks for *"one candidate exists → recorded as
  one."* When `getControlBar` identifies one, nothing is written. That is still open, and it is
  literally the issue's title.
- **The mechanism behind 37 → 4 is not pinned.** `controlHits.count >= 2` requires two **distinct**
  keywords; I named one (a track header's record-arm checkbox). Either a second keyword-bearing
  control went unnamed or a different branch fired. The probe records counts, not which branch.
- **`transportRecordArmExclusion` exists for exactly this confusion** (`AXLocalePolicy.swift:522`,
  rationale: *"distinguishes per-track record-arm from transport Record"*) and
  `looksLikeTransportContainer` never uses it. That is the next fix and it is a predicate patch, not
  a contract question.

---

# Added after the first review round: the mechanism, pinned — and the cause I published was wrong

The PR above said the arrange area qualifies because *"a track header holds a record-arm checkbox"*.
A reviewer noted the rule needs two **distinct** keywords and I had named one. Instrumenting the
predicate per survivor:

```
'Control Bar'  2-kw=True   [cycle, loop, metronome, play, record]
'Control Bar'  2-kw=False  []
'Tracks'       2-kw=True   [loop, play]   show/hide live loops grid~loop · catch playhead~play
'Tracks'       2-kw=True   [loop, play]   (same two labels)
```

**`record` is not among the keywords either arrange group matches.** The mechanism is substring
matching on short generic words:

```
"play" ⊂ "Catch Playhead"        "loop" ⊂ "Show/Hide Live Loops Grid"
```

That also killed the follow-up I had announced — wiring in `transportRecordArmExclusion` would have
removed nothing and left the four at four.

## The guard, measured in both shipped locales

`contains` cannot become word-boundary matching: Korean and Japanese labels have no word boundaries,
and `재생헤드` is reachable only by substring. So the fix is a negative label set, the shape this file
already uses. Read off a live window under `-array en` and `-array ko`:

```
Catch Playhead              재생헤드 캐치
Playhead Position           재생헤드 위치
Loop Browser                루프 브라우저
Show/Hide Live Loops Grid   Live Loop 그리드 보기/가리기   ← keeps the ENGLISH "Loop"
(one control's English name) unchanged in Korean          ← untranslated, keeps "play"
```

The last two are why **both** halves had to be measured: a guard written only in Korean misses
labels that stay English inside a Korean UI, and one written only in English misses `재생헤드 캐치`.
ja-JP is deliberately absent — nobody has read these off a Japanese Logic, and inventing them is the
defect #519 exists to remove.

**Live result: survivors 4 → 2**, both remaining are real control bars, which `getControlBar`
already separates by "holds a checkbox".

Controlled: removing the guard turns both false-friend cases red, in English and Korean. A third
case pins that it does not swallow a genuine transport container.

## One note on the suite

A gate run failed with three order-dependent failures (`TraceClearAudit`,
`QualificationFaultInjection` ×3) while a release build and a live harness were competing for the
machine. All sixteen pass in isolation, and the suite passes clean and uncontended (631s vs 697s on
the failing run). Two clean passes against one failure under load — recorded as contention rather
than asserted as proven, since three data points is not a flake analysis.

---

## Added by review 2

**The test that pinned the selection was tautological.** It compared `filter(...).first` against
`first(where:)` with the same predicate on both sides — it could not fail, and reverting the
production change left it green. Replaced by two cases that drive `getTransportBar` itself; the
shadowing one was watched to fail with the composition unguarded.

**The production comment still carried the disproven record-arm cause.** I corrected it in the test
file earlier and left the same wrong sentence standing in production, in the same file.

**Four of five census labels were wrong after this branch, and all five were correct on `main`.**
One named the wrong file. Every label already carries the function name, so the line number is
dropped — 9 baked `file.swift:NNN` strings in `Sources/` go to 0. Symbols are compiler-checked;
a line number in a string is checked by nobody, which is why three were still wrong after a clean
merge with #651.

**Not fixed, recorded:** with the false-friend guard, an *unlabelled* container holding `Record`
plus `Catch Playhead` and no text/slider hints drops to one keyword and no longer qualifies.
Measured on Logic 12.3: both live Control Bar groups are metadata-labelled and carry zero text
descendants, so the metadata branch answers before the guard is reached and this cannot bite here.
It threatens only the legacy unlabelled path.

```
SHIP GATE PASS @ 8f8c5f5b — 3895 tests · census at budget (59) · live evidence bound to head
```

### DIFF
--- a/Scripts/check-ax-locator-census.py
+++ b/Scripts/check-ax-locator-census.py
@@ -46,7 +46,21 @@ import sys
 # had been blind to. Raising a ratchet for that reason is legitimate and has to be visible, which is
 # why the reason lives here rather than in a commit nobody re-reads. Raising it because something
 # went red is not.
-BLIND_SITE_BUDGET = 37
+#
+# It rose a second time, 37 -> 59, for the same reason and found by an outside review rather than by
+# this file. The detector required `findAllDescendants(…)` and `.first` to be ADJACENT, so the
+# codebase's dominant spelling —
+#
+#     let groups = AXHelpers.findAllDescendants(…)      // one line
+#     groups.first(where: { … })                        // several lines later
+#
+# — was invisible. Measured: the thirty-seven-site list contained NONE of `Transport.swift:24`,
+# `Mixer.swift:309`, `Mixer.swift:322`, `Tracks.swift:54`, including the one site measured live to
+# be genuinely ambiguous (37 groups narrowed to 4 by its own discriminator, then first-in-tree-order
+# taken). A census blind to the defect it was built for reported "at budget" while that site shipped.
+#
+# The twenty-two are not new code. They were always there; only the instrument changed.
+BLIND_SITE_BUDGET = 59
 
 SEARCH_ROOTS = ("Sources", "Scripts")
 
@@ -67,8 +81,25 @@ def blind_sites(root):
                     out.append((path, i + 1, "findDescendant — returns first, no count"))
                 elif re.search(r"\bfindAllDescendants\(", line):
                     window = "\n".join(lines[i:i + 14])
-                    if re.search(r"\)\s*\.first\b", window):
-                        out.append((path, i + 1, "findAllDescendants(...).first"))
+                    # The result BOUND TO A NAME and then reduced — `let groups = findAll…` on one
+                    # line and `groups.first(where:)` further down. This is the dominant spelling in
+                    # this codebase and the detector could not see it: measured 2026-08-21, the
+                    # thirty-seven-site list contained NONE of `Transport.swift:24`,
+                    # `Mixer.swift:309`, `Mixer.swift:322`, `Tracks.swift:54` — including the one
+                    # site measured to be genuinely ambiguous live (37 groups narrowed to 4).
+                    #
+                    # A census that cannot see the defect it was built for is the third instance of
+                    # its own docstring's warning, found by an outside review rather than by the
+                    # instrument.
+                    bound = re.match(r"\s*(?:let|var)\s+([A-Za-z_][A-Za-z0-9_]*)\s*=", line)
+                    reduced_by_name = bool(
+                        bound and re.search(
+                            r"\b" + re.escape(bound.group(1)) + r"\s*(?:\.first\b|\.last\b|\[0\])",
+                            window,
+                        )
+                    )
+                    if re.search(r"\)\s*\.first\b", window) or reduced_by_name:
+                        out.append((path, i + 1, "findAllDescendants(...) reduced to one"))
                     elif re.search(r"\bfor\b[^\n]*\{", window) and re.search(r"\breturn\b", window):
                         # `for candidate in … { if matches { return candidate } }` is the same
                         # first-match, written long-hand. The census missed thirteen of these while
--- a/Sources/LogicProMCP/Accessibility/AXLocalePolicy.swift
+++ b/Sources/LogicProMCP/Accessibility/AXLocalePolicy.swift
@@ -767,6 +767,44 @@ enum AXLocalePolicy {
         rationale: "Counts distinct transport-control labels to classify the control bar; read-only."
     )
 
+    /// Labels that carry a transport keyword without being a transport control.
+    ///
+    /// `transportContainerControlKeywords` matches with `contains`, which is required: Korean and
+    /// Japanese labels have no word boundaries, so `재생헤드` can only be reached by substring. The
+    /// cost is that short generic words match unrelated controls, and two of them put the ARRANGE
+    /// AREA into `looksLikeTransportContainer` — measured 2026-08-21, Logic 12.3:
+    ///
+    ///     "play" ⊂ "Catch Playhead"                 "loop" ⊂ "Show/Hide Live Loops Grid"
+    ///
+    /// Those two alone gave the track area the two distinct keywords the rule needs, so
+    /// `getTransportBar`'s scan had four survivors where it should have had two.
+    ///
+    /// Measured in BOTH shipped locales by running the same window under `AppleLanguages -array en`
+    /// and `-array ko`, because the Korean forms are not translations of the English ones:
+    ///
+    ///     Catch Playhead              재생헤드 캐치
+    ///     Playhead Position           재생헤드 위치
+    ///     Playhead thumb              재생헤드 썸네일
+    ///     Loop Browser                루프 브라우저
+    ///     Show/Hide Live Loops Grid   Live Loop 그리드 보기/가리기   ← keeps the ENGLISH "Loop"
+    ///     Session Player              Session Player                ← untranslated, keeps "play"
+    ///
+    /// The last two are why this is a measured table and not a translation: a guard written only in
+    /// Korean would miss labels that stay English inside a Korean UI, and a guard written only in
+    /// English would miss `재생헤드 캐치`. Both halves were read off a live window.
+    ///
+    /// ja-JP is NOT here. Nobody has read these labels off a Japanese Logic, and the ship scope is
+    /// Desktop × {en-US, ko-KR}. Inventing them is the defect #519 exists to remove.
+    static let transportKeywordFalseFriends = LabelSet(
+        canonical: "catch playhead",
+        variants: ["playhead position", "playhead thumb", "loop browser", "session player",
+                   "show/hide live loops grid", "live loops grid",
+                   "재생헤드 캐치", "재생헤드 위치", "재생헤드 썸네일", "루프 브라우저",
+                   "live loop 그리드 보기/가리기"],
+        rationale: "Negative guard: labels carrying a transport keyword that are not transport "
+            + "controls. Measured en-US and ko-KR on Logic 12.3; read-only."
+    )
+
     /// Tempo/position slider description tokens inside the transport container.
     static let transportSliderHints = LabelSet(
         canonical: "tempo",
--- a/Sources/LogicProMCP/Accessibility/AXLogicProElements+Mixer.swift
+++ b/Sources/LogicProMCP/Accessibility/AXLogicProElements+Mixer.swift
@@ -57,15 +57,27 @@ extension AXLogicProElements {
         return findPanControlInHeader(header, runtime: runtime.ax)
     }
 
-    /// Header-level pan-slider selection (split out for deterministic testing).
-    static func findPanControlInHeader(_ header: AXUIElement, runtime: AXHelpers.Runtime = .production) -> AXUIElement? {
-        let sliders = AXHelpers.findAllDescendants(of: header, role: kAXSliderRole, maxDepth: 4, runtime: runtime)
-        if let pan = sliders.first(where: { slider in
+    /// Every slider on a header whose children describe it as a pan control, in tree order.
+    ///
+    /// Split out so a probe can call the SAME predicate the product runs rather than a copy of it.
+    /// The copy existed for one commit and is exactly what #628 is a census of: a measurement that
+    /// can drift from the code and then disagree with it for a reason unrelated to the tree.
+    static func headerPanSliderCandidates(
+        among sliders: [AXUIElement],
+        runtime: AXHelpers.Runtime = .production
+    ) -> [AXUIElement] {
+        sliders.filter { slider in
             AXHelpers.getChildren(slider, runtime: runtime).contains { child in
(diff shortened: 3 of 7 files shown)


MongLong0214:
## This branch broke its own census label — same class the #651 review just found

Found while checking merge interaction with #651, before the independent review returned.

`MainEntrypoint.swift:401` and `:406` both name the site as:

```
AXLogicProElements+Mixer.swift:63 header pan slider
```

On this branch, `Mixer.swift:63` is a **doc-comment line inside the newly added `headerPanSliderCandidates`**. The site it actually describes, `findPanControlInHeader`, moved to **line 78** — pushed down by the ~12 lines this branch inserts above it.

The reviewer on #651 caught the identical thing there (`:309/:322` → `:327/:353`). So this has now drifted **twice in one day, in two branches, in the same file.**

## The instance is not the defect

There are at least four of these baked line numbers:

```
AXLogicProElements+Transport.swift:165 playheadPositionGroupLabel
AXLogicProElements+Mixer.swift:63      header pan slider
AXLogicProElements+Mixer.swift:309     findVolumeFader     (#651: now :327)
AXLogicProElements+Mixer.swift:322     findPanControl      (#651: now :353)
```

Every one of them is a string that silently stops being true the moment anyone edits above it, and **nothing can detect that** — no test asserts these labels (I checked; the `site` assertions in `Issue549`/`Issue523` are different fields). A diagnostic whose job is to tell you where to look, pointing at the wrong place, is worse than one that says nothing.

They already carry the function name. The line number adds nothing that survives an edit.

## Merge-order consequence

These two branches both touch `AXLogicProElements+Mixer.swift` and `MainEntrypoint.swift`, and this one inserts **above** #651's region. So whichever lands second has its numbers shifted by the first.

Sequencing accordingly: **this PR first**, then #651 rebases and drops the line numbers from all four labels in one change — removing the class rather than chasing the instances. #651 will need a fresh gate run after the rebase regardless, since its live evidence is bound to its current head.

MongLong0214:
## Measured: after both merge, three of four labels point at the wrong line — and git merges cleanly

I simulated the sequence rather than reasoning about it: `main` + #650 + #651, both auto-merging with no conflict.

In that merged tree:

```
label                              points at                        the real site is
Mixer.swift:63   header pan slider   a doc-comment line               :78   findPanControlInHeader
Mixer.swift:327  findVolumeFader     a log-message continuation line  :339  return sliders.first
Mixer.swift:353  findPanControl      a log-message continuation line  :365  return sliders.count > 1 ? …
```

Both merges reported `Automatic merge went well`. **Nothing in the pipeline can see this** — not the compiler, not the 3895-test suite, not the merge, not the live gate. The labels are strings.

Worth stating plainly: **all four were correct on `main`.** They are not neglected; they are accurate right up until anyone edits above them, and both of these branches do. That is what makes the line number the defect rather than any particular stale value.

Plan stands: merge this PR first, then #651 rebases and replaces all four with function names — which are already in the labels and already stable.

MongLong0214:
## Adversarial review returned BLOCK; three defects fixed at `8f8c5f5b`, the fourth recorded

A pre-merge review at high effort (~360k tokens) read the branch, ran the suite and the census, and cross-checked the Korean labels against Apple's own documentation. Verdict was **BLOCK**. Its findings held up.

### 1. `getControlBar`-first could hand back a bar with no transport control in it — fixed

`getControlBar` ends with a branch that returns a lone labelled group holding **zero** checkboxes. Composing it unconditionally meant `findTransportButton` would search only that and return nil, where the scan would have found a working container.

**My first fix was wrong, and the new test caught it.** I validated with `looksLikeTransportContainer` — whose *first branch matches on metadata*, so anything described "Control Bar" satisfies it whether or not a single control is inside. It returned the shell anyway. I had written *"an accessor that answers by NAME and a caller that needs a CAPABILITY are not the same question"* and then used a name-based predicate as the validator.

The control-hit computation is now extracted (`transportControlKeywordHits`) and reused — not copied, since a copied predicate that drifts is what #628 is a census of.

### 2. The test that claimed to pin the selection was tautological — replaced

It compared `filter(...).first` against `first(where:)` **with the same predicate on both sides**. It could not fail; reverting the production change left it green. It was asserting a property of Swift.

Two cases now drive `getTransportBar` itself, and the shadowing one was **watched to fail** with the composition unguarded.

Building it corrected my own fixture too: with the shell placed *first*, the bare scan returns it as well — the metadata branch qualifies "Control Bar" with nothing inside, which is pre-existing behaviour this branch does not change. Asserting on that would have made the case fail on `main` as loudly as here.

### 3. The production comment still carried a disproven cause — fixed

`AXLogicProElements+Transport.swift` still said the arrange area matched *"because a track header holds a record-arm checkbox"*. Instrumentation disproved that — `record` never matched; it was `"play"` inside `"Catch Playhead"` and `"loop"` inside `"Show/Hide Live Loops Grid"`. I had corrected it in the test file and left it standing in production: **the same wrong sentence, in the same file, twice.**

### 4. The sparse-bar false negative — real, narrow, recorded not guessed at

With the guard, an **unlabelled** container holding `Record` + `Catch Playhead` and no text/slider hints drops from two keywords to one and no longer qualifies. Apple does document hiding individual control-bar buttons, so sparse bars are supported rather than synthetic.

Measured on Logic 12.3 before deciding: both live Control Bar groups **are** metadata-labelled, and both carry **zero** text/field descendants within depth 4. So the metadata branch answers before the guard is ever reached, and this cannot bite on this host — it threatens only the legacy unlabelled path, exactly as the reviewer scoped it. Redesigning the predicate against a case this host cannot produce is how a guard gets loosened for a hypothesis.

## Bonus: this branch broke four of five census labels

Following the earlier comment — I checked all of them rather than the one I'd spotted:

```
Transport.swift:24   looksLikeTransportContainer   -> a bare "//" comment (also names the wrong FILE)
Transport.swift:165  playheadPositionGroupLabel    -> an entirely different function
Mixer.swift:309/:322 findVolumeFader/findPanControl-> a parameter line and "return described"
Mixer.swift:63       header pan slider             -> a doc comment
Tracks.swift:54      isTrackHeadersGroup           -> still correct
```

**All five were correct on `main`.** Every label already carries the function name, so the line number is dropped — the class removed rather than the instances chased.

## Verification

```
20 Issue628 tests pass · locator census at budget (59, unchanged) · ship gate running
```

MongLong0214:
## Measured: this removes the whole class, not the instances

```
origin/main                               9 baked  file.swift:NNN  strings in Sources/
origin/work/628-transport-ambiguity (before)   9
8f8c5f5b (this fix)                            0
```

Every remaining diagnostic names a **symbol** instead. Symbols are checked by the compiler; a line number in a string is checked by nobody, which is why four of five were wrong on this branch and three were wrong after a clean merge with #651.

Where a location genuinely is needed in future, `#file`/`#line` at the site are maintained by the compiler. A string pointing into *another* file has nothing that updates it.

MongLong0214:
```
SHIP GATE PASS @ 8f8c5f5b — 3895 tests · locator census at budget (59) · live evidence bound to head
live_628_counting_locator: 17/17 checks, 14 mutation-backed, all failure counters zero
```

The extraction was checked against `main` line by line: `transportControlKeywordHits` is `main`'s reduce body verbatim, plus this branch's false-friend guard and nothing else. The census budget did not move, so it added no blind lookup.

MongLong0214:
## Scope check on defect 1: what this fixes and what it deliberately leaves

Checked against `main` rather than assumed:

```
main:82   if withControls.isEmpty, labelled.count == 1 { return labelled[0] }   <- PRE-EXISTING
main      getTransportBar scans directly; it does NOT consult getControlBar     <- composition is ours
```

So the shell-returning branch is not this branch's doing. **The composition is**, and that is what is fixed.

The residual, recorded rather than fixed here: `findControlBarCheckbox` calls `getControlBar` and immediately searches it for a checkbox, so on a tree with exactly one labelled group holding none it returns nil where a working container may exist elsewhere. That is the same *name vs capability* shape — and it is pre-existing behaviour on `main`, reachable without any of this branch's changes.

Fixing it here would mean changing a path this branch does not touch, on a tree measured not to occur on this host (both live Control Bar groups are labelled and one holds twenty checkboxes, so `withControls.count == 1` answers first). That is worth its own change with its own evidence, not a rider on this one.

PR #651, opened 2026-08-21
feat(#628): the mixer fallback is observable, and measured not to fire

MongLong0214:
Second half of #628. #650 covers the transport scan; this covers the mixer strip, and it has to fix the live library before it can prove anything.

> Updated at `45fbb0e7` after an adversarial review returned **BLOCK** on nine defects. All nine are closed; details in the comments. This body describes the current head.

## What the code does

`findVolumeFader` ends `return sliders.first` and `findPanControl` ends `return sliders[1]`. Those indices are identity taken from tree order. When the description matches they are never reached — but nothing distinguished *"the description matched"* from *"position 1 happened to be pan"*, and only one of those is a fact about the strip. Both now say which happened.

**The element that comes back is unchanged.** The reviewer specifically tried to find an input where it differs and could not. Eight unit tests pin the selection either way.

Two of those eight exist because the review found the other six couldn't tell `described.first` from `described.last` — every fixture had exactly one described slider. The new pair was **watched to fail** under `.last`, and the original six watched to pass.

## Why this needed a library fix first

The instrumentation is `Log.info`, which writes to stderr (`Logger.swift:60`) — and the live driver launched the server with `stderr=subprocess.DEVNULL`.

**A log nothing can read is not observability.** A harness asserting on it would have passed for exactly the reason a broken one would. `Driver` now captures stderr to a *file*, not a pipe: nothing drains that stream while `_read` blocks on stdout, so a pipe would deadlock the server once the buffer filled — silently, and only on the runs that logged the most.

## The control, and why the first version of it was fake

`server_logged(needle, since=0)` returns `(present, channel_is_live)`.

- **`channel_is_live` requires an `[INFO]` line**, not merely a non-empty file. `Log` gates on `LOG_LEVEL`, which the driver *inherited* — at `warn`, every line this run asserts absent would vanish while one unrelated warning kept the file non-empty and the control satisfied. The level is now pinned in the child env.
- **`since` scopes the window.** Without it, the post-write check re-read the whole log and inherited the earlier read's silence. A check that *reads* as scoped and is not is worse than an unscoped one.

## The mutation was run, not claimed

`evidence.py` is explicit that a `mutation` string is normally the author's claim and a reviewer's job to judge. This one was applied, built, and executed — twice, the second time against the hardened checks:

```
clean run     11 checks · 11 passed
mutated run   11 checks ·  7 passed     isVolume AND isPan forced false
```

The four that flipped are exactly the four the mutation is named on. The scoped window proved **load-bearing rather than decorative**: under mutation the write check reported the fallback *inside* its 1666-byte window, so the offset catches a line the write itself emitted.

## Two vacuous passes this caught in its own harness

1. The first version drove a write that was **refused** (`track` is an index key, not a name) and its *"the arrangement did not move"* assertion passed **because nothing had happened**. The write now steers by `target_ref` — no ordinal — and the run asserts the fader moved, comparing the operation's own `observed_before`/`observed_after` rather than a poller value up to five seconds stale.
2. `rows > 0` was satisfied by exactly the state that makes the run vacuous: a strip with no sliders yields a row via `?? 0.0` and logs nothing. It now requires a non-default volume.

## Gate

```
SHIP GATE PASS @ 45fbb0e7 — 3895 tests · live evidence bound to head
live_628_mixer_fallback_is_visible: 11/11 checks, 2 captures, 1 visual, 1 recording, 2 ops driven
harness evidence coverage (#612): ok
```

## What this does not claim

That no Logic layout reaches the fallback. It says **this one does not**, on a strip whose sliders carry descriptions — the configuration the product is actually used in. The fallback is instrumented rather than removed, because a path nobody has seen fire is not a path known to be dead.

### DIFF
--- a/Scripts/check-ax-locator-census.py
+++ b/Scripts/check-ax-locator-census.py
@@ -60,7 +60,14 @@ import sys
 # taken). A census blind to the defect it was built for reported "at budget" while that site shipped.
 #
 # The twenty-two are not new code. They were always there; only the instrument changed.
-BLIND_SITE_BUDGET = 59
+#
+# 59 -> 57 (2026-08-21, #628). Two sites converted, so the ratchet takes the ground:
+# `findVolumeFader` and `findPanControl` no longer take a first match in the long-hand shape — they
+# build the candidate list and can say how many there were. This is the bar FALLING because the
+# defect was fixed, which is the only direction it is allowed to move. The check fails when the
+# count comes in UNDER budget too, on purpose: ground gained and not recorded is ground that the
+# next change can quietly give back.
+BLIND_SITE_BUDGET = 57
 
 SEARCH_ROOTS = ("Sources", "Scripts")
 
--- a/Scripts/livekit/evidence.py
+++ b/Scripts/livekit/evidence.py
@@ -191,9 +191,14 @@ class Driver:
         self._stderr_file = tempfile.NamedTemporaryFile(
             prefix="lpm-server-stderr-", suffix=".log", delete=False)
         self._stderr_path = self._stderr_file.name
+        # `Log` gates on LOG_LEVEL (`Logger.swift:69`), which the driver would otherwise INHERIT.
+        # At LOG_LEVEL=warn every `Log.info` vanishes while an unrelated warning still fills the
+        # file -- so a harness asserting absence would see a non-empty log and conclude the channel
+        # carried, when the only lines that could have carried were suppressed. Pinned, not inherited.
+        env = dict(os.environ, LOG_LEVEL="info")
         self.proc = subprocess.Popen(
             [self.binary], stdin=subprocess.PIPE, stdout=subprocess.PIPE,
-            stderr=self._stderr_file, text=True, bufsize=1)
+            stderr=self._stderr_file, text=True, bufsize=1, env=env)
         self._send("initialize", {
             "protocolVersion": "2024-11-05", "capabilities": {},
             "clientInfo": {"name": "livekit", "version": "1"}})
@@ -293,14 +298,21 @@ class Driver:
         except Exception:
             return ""
 
-    def server_logged(self, needle):
+    def server_logged(self, needle, since=0):
         """(present, channel_is_live) -- the second value is the control.
 
-        An empty log makes every `needle in log` False, so an absence assertion is vacuous until
-        something has come through. Callers assert on BOTH: absent AND the channel carried.
+        `since` is a byte offset from a previous `len(server_log())`, so a caller can ask about the
+        window AFTER something it did rather than about the whole run. Without it every later read
+        still sees the earlier lines, and "nothing was logged by the write" is indistinguishable
+        from "nothing was logged by the read either" -- the check reads as scoped and is not.
+
+        `channel_is_live` requires an `[INFO]` line, not merely a non-empty file. A log holding only
+        warnings proves the stream is connected and proves nothing about whether an INFO line could
+        have arrived, which is the level every caller of this asserts absence at.
         """
         log = self.server_log()
-        return (needle in log, bool(log.strip()))
+        window = log[since:] if since else log
+        return (needle in window, "[INFO]" in log)
 
     def close(self):
         try:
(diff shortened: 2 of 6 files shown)


MongLong0214:
## Adversarial review returned BLOCK; nine defects, all closed at `45fbb0e7`

A pre-merge review at high effort read the branch and blocked it. **The element-selection claim survived** — the reviewer could not construct an input where `findVolumeFader` or `findPanControl` returns a different element than before this diff. Everything else it found was real, and every one was in the *evidence*, not the product:

| # | defect | why it mattered |
|---|---|---|
| 1 | **The write-path claim was false.** `mixer.set_volume` reaches `findVolumeFader` via `findTrackHeaderVolumeFader`, but header pan goes through `findPanControlInHeader` — a different function this branch does not touch. | the check said "calls both locators"; it now claims only the one the write reaches |
| 2 | **Absence searched the whole log every time**, so the post-write reading saw the earlier read's silence | a check that *reads* as scoped and is not is worse than an unscoped one. `server_logged` takes a byte offset; the write is measured from a mark taken immediately before it |
| 3 | **The control only proved the file was non-empty.** `Log` gates on `LOG_LEVEL`, which the driver *inherited* — at `warn` every fallback line vanishes while an unrelated warning keeps the file non-empty | level pinned in the child env; the control requires an `[INFO]` line, not any byte |
| 4 | **`rows > 0` was satisfied by the exact state that makes the run vacuous** — a strip with no sliders yields a row via `?? 0.0` and logs nothing | now requires a non-default volume, i.e. the locator returned something |
| 5 | **"The write moved the fader" compared against a poller value up to 5s stale** | a no-op write could pass on cache lag; uses the operation's own `observed_before`/`observed_after` |
| 6 | **The absence reading happened before `stop_recording` waits out the recording**, during which the AX poller keeps cycling | a fallback in that tail was outside every assertion; a final reading is taken after the recorder stops |
| 7 | **The mutation string named only `isVolumeFader`** while the mutation actually run forced both — a mutation that cannot affect pan was attached to the pan checks | provenance now matches what was run |
| 8 | **Every described-slider fixture had exactly one match**, so `described.first` and `.last` were indistinguishable | two cases pin tree order; **both watched to fail under `.last`**, and the six existing cases watched to *pass* under it — confirming the gap was real |
| 9 | **Census labels still pointed at `Mixer.swift:309/:322`** | this branch moved those returns to `:327/:353`; the diagnostic named the wrong function |

Plus `Driver.close()` leaked — `delete=False` is required because the server holds the file open, which makes cleanup that function's job, and it wasn't doing it.

## The mutation re-run against the hardened checks

```
clean     11 checks · 11 passed
mutated   11 checks ·  7 passed
```

Four flipped, all four named on the mutation. The scoped window proved **load-bearing rather than decorative**: under mutation the write check reported the fallback *inside* its 1666-byte window, so the offset catches a line the write itself emitted instead of inheriting the earlier read's.

```
SHIP GATE PASS @ 45fbb0e7 — 3895 tests · live evidence bound to head
```

## One correction to the review

It said `set_volume` "does not call both changed locators" and that it calls only `findTrackHeaderVolumeFader`. The second half overstates: `findTrackHeaderVolumeFader` *does* call `findVolumeFader` (`AXLogicProElements+Mixer.swift:49`). The pan half was correct and is the half that made my claim false.

MongLong0214:
## Rebased onto #650 at `a65f506e` — and the ratchet moved

The rebase had exactly one conflict, in `MainEntrypoint.swift`, resolved by **dropping this branch's change entirely**: #650 converted the census site labels from `file:line` to symbol names, which supersedes this branch's `:309 → :327` edit. The one label #650 left carrying a filename now names its symbol like its five siblings.

## The census came in UNDER budget, which this check fails on purpose

With both halves of #628 in one tree:

```
BLIND_SITE_BUDGET  59 -> 57
```

`findVolumeFader` and `findPanControl` no longer take a first match in the long-hand shape — they build the candidate list and can say how many there were. That is the bar **falling because the defect was fixed**, which is the only direction it may move.

The check rejecting an under-budget count is the part worth naming: ground gained and not recorded is ground the next change can quietly give back.

## Gate

```
SHIP GATE PASS @ a65f506e — 3903 tests · live evidence bound to head
live_628_mixer_fallback_is_visible: 11/11 checks, all failure counters zero
```

Live evidence was regenerated after the rebase — the previous document was bound to a head that no longer exists. This is also the **first run of the mixer harness against a tree that contains the transport changes**, and it passes there.

PR #664, opened 2026-08-22
feat(#289): source priority in the cache commit rule

MongLong0214:
First increment of ADR-006 (#289): the cache commit rule learns **who is asking**.

## The gap, and why it was not hypothetical

`StateCache.accepts` implemented two of the ADR's four commit conditions — project epoch and section revision. A bare compare-and-swap is first-writer-wins, so two refreshes that capture the same revision race and the loser is discarded regardless of authority. When the winner is a slow background poll and the loser is a mutation verification, the cache keeps the older, less trustworthy read. That is the ADR's *"slow background polls must not overwrite mutation verification"*, inverted.

It was already written down as accepted behaviour: `Issue289StaleRefreshTests.concurrentWritesFromSameObservationAllowOnlyOne` asserts `surviving.tempo == 101 || surviving.tempo == 202`. **That test still passes unchanged.** This adds an ordering it never constrained rather than rewriting what it claimed.

## The rule

`RefreshSource` ranks the four sources the ADR names. A write may displace already-committed revisions only when **every commit in the gap** is strictly outranked — the question is about the gap, not the newest commit alone, or a background poll landing on top of an explicit read would let a subscription refresh erase a read it never outranked. Two clauses hold the edges:

- a revision whose author has been pruned from the bounded history is **not** displaceable — forgetting who wrote something is not evidence it was unimportant;
- a stale project epoch is never displaceable at any rank — that value describes a different project.

## A defect I introduced and caught before review

I defaulted the source-less overload to `.explicitRead`, then checked who actually calls these. Every caller today is `StatePoller` — a background loop by its own file comment. That default would have labelled every poll second-highest, so a poll could not be displaced by anything below a mutation verification: the exact mislabelling this rule exists to prevent, applied silently at every unconverted call site. The default is now `.backgroundPoll`, the bottom rank.

## Mutation-tested, and the first pass found two dead guards of mine

```
unknown revision is displaceable        0 -> 4 failing
epoch guard removed from the branch     0 -> 4 failing
equal rank displaces                         16 failing
source-less defaults to explicitRead          5 failing
```

The epoch one is the instructive failure: I had written a test for it, believed it covered the branch, and it never entered the branch at all — it advanced the epoch without advancing the revision, so the gap was empty and the function returned early. A test aimed at the wrong place looks identical to a test that passes.

## Scope — counted, not estimated

```
conditional-write overloads on StateCache   11
overloads taking an explicit source          4   (transport)
call sites still on the source-less form     4   (project, tracks x2, channelStrips)
```

The half-converted state fails safe: unconverted callers take the bottom rank and stay displaceable. But priority distinguishes nothing in those sections yet.

**Not done, and not measured:** `RefreshCoordinator` (single-flight, coalescing, deadlines, backpressure) does not exist; subscription backpressure is untouched; the benchmark matrix the ADR names (128/256, synthetic 1,000 tracks, 50 subscriptions) needs fixtures that do not exist.

Behaviour is unchanged when `FeatureFlags.adr006VersionedCache` is off — this path is the conditional write itself, not the flag-gated envelope.

### DIFF
--- a/Sources/LogicProMCP/State/StateCache.swift
+++ b/Sources/LogicProMCP/State/StateCache.swift
@@ -31,6 +31,13 @@ actor StateCache {
     private var projectEpoch: UInt64 = 0
     private var sectionRevisions: [CacheSectionID: UInt64] = [:]
     private var droppedStaleWriteCounts: [CacheSectionID: UInt64] = [:]
+    /// Which source produced each committed revision, per section. ADR-006's commit rule has to
+    /// answer "was everything committed since my observation lower-priority than me?", and that is
+    /// a question about the commits in the GAP, not about the newest one alone: a background poll
+    /// landing on top of an explicit read would otherwise let a subscription refresh erase the read
+    /// it never outranked. Bounded — entries at or below the observable window are pruned.
+    private var revisionSources: [CacheSectionID: [UInt64: RefreshSource]] = [:]
+    private static let revisionSourceWindow: UInt64 = 64
 
     /// Whether Logic Pro has an open document with a visible window.
     /// Defaults to true (optimistic) — StatePoller sets to false when no document detected.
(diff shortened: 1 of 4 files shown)


MongLong0214:
**Withdrawing this — the mechanism is unsound, and it makes correctness worse than `main`.**

The review found the losing interleaving and it holds:

```
1  mutation verification V captures rev 0, reads tempo 140, is delayed
2  poll P also captured rev 0 but reads LATER, after a real change to 160,
   and commits 160 as rev 1
3  V resumes with observed=0; rev 1 is .backgroundPoll, so my rule accepts
   and overwrites the newer 160 with the stale 140
```

The existing compare-and-swap rejects step 3. **Rank says how trustworthy a read is, not when it was observed**, and I substituted one for the other.

`StatePoller` even states the invariant I broke, in a comment I read past:

> `// This must happen before `execute`: observing after the AX read returns would make a stale result look current and reintroduce the overwrite race this guard is meant to prevent.`

Revision equality *is* the freshness proof in this design. My displacement branch discards that proof. No amount of patching the rank comparison fixes it, so this is a withdrawal rather than a revision — a correct version needs what ADR-006 actually specifies: an observation generation, or the `RefreshCoordinator` with cancellation, so that "newer" is a fact about the read rather than an inference from the reader's identity.

Also confirmed against me, and worth recording separately: `advanceSectionRevision`'s global `.backgroundPoll` default labels every *unconditional* commit as background, including explicit resource reads and mutation-verification readbacks in `TransportDispatcher`. And none of the new logic consults `FeatureFlags.adr006VersionedCache`, so the changed acceptance semantics run with the flag off — not behaviour-preserving.

One reviewer sub-claim I checked and did not accept: mutating only the source-less default (leaving the poller call site alone) does fail 5 tests, so the commit's mutation figure was measured correctly. The narrower point stands — `sourcelessWritesAreNotTreatedAsAuthoritative` alone cannot pin the default, because mutation verification outranks `.explicitRead` too.

`main` is unaffected; nothing here was merged. Findings carried to #289.

PR #669, opened 2026-08-23
fix(#668): refresh_cache moves to the medium deadline class

### LINKED ISSUE #668: system.refresh_cache exceeds its own 25s deadline on projects around 74 tracks
**Next action: decide whether to raise `refresh_cache`'s deadline class, split the poll cycle per section, or both — it is a design call; the measurement is done.**

## What happens

On a 74-track project, `system.refresh_cache` takes a **median of 12.6s** and **hits its 25s deadline in roughly 1 call in 5**, returning `operation_timeout`. The failure is honest — no false `refreshed: true` — but the operation simply does not complete.

Shipped in v3.14.0:

```
OperationRegistry.swift:648   (.systemRefreshCache, "refresh_cache", readOnly, DeadlineClass.short, ...)
OperationRegistry.swift:190   case .short: 25
```

## Measured, on this machine

10 paired samples at a confirmed 74-track fixture:

```
      transport read   full cycle
  1        3202          11307
  2        2303           4005
  5        1921          13360
  6       26689          26676   <- ceiling
  7       12005          26674   <- ceiling
  8        4657          20293
 10        2046          12597

cycle median 12597 ms   (min 4005, max 26676)
```

`26674 / 26676 / 26689` recur across different operations: that is the 25s deadline plus round-trip, not load. **Samples at the ceiling are censored** — the true cost is ≥ that, so the median understates it.

Scaling, same fixture family:

```
 1 track    cycle ~= transport read alone
40 tracks   cycle 3605 ms, non-transport portion inside the noise
74 tracks   cycle 26668 ms, non-transport portion ~21s
```

The cost is **not** the transport section (which does its own live AX read per call, 0.7–12s); it is everything else in the cycle, and it appears between 40 and 74 tracks.

## Why it matters

`refresh_cache` is public surface. A user with a normal-sized session gets a timeout on a read-only operation one time in five, and the failure rate rises with project size. Nothing here is exotic — 74 audio tracks is a small project by Logic standards.

## Reproduction

```
1  build a fixture with tracks.create_audio (expect sporadic operation_timeout;
   ~27 creates per 653s with pacing)
2  confirm logic://tracks reports readable == true and the count you expect
3  time system.refresh_cache repeatedly; watch for the ~26.67s ceiling
```

Found while measuring ADR-006 (#289); recorded there in detail. Filed separately because it is a defect in shipped behaviour rather than an ADR line item, and it should not depend on that ADR's design decisions to be fixed.

MongLong0214:
Mitigates #668. **One functional line: `refresh_cache` moves `DeadlineClass.short` (25s) → `.medium` (90s).**

## Why

`refresh_cache` drives a full poll cycle. On a 74-track project that cycle costs a **median of 12.6s**, and roughly **1 call in 5 hits the 25s ceiling** and returns `operation_timeout`. It is a read-only operation on the public surface; leaving it failing one time in five while the cause is investigated is not the right trade.

## 90s is a choice, not computed headroom — and the code says so

Every 74-track sample at ~26.67s was **censored** by the old deadline: `26674 / 26676 / 26689` recur across different operations, which is a ceiling, not load. The true maximum is unknown, so `median × N` would be arithmetic on a truncated distribution.

I could not measure an uncensored maximum — the fixture had already been removed from the owner's project, and rebuilding it costs ~11 minutes and alters a real Logic document. So the class is deliberately generous and labelled as a choice, both in the registry comment and in the commit message.

## This does not close #668

The cycle still costs what it costs; a large enough project exhausts 90s too. **What closes the issue is splitting the poll cycle per section**, sequenced after the #289 coordinator decision because the two overlap. Stated here explicitly so the issue is not closed on the deadline bump.

## Verification

```
registry test pins the deadline class per operation, so both moved together
reverting the source alone           -> 5 assertions fail
ship gate @ 57be38fb                 -> 8/8, full suite 3922 tests
live evidence @ 57be38fb             -> 11/11, zero failure counters
```

## Two things the gate caught that I had missed

```
C1a  the commit message contained "session" — a token the public-surface filter rejects.
     I meant a Logic session; the filter is right that it reads as internal process
     metadata on a public commit. Reworded to "a real Logic document".
C3a  Package.resolved had drifted back in during branch switching.
```

Amending for C1a moved the head, which invalidated the live evidence bound to the previous one — the gate then failed with `no passing live evidence bound to 57be38fb` and pushed nothing. Correct behaviour; evidence regenerated at the new head.

### DIFF
--- a/Sources/LogicProMCP/Server/OperationRegistry.swift
+++ b/Sources/LogicProMCP/Server/OperationRegistry.swift
@@ -645,7 +645,18 @@ enum OperationRegistry {
     } + ([
         (.systemHealth, "health", Mutability.readOnly, DeadlineClass.short, VerificationPolicy.none, []),
         (.systemPermissions, "permissions", Mutability.readOnly, DeadlineClass.short, VerificationPolicy.none, []),
-        (.systemRefreshCache, "refresh_cache", Mutability.readOnly, DeadlineClass.short, VerificationPolicy.none, []),
+        // #668 — `.short` (25s) was wrong for this one. `refresh_cache` drives a full poll cycle,
+        // whose cost grows with project size: measured median 12.6s on a 74-track project, with
+        // roughly 1 call in 5 hitting the 25s ceiling and returning `operation_timeout`.
+        //
+        // `.medium` (90s) is NOT a computed headroom figure. The 74-track maximum is CENSORED —
+        // every sample at ~26.67s was cut by the old deadline, so the true maximum is unknown and
+        // any multiple of the median would be arithmetic on a truncated distribution. This is a
+        // deliberately generous class chosen so the operation can finish, not a measurement.
+        //
+        // This is mitigation, not a fix: the cycle still costs what it costs, and a large enough
+        // project will exhaust 90s too. Splitting the poll cycle per section closes #668.
+        (.systemRefreshCache, "refresh_cache", Mutability.readOnly, DeadlineClass.medium, VerificationPolicy.none, []),
         (.systemListRecentTraces, "list_recent_traces", Mutability.readOnly, DeadlineClass.short, VerificationPolicy.none, ["limit"]),
         (.systemGetTrace, "get_trace", Mutability.readOnly, DeadlineClass.short, VerificationPolicy.none, ["trace_id"]),
         // WHY: trace deletion destroys internal evidence, not Logic state, so it stays readOnly and bypasses the Logic-write gate.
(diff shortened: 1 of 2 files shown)


MongLong0214:
**Withdrawing — the mitigation does not work end-to-end, and review found why.**

Raising the server deadline to 90s does not help any real caller, because every first-party client budget is shorter and unchanged:

```
Scripts/live-e2e-test.py:36              strict mode 40s  -> returns None at 40s (untyped)
QualificationTransport.swift:475/:1431   default 10s      -> SHORTER than the 12.6s median
session bootstrap callers                5s hard-coded
```

Verified independently, not taken on the reviewer's word.

**So the change trades a typed `operation_timeout` at 25s for an untyped client timeout at 40s (or 10s).** That is worse than the status quo: the caller loses the error code and waits longer to lose it. And qualification at 10s means a large-project refresh fails there before the new server deadline could ever apply — the 12.6s median is already over that budget.

I had asked the reviewer specifically whether raising the deadline makes any caller worse off (R4), because it was the obvious counter-cost of my own change. It does, and I had not checked the client side myself before opening the PR.

**#668 is not fixed and stays open.** What a working mitigation would require is raising the client budgets in step with the server class — which touches the E2E harness, the qualification transport, and bootstrap callers — or a bounded partial-refresh that returns within the existing budgets. Both are larger than a one-line class change and neither should be done without deciding which is right.

Nothing was merged; `main` is untouched. Findings carried to #668.

Issue #412, opened 2026-07-19
Saga lifecycle is not bounded end-to-end: wedged step can strand the journal and pin the mutation gate


MongLong0214:
## Debt record (follow-up to PR #378's deadline-budget preflight)

PR #378 added a preflight that rejects a saga plan whose worst-case step time cannot fit the command deadline budget. The runtime enforcement, however, remained coarse: only the transport-level timer bounds the client response, while the saga's real lifecycle — blocking step dispatch, compensation/unwind, the terminal journal write, and replay-unlock — runs in detached work that can outlive the response.

### Failure modes
- A step wedged in a synchronous AX call leaves the journal `inProgress` forever: every replay of that idempotency key is refused and the replay lock never clears.
- The saga's `releaseOnly` mutation-gate claim is never reclaimed, locking out every future saga until process restart.
- Compensation has no deadline checkpoint, so inverses can keep dispatching after the client already received a timeout — or, if cooperative cancellation suppresses them, abandoned rollback is misreported as `compensationFailed` instead of honest uncertainty.
- The timeout body can under-report a write that was already in flight.

### Required properties
1. No new step or inverse dispatch after the lifecycle deadline is observed (an already-in-flight synchronous call is not interruptible and must be reported fail-closed).
2. The journal always reaches a terminal record within the lifecycle budget — never stuck live.
3. The gate becomes reclaimable after a bounded grace on timeout — never released immediately (would race orphaned work), never pinned permanently.
4. The client response body always equals the journal terminal record.
5. Abandoned compensation is reported as rollback-uncertain, never as a false compensation failure; the timeout envelope is fail-closed (`write_attempted`/`write_boundary_crossed` true, `safe_to_retry` false).

PR #670, opened 2026-08-23
feat(#668): a command can see its own deadline

MongLong0214:
Enabling increment for #668. **This changes no observable behaviour on its own, and that is deliberate.**

## What it does

`runWithDeadline` computes a per-command budget and throws it away — the work closure never learns how long it has, so it cannot stop early.

```
OperationTraceContext.deadline        new TaskLocal, bound INSIDE the detached work task
OperationTraceContext.remainingBudget derived; clamps to zero, never negative
LogicProServer                        binds it from the same `deadline` the timeout races against
```

Bound inside the detached task for the reason the trace context already documents there: `Task.detached` severs TaskLocal inheritance, so a value bound in the caller is invisible below it.

## Why cancellation is not enough — this is the part that killed two earlier attempts

Cancellation *does* reach the work task (`workTask.cancel()` at the deadline). But `DeadlineRace` is first-past-the-post:

```
LogicProServer.swift:928   private var resumed = false
LogicProServer.swift:936   if resumed { return false }     <- the second finisher is discarded
```

By the time cancellation lands, the timeout already owns the continuation. **A partial result computed after that point is thrown away.** So "make `pollOnce` check `Task.isCancelled`" produces code that is correct in isolation and does nothing observable.

Two withdrawn attempts on this issue, for the record:

```
#669   raise the deadline class 25s -> 90s
       refuted: the shortest client budget in this tree is 3s; live-e2e strict is 40s;
       qualification is 10s — already BELOW the 12.6s median it was meant to accommodate
2nd    check Task.isCancelled inside the poll
       refuted before writing it, by reading the race above
```

Neither touched why the cycle cannot stop early: **it does not know when early is.**

## Mutation-tested

```
injection removed (declared but never bound)   3 failing
zero clamp removed (negative budget allowed)   3 failing
```

The first mutation is the one that matters. `#289` has a `VersionedSnapshot` type with zero constructors — a design translated into types and never connected — and this could have shipped as the same shape.

## What this does NOT do

It does not make `refresh_cache` return within its deadline. Nothing consumes the budget yet.

The next increment makes `pollOnce` return what it completed — and it has a constraint found by reading the code rather than by writing it: `hasDocument = projectReady || tracksReady`, so a budget check placed **before** those two sections would let "ran out of time" enter the same branch as "both reads failed", and the poller could conclude the document is closed. The checks must sit after that decision, guarding transport/mixer/markers only.

Gate: `SHIP GATE PASS` at `540efe75`, full suite 3925 tests, live evidence 11/11.

### DIFF
--- a/Sources/LogicProMCP/Server/LogicProServer.swift
+++ b/Sources/LogicProMCP/Server/LogicProServer.swift
@@ -805,8 +805,15 @@ actor LogicProServer {
                 }
             )
             let workTask = Task.detached(priority: .userInitiated) {
-                let result = await OperationTraceContext.$current.withValue(traceContext) {
-                    await work()
+                // #668: bind the budget INSIDE the detached task, for the same reason the trace
+                // context is bound here — `Task.detached` severs TaskLocal inheritance. Bound from
+                // the same `deadline` the timeout task races against, so work that consults it is
+                // reasoning about the identical instant rather than an approximation of it.
+                let budgetDeadline = ContinuousClock().now.advanced(by: .milliseconds(Int(deadline * 1000)))
+                let result = await OperationTraceContext.$deadline.withValue(budgetDeadline) {
+                    await OperationTraceContext.$current.withValue(traceContext) {
+                        await work()
+                    }
                 }
                 if let heldClaim { heldMutationGate?.release(heldClaim) }
                 let didWin = race.resume(continuation, returning: result)
--- a/Sources/LogicProMCP/Server/OperationTrace.swift
+++ b/Sources/LogicProMCP/Server/OperationTrace.swift
@@ -77,6 +77,32 @@ final class OperationTraceWriteBoundaryArm: @unchecked Sendable {
 /// nested scope with a fresh box carrying the parent's trace ID so child
 /// traces correlate instead of clobbering the parent registration.
 final class OperationTraceContext: @unchecked Sendable {
+    /// #668 — the deadline the enclosing command is racing against, as an absolute instant.
+    ///
+    /// `runWithDeadline` computes this and then discards it: the work closure never learns its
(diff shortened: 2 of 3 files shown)


MongLong0214:
**Withdrawing. The central claim in my own comment and commit message is false, and review proved it.**

I wrote that the budget is *"derived from the same `deadline` the timeout task races against, so anything consulting it reasons about that instant rather than an approximation."* It is not:

```
timeout timer   DispatchQueue.asyncAfter(deadline: .now() + scheduleAfter)   caller thread, DispatchTime
my binding      ContinuousClock().now.advanced(by: .milliseconds(...))       inside the detached task
```

**Two different clocks, read at two different moments. There is no shared absolute instant.** If the work task starts late, the budget runs past the timeout; if the caller is delayed between launching the task and scheduling the timer, the timeout can fire before work even begins while work then observes nearly its full budget. The #412 `outerAbsoluteDeadline` path is deterministically wrong: the timer uses that precomputed instant and my binding still uses `now + deadline`.

`Int(deadline * 1000)` also truncates, so a sub-millisecond override binds an already-expired budget against a timer that has not fired.

**And the tests were built so none of this could show.** `workSeesItsBudget` reads the budget immediately after production computes `now + 5s`, so the 4–5s assertion is close to tautological — it never observes the timeout target at all. My mutation report is also wrong: I recorded "3 failing" for both mutations, but each mutation can only fail one test, and I did not check which.

R4 was the one thing I got right: the path into `StatePoller.pollOnce` is intact, no Task boundary, confirmed independently by the review.

**Correct shape, for the next attempt:** compute ONE absolute instant before launching either side, hand the same value to the timer and to the TaskLocal, and cover `outerAbsoluteDeadline`. The test has to compare the budget against the instant the timeout actually uses, not against a freshly recomputed one.

Nothing merged; `main` untouched. Findings carried to #668.

Issue #293, opened 2026-07-10
ADR-010: MIDI readback decision record — implementation consolidated into #302


MongLong0214:
# ADR-010 — MIDI readback provider decision and implementation hand-off

> **SSOT revision: 2026-09-21. SUPERSEDED as an execution issue; NOT a completed MIDI readback feature.** All remaining implementation and qualification requirements are explicitly owned by #302. Closing this duplicate queue does not waive any of those requirements. #308 is the programme index.
>
> Audit baseline: `d067e91958ad3ff1a1bdfb1e6e39a3e23e786517`. This update reconciles source/history; no new live acceptance run is claimed.

## Decision retained

Use independent observation of the actual Logic MIDI region, not the input sequence/cache that the server sent. The selected first observation route is Event List AX. The initial expected root is pre-committed authored intent, bound to the import/write before Event List observation. These are two separate trust roles.

[The shared scope decision](https://github.com/MongLong0214/logic-pro-mcp/issues/302#issuecomment-5060956361) chose those roles. It did not establish a completed production provider. Controlled export, a project-package parser and playback capture remain alternative investigations, not automatic fallbacks or extra providers an agent must build now.

## What was implemented or observed

- The completeness and independent-verification groundwork associated with #445/#447 exists, followed by the Event List collector/provider work including #517.
- `--probe-event-list` can observe note rows from a release artifact without granting a production identity/independence claim.
- The later #302/#881 investigation records actual note headers and values, en-US/ko-KR observations and virtualization behavior; it supersedes claims that the note columns simply do not exist.
- The region-identity and coordinate-origin investigations produced useful constraints, but the draft `docs/tickets/issue-293-live-region-identity/STATUS.md` and `issue-293-lri1-region-origin/STATUS.md` are not evidence that their proposed production changes landed.
- At the audited source, `RegistryResolvedIdentityProof` still has only a debug `FAULT_TEST_SEAM` mint. The production resolver and complete public provider path are not finished.

## Exact transfer to #302

| Requirement previously carried here | New authoritative location |
|---|---|
| Selected/requested region identity, complete candidate set, non-positional track association | #302 M1 and target-negative acceptance |
| Event List ownership and observation-scoped freshness | #302 M1/M3 |
| Exact origin, absolute-versus-region-relative timing, PPQ and meter interpretation | #302 M2 |
| Header/filter/count/end/virtualization completeness, stable harvest and duplicate preservation | #302 M3 |
[...]

PR #711, opened 2026-08-29
fix(#293): the Event List readback could not run in any language but English

MongLong0214:
The roadmap said #293's next step was rebuilding the collector's expected shape from a measured tree. **That was already done** — the per-column reading matches the 2026-08-10 measurement exactly, flag cells included. Running the harness is what found the real blocker, and it was not the shape.

## The product could not start

```swift
AXHelpers.getDescription(tab, runtime: runtime) == "Event"
```

Logic runs in Korean on this machine and calls that tab `이벤트`. So `EventListReadbackCollector` threw `eventTabNotFound` **before reading anything** — a readback that cannot start, rather than one that reads wrong. It goes through `AXLocalePolicy.eventListTab` now, declared with the four tabs measured beside it: `이벤트`, `마커`, `템포`, `조표 및 박자표`.

**Scope of that claim, stated exactly:** English and Korean are proven by a live run. Japanese is declared and unproven. A finite alias table is not locale-independence, and this PR does not claim it is.

## The first fix for the harness was too large, and review said so

`located_band` **already** translated names through `AX_REGION_LABELS`. Every region in that table had its row and `Control Bar` did not. **One missing line, not a missing mechanism** — the parallel alias apparatus this branch first grew has been removed again.

What remained was a third copy of the same idea: the harness shelled out to the band tool itself, so it never saw the table. With the Korean row added, the shared helper found the bar and the harness still reported nothing — asking a different question with the same words. It delegates now.

## The other bindings were real

| where | what was wrong | observed |
|---|---|---|
| View menu | found by **position** (`menu bar item 9`) and English item name | actually `보기` / `목록 편집기` |
| the eight columns | asserted as English literals | `['L','M','위치','상태','채널','번호','값','길이/정보']` |

The column case is the inverted one worth naming: **the product had already read the Korean table correctly and returned both notes** when that assertion failed. My first diagnosis — "the collector's header binding is English-only" — was wrong; `readHeaders` compares through LabelSets. The English binding was in the harness.

Each rendered title is now compared against the LabelSet its column declares, parsed from the policy source. That parser **strips Swift comments first**, because a commented-out variant is not a declared one.

## A false clean, closed

`tab_state` covers all four list tabs, not the two this run drives. Narrowing it made a session that started on Tempo or Signature read as *"no list tab is selected"* — nothing to restore, restoration skipped, run reports itself clean.

## The guard, aimed at the copy that exists

`check-livekit-locale-aliases.py` checks `AX_REGION_LABELS`, which is what the locator reads. Its first version checked two lists written for the guard itself while the locator went on using another — a guard aimed at nothing.

Comparison splits on script, and the reason is not stylistic:

- an **ASCII** variant differs from what Logic renders only in case, and the product folds case → case-insensitive
- `컨트롤 막대` is **a whole language** the product claims to know, and nothing folds it into `Control Bar` → verbatim

Folding everything would have accepted `"CONTROL BAR"` as covering `"control bar"` while Logic renders `"Control Bar"` and the tool matches neither. Comparing everything exactly fails on today's tree, because the policy's variants are **normalisation inputs, not strings Logic emits**.

Split that way it found a real gap: the policy declares four ASCII spellings of the rail and the table tried one.

## Result

```
10 checks · 10 passed · cached_reads_used_as_live 0 · exit 0
```

Mutation-tested in both directions: removing the Korean row reproduces the original failure, and folding case for a non-ASCII spelling makes the guard accept a script it cannot reach.

## Deliberately not here

`live_628_counting_locator` has the same shape of binding, but its **captures do not settle**: the rail it photographs carries level meters that blink, and the display check calls its band not wholly within one screen of three. That is an environmental wall with a recorded shape, and carrying it would mean shipping a harness this branch cannot prove.

ADR-010's body — `assessReadback`, the public provider — stays closed. This goes as far as *the collector actually reads*.

Refs #293.

### DIFF
--- a/Scripts/check-livekit-locale-aliases.py
+++ b/Scripts/check-livekit-locale-aliases.py
@@ -1,0 +1,121 @@
+#!/usr/bin/env python3
+"""The live kit's locale aliases must not drift from `AXLocalePolicy`.
+
+`evidence.py` carries `CONTROL_BAR_NAMES` and `TRACK_HEADER_NAMES` because the harnesses cannot
+import Swift. That makes them a SECOND COPY of a label list, and a second copy goes stale the day
+someone adds a locale to the first — silently, because a missing alias does not look like a bug: the
+band lookup answers "no element with that exact AXDescription", the harness fails a precondition
+about a window frame, and nothing in that message says the word it wanted was spelled for another
+language.
+
+Measured 2026-08-29, which is why this exists: three harnesses passed `"Control Bar"` to the band
+tool on a Logic running in Korean. Each failed at a precondition unrelated to its own subject, so
+none of them could produce evidence at all on this machine — and under the source-coverage rule a
+harness that cannot run leaves its subject unproven however good its checks are.
+
+A copy with a check is a copy that cannot go stale. This is that check.
+
+WHAT IT DOES NOT DO: it does not require the two lists to be EQUAL. The policy is normalised and
+matched leniently by the product; the live kit compares exactly, so it legitimately carries the
+cased spellings Logic actually emits (`Tracks header`) beside the lower-cased policy variants. The
+rule is one-directional and that is the direction that matters — every policy variant must be
+reachable from the live kit, so a locale the product learned is a locale a harness can find.
+"""
+import ast
+import os
+import re
+import sys
+
+REPO = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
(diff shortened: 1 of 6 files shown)


MongLong0214:
Superseded by #712.

The commits were amended while acting on review — the parallel alias apparatus came back out, the harness's third copy of the band lookup was found and removed, and `check-ax-locator-census.py` asked for its budget to be lowered. Updating this PR would have needed a force push, which is a call I route to the owner rather than make, so the work moved to a fresh branch instead. Nothing is lost: #712 carries every commit plus the review fixes.

PR #714, opened 2026-08-29
fix(#293): the run's own tidying was inside the bracket it was judged by

MongLong0214:
Two changes, and the second was found by making the first.

## The harness declares the one file it drives

The source-coverage rule obliges a harness when a branch touches a path that harness claims. This adds the second declaring harness — three claimed paths across two.

**One file, and the restraint is the point.** `--probe-event-list` calls `observeNoteTable`, which drives `findEventTab` → `findEventPaneAndTable` → `readHeaders` → `readRows` → `readRow`: the whole collector and nothing else in the module. Claiming `MIDIReadback/` entire would cover `MIDINoteCanonicalizer`, `MIDIRegionDiff` and seven more this run never touches, and **a claim wider than the drive is how a coverage rule starts lying on its owner's behalf.**

`live_628_counting_locator` gets none, for the same reason inverted: it drives `censusDescendant`, one function in a 1300-line policy file that changes whenever a label is declared.

Checked rather than assumed — editing the collector obliges the harness, **deleting** it obliges it too, and a sibling nobody claims stays unproven:

```
claimed  : 9 of 213
unproven : 204
```

That number is printed on every run. It is a floor, not a proof, and it is the honest size of the gap.

## Re-running the harness found a false failure

Adding the declaration meant running it again — from a **closed** pane this time, which is the state a machine is normally in and not the one the earlier clean runs were taken from. They were clean because the pane happened to be open already.

Two symptoms, one cause:

- the restore ran **before** the `after` shot, so closing the pane changed the layout between the two frames the visual compares, and `the-transport-is-undisturbed-by-the-probe` went red **for the restoration rather than for the probe**
- the restoration predicate contradicted itself in exactly that case: when the run opens the pane it also closes it, so `final_state` is empty by design — while the second clause demanded the remembered tab still read selected, **a selection inside a closed pane**

The detail string was worse than the predicate. It asserted *"the pane is back to open=False"* from `pane_was_open`, a value read at the **start** — so the sentence claimed a restoration it had not measured while the flag beside it said the restoration failed.

## Verified from both starting states

```
pane CLOSED at start  ->  10 checks, 10 passed, restorations_failed 0, visual_failed 0, exit 0
pane OPEN   at start  ->  10 checks, 10 passed, restorations_failed 0, visual_failed 0, exit 0
```

The verdict no longer depends on whether somebody left a panel open — which is the mirror of the false-*clean* a review found in this same harness earlier today, and the reason both starting states are run rather than the convenient one.

Refs #293.

### DIFF
--- a/Scripts/livekit/live_293_the_collector_reads_a_real_note_row.py
+++ b/Scripts/livekit/live_293_the_collector_reads_a_real_note_row.py
@@ -41,6 +41,18 @@ import time
 sys.path.insert(0, os.path.dirname(os.path.abspath(__file__)))
 import evidence as E  # noqa: E402
 
+# What this harness proves, for `harness_evidence_coverage.py`. A branch touching this path must
+# run it and produce a clean document.
+#
+# ONE file, deliberately. `--probe-event-list` calls `observeNoteTable`, which drives
+# `findEventTab` -> `findEventPaneAndTable` -> `readHeaders` -> `readRows` -> `readRow` — the whole
+# collector and nothing else in the module. Claiming `MIDIReadback/` entire would cover
+# `MIDINoteCanonicalizer`, `MIDIRegionDiff` and seven more this run never touches, and a claim
+# wider than the drive is how a coverage rule starts lying on its owner's behalf.
+COVERS = [
+    "Sources/LogicProMCP/MIDIReadback/EventListReadbackCollector.swift",
+]
+
 WT = sys.argv[1] if len(sys.argv) > 1 else ""
 HEAD = sys.argv[2] if len(sys.argv) > 2 else ""
 if not WT or not HEAD:
--- a/Scripts/livekit/live_293_the_collector_reads_a_real_note_row.py
+++ b/Scripts/livekit/live_293_the_collector_reads_a_real_note_row.py
@@ -386,14 +386,6 @@ ev.check("293/the-flag-cells-really-are-childless",
          f"children={children!r} L={first.get('L')!r} M={first.get('M')!r}",
          GUARD_MUTATION)
 
-# Put the pane back the way it was found: the tab first, then the pane itself.
-if selected_before and selected_before != "Event":
-    tabs(select=selected_before)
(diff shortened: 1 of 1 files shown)


MongLong0214:
### 나머지 하네스에도 같은 형태가 있는지 훑었습니다 — 없었습니다

이 버그(뒷정리가 visual 브래킷 안)가 다른 데도 있는지 보려고 grep 휴리스틱을 썼고, **휴리스틱이 먼저 틀렸습니다.**

첫 시도는 *첫* 토글과 *마지막* 촬영을 비교했는데, 첫 토글은 **여는** 쪽이라 잘못된 것을 재고 있었습니다. 고쳐서 *마지막* 토글로 다시 재니 셋이 걸렸습니다:

```
live_549_cell_does_not_veto.py     lastToggle@107  lastShot@209
live_565_empty_marker_list.py      lastToggle@160  lastShot@286
live_608_first_call_is_not_refused.py  lastToggle@164  lastShot@198
```

**셋 다 읽어봤고 셋 다 오탐입니다.**

- `live_608` — 토글은 negative control 로 Save 패널을 **여는** 것이고, 닫는 건 `key code 53` 입니다. 촬영 **전에** 일어납니다. 제 grep 은 메뉴 클릭만 봐서 그 해제를 못 봅니다.
- `live_549`, `live_565` — 토글은 여는 쪽이고 `ev.restored` 는 마지막 촬영 **뒤에** 옵니다.

그러니 이 형태를 가진 건 #293 하나였고, 이 PR 이 그걸 고칩니다.

적어두는 이유: **grep 이 셋을 가리켰고 읽어보니 0이었습니다.** 읽지 않고 "세 하네스에 같은 버그" 라고 적었으면 그건 제가 잰 것보다 넓은 주장이었을 겁니다. 휴리스틱은 여는 토글과 닫는 토글을, 그리고 메뉴 해제와 키 해제를 구분하지 못합니다 — 그래서 후보 목록이지 결과가 아닙니다.

Issue #290, opened 2026-07-10
ADR-007: Selector Atlas — preserve shipped selectors; finish restored qualification integration


MongLong0214:
# ADR-007 — Preserve implemented selectors; complete the restored qualification adapter

> **Execution SSOT r2 — 2026-09-21. OPEN / selected integration remainder. `status: ready` means the bounded evidence-adapter work can start; it does not mean live qualification or release activation is complete.** #373 owns the common evidence/verifier/release interface; #284 owns parent acceptance and #308 scheduling. Source checkpoint: `1b36b90bb8f339c82a3d606c2f44cab08ef18da6`.

## 1. Why reopened, and what not to rebuild

[The August 30 receipt](https://github.com/MongLong0214/logic-pro-mcp/issues/290#issuecomment-5465777294) records adopted selectors, sanitized fixtures, operation mapping and a qualification-diff connection from #700–#717. That implementation is real.

The qualification subsystem was then removed and restored. Current `AtlasQualification.caseFor` documents that normal invocations do not arm its diff and that its emitted case carries `evidenceFiles: []`. Simply setting the flag would run into the verifier's evidence-cardinality requirement. Restoration did not establish unconditional current release enforcement.

**Reopening owns this specific integration gap, not an atlas rewrite.** Production selector adoption is separate from opt-in qualification-diff configuration. Preserve working runtime selectors even when qualification is not being run.

## 2. Original seven criteria, explicitly accounted for

| Original acceptance | Maintained / remaining owner |
|---|---|
| Desktop en/ko baseline fixtures | Historical five-fixture delivery retained; qualify actual required selected scopes/axes through A2 below. Creator was removed from product scope, not silently omitted. |
| Duplicate candidates do not become first-match mutation | Existing selector contract, regression-tested here. |
| Unknown/unsupported UI signature refuses unsafe mutation | Existing contract, maintained here. |
| Selector drift maps to affected operations | Existing mapping; preserve pan-versus-volume and record-arm-versus-global-record distinctions. |
| Fixtures contain no private user names/content | Preserve sanitization and verify new captures. |
| Geometry alone cannot produce high confidence | Preserve existing evidence/scoring constraint; no new scoring framework. |
| New-version qualification includes diff and refusal blocks promotion | **A1–A3 here**, current remaining integration through #373. Not complete merely because the case type exists. |

[...]

PR #716, opened 2026-08-29
fix: the arrange window's title was English, and I called that an environmental wall

MongLong0214:
`logic_window` defaulted to the substring `"Tracks"`. Logic names the arrange window after the current view, so a Korean Logic says `… - 트랙` and the lookup returned `None`. `shot` then took its no-window branch and recorded three things:

```
settled: false      region: null      wholly_within: false
```

Those are the symptoms of a capture that **never happened**.

## I read them as a settling problem

And wrote so — in a commit message and in a pull request — that the rail carries blinking level meters and does not fit one screen of three. That was **recalled, not measured**. The rail band is `(603, 162, 325, 1465)` against a 1050-tall window, so the clip is real; it is also irrelevant, because there was no window to clip against.

The measurement that settled it took one line: the capture record said `region: null` while the visual beside it carried a region, which cannot both be true of a rail that resolved. It resolved. The window did not.

## Eighteen harnesses take that default

On this machine none of them could capture anything, and each recorded it as an unsettled capture rather than as a missing window. The default is now every spelling of the arrange window this repository has measured; a caller naming a specific window still gets exactly that.

## `live_628` also still had English container names

Reverted out when an earlier branch was split, and never put back: the rail for `--probe-locator-from`, and the control bar for the ambiguity case. Both now read the description off the element the band lookup found.

```
before   19 of 21 checks, 2 captures that never ran
after    21 of 21, captures 2, unsettled 0, straddling 0, visual_failed 0, exit 0
```

`live_293` (10/10) and `live_290` (13/13) re-run clean at this head too — the three harnesses this touches all produce evidence.

## Also here: the #290 roadmap row

Six of seven criteria are measured met and the diff is wired (#715). What is left is one `en` control-bar fixture and a criterion that still asks for **Creator** baselines after Creator left product scope on 2026-07-17 — a wording fix, not work.

---

This is the fourth locale binding found this week and the deepest: not a harness, but the **library default** every harness inherits. It is also the second time I named a cause from memory instead of measuring it — the first was calling the collector's header binding English-only when `readHeaders` compares through LabelSets.

### DIFF
--- a/Scripts/livekit/evidence.py
+++ b/Scripts/livekit/evidence.py
@@ -156,12 +156,32 @@ def _running_harness_name():
     return _safe_name(f"unidentified-harness-pid{os.getpid()}")
 
 
-def logic_window(title_contains="Tracks"):
+# The arrange window's title, in every language this repository has measured it in. Logic names it
+# after the current view, so an English Logic says `… - Tracks` and a Korean one `… - 트랙`.
+#
+# This sits beside `logic_window` because its DEFAULT was the English word, and eighteen harnesses
+# take that default. Measured 2026-08-29 on a Korean Logic: `logic_window("Tracks")` returned None,
+# so `shot` took its no-window branch and recorded `settled: false`, `region: null` and
+# `wholly_within: false` — and I read those three as a settling problem and blamed blinking level
+# meters. They were one missing translation, and the capture never happened at all.
+ARRANGE_WINDOW_TITLES = ["Tracks", "트랙", "トラック"]
+
+
+def logic_window(title_contains=None):
     """The on-screen bounds of a Logic window, via CoreGraphics rather than AX.
 
     Deliberately not an AX read: the harness must be able to see the window even when the AX tree is
     exactly what is under test. Returns None when no matching window is on screen.
+
+    `title_contains` defaults to every known spelling of the arrange window rather than to one of
+    them. A caller naming a specific window still gets exactly that.
     """
+    if title_contains is None:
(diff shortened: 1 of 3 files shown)


MongLong0214:
### 고친 기본값이 다른 하네스에서도 통하는지 확인했습니다

이 PR 이 손댄 셋 말고, 그 기본값을 쓰는 다른 하네스 둘을 같은 head 에서 돌렸습니다.

```
live_543_lookup_names_the_step   exit 0   checks 6/6   captures 2   unsettled 0
live_549_cell_does_not_veto      exit 1   checks 0/1   captures 0
```

**543 이 이 변경의 값을 보여줍니다** — 캡처 둘이 실제로 일어났고 정착했습니다. 이 머신에서 전에는 창을 못 찾아 아무것도 못 찍었을 하네스입니다.

**549 는 로케일 때문이 아닙니다.** 자기 전제에서 멈춥니다:

```
549/precondition-the-trigger-is-present  ->  failing_node_present=None
```

특정 노드가 있는 프로젝트를 요구하는데 지금 열린 건 `lpm-606-warm` 입니다. 프로젝트 상태 문제이지 이 PR 이 고치는 부류가 아니라 손대지 않았습니다 — **여기서 고쳐서 green 을 만들면 이 변경이 무엇을 고쳤는지가 흐려집니다.**

즉 18개 하네스 전부가 이 커밋으로 돌게 되는 건 아니고, **창을 못 찾아 막혀 있던 것들이 풀립니다.** 나머지는 각자의 전제를 갖고 있고 그건 별개입니다.

PR #721, opened 2026-08-30
fix: recover a modifier left held when a flagged chord is interrupted (discussion #458)

MongLong0214:
Discussion #458 reported Logic becoming hard to use after running the integration: track volume
moving 0.1 dB at a time, tracks and locators not draggable. That is the signature of a held
Control during mouse interaction. The discussion has been open 25 days with an acknowledgement and
no change.

I could not reproduce it in normal use and still cannot, so this does not claim to be that user's
cause. What it closes is a window I found while looking for one.

## The window

A flagged chord is three posts: key-down with flags, key-up with flags, and a key-up with **no**
flags whose only job is to release the modifier. `AXMouseHelper.postFlaggedKeyEvent` sends all
three back to back, so the sequence is correct — but they are three separate calls to
`CGEvent.post`, and a process that dies between the second and the third leaves macOS believing
Control or Shift is held while no physical key is down.

The only way I produced the reported state was by deliberately omitting that third event. It is
reachable, and nothing recovered from it.

## Why a marker and not a blind clear

Posting a modifier-clear at startup unconditionally would also fire while a user is genuinely
holding Shift. Nothing available here can tell a synthetic held flag from a physical one —
`CGEventSource.flagsState` reports the combined state and cannot say who put a flag there.

A marker written before the chord and removed after it answers the question that actually matters:
*did our process leave a chord unfinished?* Present means yes, and the recovery posts the same
key-up the interrupted sequence owed. Absent means nothing is owed and nothing is posted. It
cannot fire because of anything a user is doing.

The marker is removed **before** the post, so a crash inside the post cannot make this repeat on
every subsequent launch — a recovery that replays a keystroke at every start is worse than the
state it recovers.

## What it does not cover, stated rather than implied

- `SIGKILL` and hard power loss remove the process without removing the marker. That is the case
  this recovers.
- A marker that cannot be written (read-only home, sandbox) leaves the window exactly as wide as
  it is today, and the chord still runs. Refusing to send a keystroke because a breadcrumb failed
  would trade a rare stuck modifier for a broken feature.
- It does not prove the reporter's cause. If they can still reproduce it, the marker's presence at
  the next start is now a fact we can ask them for.

## Evidence

Six cases in `StuckModifierRecoveryTests`, one per branch: no marker posts nothing; an armed chord
that never disarmed is cleared; a completed chord owes nothing; the marker is gone while the post
runs; an undecodable marker posts nothing and does not survive; and the recovery replays the
clear, never the chord's flags.

Full suite green, public-surface preflight clean, live evidence regenerated (13/13).

### DIFF
--- a/.commitlore-policy.json
+++ b/.commitlore-policy.json
@@ -1,0 +1,6 @@
+{
+  "mode": "auto",
+  "unattended": true,
+  "max_records_per_commit": 1,
+  "require_verified_evidence": true
+}
--- a/Scripts/check-falsifiable-adoption.py
+++ b/Scripts/check-falsifiable-adoption.py
@@ -37,7 +37,7 @@ import sys
 REPO = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
 
 # The measured floor. Raise it when a harness converts; never lower it to make a branch pass.
-FLOOR = 4
+FLOOR = 5
 _FALSIFIABLE_PARAMETERS = (
     "tag", "predicate", "observation", "counterexample", "expected", "mutation", "modal_snapshot",
 )
--- a/Scripts/livekit/ax_control_bar_control_values.swift
+++ b/Scripts/livekit/ax_control_bar_control_values.swift
@@ -1,0 +1,166 @@
+// Read the AXValue of every button and checkbox below ONE named control-bar AXGroup.
+//
+//   ./ax_control_bar_control_values '["control bar", "localized control-bar name"]'
+//
+// The caller supplies the label family from AXLocalePolicy through `E.label_set`; this probe does
+// not carry a second, guessed localization table.  It refuses a missing or ambiguous group, a
+// failed AXValue read, an unnamed/duplicate control, or zero controls.  Any of those states would
+// make a `description -> AXValue` map incomplete, and an incomplete map is not evidence that no
+// transport control changed.
+import AppKit
+import ApplicationServices
+import Foundation
+
+enum ProbeError: Error, CustomStringConvertible {
+    case message(String)
+    case ax(attribute: String, code: AXError)
+
+    var description: String {
+        switch self {
+        case let .message(message): return message
+        case let .ax(attribute, code): return "AX read \(attribute) failed: \(code.rawValue)"
+        }
+    }
+}
+
+func emit(_ body: [String: Any]) -> Never {
+    let data = try! JSONSerialization.data(withJSONObject: body, options: [.sortedKeys])
+    print(String(data: data, encoding: .utf8)!)
+    exit(0)
+}
+
+func attribute(_ element: AXUIElement, _ name: String) throws -> CFTypeRef {
+    var value: CFTypeRef?
+    let result = AXUIElementCopyAttributeValue(element, name as CFString, &value)
+    guard result == .success, let value else { throw ProbeError.ax(attribute: name, code: result) }
+    return value
+}
+
+func string(_ element: AXUIElement, _ name: String) throws -> String {
+    guard let value = try attribute(element, name) as? String else {
+        throw ProbeError.message("\(name) is not a string")
+    }
+    return value
+}
+
+func children(_ element: AXUIElement) throws -> [AXUIElement] {
+    var value: CFTypeRef?
+    let result = AXUIElementCopyAttributeValue(element, kAXChildrenAttribute as CFString, &value)
+    // AX reports a leaf as `noValue`, which means zero descendants, not a failed tree read.
+    if result == .noValue { return [] }
+    guard result == .success, let value, let children = value as? [AXUIElement] else {
+        throw ProbeError.ax(attribute: kAXChildrenAttribute as String, code: result)
+    }
+    return children
+}
+
+func optionalString(_ element: AXUIElement, _ name: String) -> String? {
+    var value: CFTypeRef?
+    guard AXUIElementCopyAttributeValue(element, name as CFString, &value) == .success else {
+        return nil
+    }
+    return value as? String
+}
+
+func jsonValue(_ raw: CFTypeRef) -> Any? {
+    if let number = raw as? NSNumber { return number }
+    if let text = raw as? String { return text }
+    return nil
+}
+
+let labels: [String]
+do {
+    let raw = CommandLine.arguments.dropFirst().first ?? ""
+    guard let data = raw.data(using: .utf8),
+          let decoded = try JSONSerialization.jsonObject(with: data) as? [String] else {
+        throw ProbeError.message("expected one JSON array of control-bar labels")
+    }
+    labels = decoded.filter { !$0.trimmingCharacters(in: .whitespacesAndNewlines).isEmpty }
+    guard !labels.isEmpty else { throw ProbeError.message("control-bar label family is empty") }
+} catch {
+    emit(["ok": false, "error": String(describing: error)])
+}
+
+guard let app = NSWorkspace.shared.runningApplications.first(where: {
+    ($0.bundleIdentifier ?? "").contains("logic")
+}) else {
+    emit(["ok": false, "error": "Logic is not running"])
+}
+
+do {
+    let application = AXUIElementCreateApplication(app.processIdentifier)
+    let mainWindow = try attribute(application, kAXMainWindowAttribute as String) as! AXUIElement
+    var groups: [AXUIElement] = []
+
+    func findControlBars(_ element: AXUIElement, depth: Int = 0) throws {
+        guard depth <= 24 else { throw ProbeError.message("control-bar walk exceeded depth limit") }
+        for child in try children(element) {
+            if try string(child, kAXRoleAttribute as String) == (kAXGroupRole as String),
+               let description = optionalString(child, kAXDescriptionAttribute as String),
+               labels.contains(where: { $0.caseInsensitiveCompare(description) == .orderedSame }) {
+                groups.append(child)
+            }
+            try findControlBars(child, depth: depth + 1)
+        }
+    }
+
+    try findControlBars(mainWindow)
+
+    // Measured 2026-08-30: Logic nests a control-bar AXGroup inside another AXGroup carrying the
+    // SAME description, so a description match alone returns two. They are not two control bars —
+    // one contains the other. Keep the OUTERMOST, defined structurally as the candidate that has no
+    // other candidate above it. That is a containment test, not a choice by tree order, and if two
+    // genuinely disjoint control bars ever appear this still refuses rather than picking one.
+    var outermost: [AXUIElement] = []
+    for candidate in groups {
+        var enclosed = false
+        for other in groups where !CFEqual(candidate, other) {
+            var walker = candidate
+            var hops = 0
+            while hops < 24, let raw = try? attribute(walker, kAXParentAttribute as String),
+                  CFGetTypeID(raw) == AXUIElementGetTypeID() {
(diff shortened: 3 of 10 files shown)


MongLong0214:
## Adversarial review refuted all three safety claims. Not merging this as it stands.

I put this head through a blind adversarial review asked to *refute* the claims, not confirm them.
It refuted all three, each with a line to point at. Every one is real.

### C1 — "it cannot fire because of anything a user is doing" is false

`defaultURL` is **one fixed path shared by every process**. So:

1. Process A arms the marker and is mid-chord, healthy, uninterrupted.
2. The user launches process B — two MCP clients at once is ordinary, not exotic.
3. B's `start()` calls `recoverIfNeeded()`, sees A's **live** marker, deletes it, and posts a
   key-up.

A keystroke is posted for a chord that was never interrupted. That is precisely the property the
type comment claims is impossible. And it is worse than a spurious keystroke: B deleted A's
marker, so if A dies later, nothing recovers it.

### C2 — arm/disarm can leak, and can disarm someone else

No single-stack path escapes the `defer`. But `disarm`'s `try? removeItem` swallows failure, so a
marker can survive a *completed* chord and fire an unowed key-up at the next start. And under
overlap the fixed path breaks ownership outright: A arms, B overwrites, then A's `defer` deletes
**B's** marker.

### C3 — the wiring is untested

This is the one that matters most.

> Removing the production arm/defer at `AXMouseHelper.swift:116-117`, removing startup recovery at
> `LogicProServer.swift:1184`, or changing the production clear's flags at
> `StuckModifierRecovery.swift:103` leaves **all six tests passing**.

Five of the six tests are individually effective against mutations to the pure core — the review
named a killing mutation for each. But nothing tests that the core is *connected*. I could delete
the entire integration and the suite stays green.

The sixth test, "the recovery replays the CLEAR, never the chord", is decoration for its headline
claim: it never invokes the default production closure, so setting `clear.flags` to a nonzero
value fails nothing. The test asserts the marker *records* flags; the safety property is that the
recovery *does not replay* them, and that is untested.

No dead assertions and no Sendable problems were found.

### What I am changing

- The marker gets an **owner**: per-process filename plus a recorded pid, and recovery fires only
  for a marker whose pid is not alive (`kill(pid, 0)`, with any ambiguous errno treated as alive —
  skipping is the safe direction). Residual, stated rather than hidden: **pid reuse** can make an
  orphaned marker look alive, which skips a recovery and leaves the modifier stuck. That is the
  state before this feature; it never posts a keystroke the user did not cause.
- Recovery posts **only if the marker was actually removed**, so a removal failure cannot make the
  keystroke repeat at every launch. A failed `disarm` logs instead of vanishing.
- The clear-event construction becomes a pure function with a test asserting zero flags.
- The chord body and the startup maintenance block become testable seams, so deleting either call
  fails a test. The one residual there — deleting the call to the startup helper from `start()`
  itself — will be written down rather than implied away.

MongLong0214:
## Second review: the ownership fix holds, the prose does not, and two real defects remain

Same discipline as the first round — a blind adversarial pass asked to refute, not confirm.

### What now holds

Every wiring mutation is caught except one, and that one is the residual I wrote into the commit message rather than a surprise:

| mutation | caught by |
|---|---|
| delete the arm call | `axMouseHelperPostChordKeepsMarkerArmedAcrossEveryPost` |
| delete the disarm/defer | same |
| move arm after the first post | same |
| delete the recovery inside the startup helper | `testLogicProServerStartupMaintenanceRunsCleanupAndModifierRecovery` |
| **delete the default-branch call to the helper** | **nothing** |
| change the clear's flags to `.maskControl` | `clearEventNeverReassertsTheChordFlags` |
| make `shouldRecover` always true | `recoveryOwnershipDecisionIsConservative` |

Fourteen tests, and the review named a killing mutation for **every one** — none is decoration. No dead assertions. No new Sendable violation.

### Two real defects

**Two chords in one process share one marker path.** `postFlaggedKeyEvent` is `@Sendable` and `Runtime` is `@unchecked Sendable`, so concurrent calls are allowed, and the filename is per-*process*. A arms, B arms over it, A's `defer` deletes **B's** marker, and B is mid-chord and now unrecoverable. Fixing with a lock and a nesting counter: arm on 0→1, disarm on 1→0.

**"cleared a modifier" is logged when nothing may have been posted.** The production poster returns `Void`, so if `clearEvent` returns nil the code still counts a recovery and logs one. The poster will return `Bool`, and only a constructed-and-posted event counts. A `post` that returns normally is still not proof of delivery, and the comment will say so rather than implying it.

### Eight places where the prose claims more than the code does

This is the largest item, and in this repository it is a defect and not a wording nit — the whole point of the change is that the comment said something the code did not do.

- A PID does not identify a process across reuse **or across a reboot**.
- `ESRCH` proves only that no process currently holds that PID — not that the owner is "known-dead" or that a clear is "owed".
- Hard power loss leaves a marker file but does **not** leave a held event state, so listing it as a case this recovers is wrong.
- The marker records no UID, no boot identity, no process start time, and no proof the clear was not already posted.
- "Every surviving completed marker is visible in the logs" is false: a kill after the clear and before `disarm` leaves one silently.
- "Safe on every start" and "concurrent starts race safely" are both refuted — by reboot-stale markers, and by the window between removal and post where another writer can recreate the file. That second window is the *deliberate* trade (remove first so a crashing post cannot loop on every launch) and will be labelled as deliberate rather than defended as absent.
- The startup comment says both actions repair artifacts a prior process left behind; the recovery has established neither provenance nor that a modifier is still held.

The residual list goes in once, plainly: PID reuse, reboot-stale markers, a marker under a different UID or home, no proof a modifier is still held at recovery time, and no test covering the production `kill`/errno path or actual CGEvent delivery.

The review was static — the managed worktree is read-only, so it did not build or run anything. The mutation verdicts above are its reading of the tests, and the suite result quoted in the last comment is mine.

MongLong0214:
## Third review: the logic holds, two code blockers remain, and the prose is still ahead of the code

### What now holds

Every mutation in the list I gave it is caught, including the ones this round added:

| mutation | caught |
|---|---|
| remove the nesting counter | yes |
| arm on every `begin` instead of at depth 0 | yes |
| disarm at every `end` instead of at depth 0 | yes |
| let a nested chord overwrite the outer key code | yes |
| count a recovery when the poster returns false | yes |

Twenty-four focused tests pass, no dead assertions, and no strict-concurrency diagnostic.

### Two code blockers

**An unmatched `end` disarms someone else's live chord.** I replaced the original `precondition` with a clamp that calls `disarm()` anyway. That is worse than the trap: at depth 0 nothing is armed, so it removes a marker this counter never wrote. It will log and return without disarming.

**Callbacks run while the lock is held.** `NSLock` is not recursive, so a callback that re-enters the chord path deadlocks. In production the callbacks are file writes and cannot re-enter, so this is a latent hazard rather than a live bug — and I am keeping them under the lock deliberately, because holding it across `arm` is what orders the marker before the first post. Moving them out would trade a hazard nothing reaches for a real ordering hole. The requirement goes on the type in one sentence instead.

### One thing I am deciding rather than fixing

Two threads can overlap non-LIFO — A arms, B begins, A ends, B ends and disarms — leaving the marker holding A's key code while B was the chord still running.

The consequence is smaller than it looks, and the reason has to be written down or this reads as a silent bug: **the recovery posts with zero flags, and it is the zero-flags event that releases the modifier, not the key code.** A key-up naming the wrong virtual key still clears the modifier state, and a key-up for a key nobody is holding releases nothing. A stale key code costs accuracy in a log line, not correctness of the clear. That goes in the type comment and in the residual list, stated rather than hidden.

### Fifteen more places where the prose is ahead of the code

The review quoted each one. A sample, to show the kind:

- *"A nested chord deliberately keeps the outer chord's key code"* assumes successful arming **and** LIFO nesting; neither is guaranteed.
- *"Recording both would require a marker per chord"* is simply false — one marker could hold several records. It was not built, which is a different sentence.
- *"Clears a synthetic modifier that an interrupted process left held"* — recovery establishes neither that a modifier is still held nor that the event was delivered.
- *"The sequence... leaves macOS believing Control or Shift is held"* — the flags may be Option, Command, or empty, and physical key state is never observed.
- *"[the zero-flags contract] cannot drift behind the recovery tests"* — replacing the production default poster leaves every injected test and the direct `clearEvent` test green.
- The suite docstring says the cases *"pin"* a process-death window and a held modifier. No test kills a process, observes system modifier state, or verifies delivery.

This is the third round in a row where the code survived and the comments did not. That is worth naming: the comments are where I keep overstating, and they are read as the contract.

### The residual list grows

It was accurate and incomplete. Adding: an unmatched `end`; non-LIFO two-thread ownership; the re-entrant-callback hazard; markers are unauthenticated and the filename's PID is never checked against the payload's; a failed `disarm` leaves a completed marker; the remove-before-post loss window; and the untested production `postChord` and `start()` wiring.

MongLong0214:
## Live-proven in the release artifact, and the control that proves it had to be rebuilt

Ship gate **PASS** with live evidence bound to this head.

### The gap that made a live harness necessary

`harness_evidence_coverage` reported all three changed `Sources/` paths as **unproven** — no harness claimed them — so the gate would have been satisfied by an unrelated clean document. That is exactly the substitution that tool exists to stop.

`live_458_a_stale_chord_marker_is_cleared_on_start.py` claims those paths and proves the ownership logic against the **built server**, not in a unit test:

- a marker owned by a **genuinely dead pid** — taken from a real child process that has exited, not an invented number — is cleared at startup
- a marker owned by a **live pid**, the harness's own, **survives**. This is the case that separates ownership from "delete every marker at startup", and the first version of this feature would have failed it
- an undecodable marker is discarded without posting

**11 of 11, `is_clean` true.**

### The pixel control could not answer its own question

The first version asserted that a band of the control bar does not change across the recovery. It failed, and I measured rather than assumed: three captures of that exact band with nothing happening are **byte-identical**, so the region is not self-changing and the difference was real — but it was **twelve retina columns, six points wide**, at the right edge of the LCD where the time and key signature popups sit. A chevron or a focus ring, most plausibly the artifact starting.

Six points of redraw and "the recovery pressed a key" are indistinguishable in pixels. Widening the band would have made the assertion unable to fail; deleting it would have left the recovery with no control at all.

**The control is now AX**: every control-bar checkbox and button, read as description → value, before the artifact starts and after the startup recovery, asserted identical. A zero-flags key-up that pressed anything in the transport moves one of them. The visual assertion stays, demoted to what it can establish — that the recording context is idle — and still turns red under playback.

Resolving the control bar needed the same care as the mixer did earlier today: a description match returns **two** `AXGroup`s, because Logic nests one inside another carrying the same description. They are not two control bars; one contains the other. The outermost is kept, defined as the candidate with no other candidate above it — containment, not tree order — and two genuinely disjoint control bars would still refuse rather than be picked between.

### Four review rounds, and the code outlasted the prose every time

Rounds one through three are in the comments above. The pattern is worth naming: the logic survived each pass and the **comments did not**. Fifteen sentences claimed more than the code did — a PID is not identity across reuse or a reboot, `ESRCH` proves only that nothing currently holds that pid, hard power loss leaves a file but no held event state, arming is best effort, removal is attempted rather than assured.

### Residual limitations, written down rather than implied

PID reuse; reboot-stale markers; a marker written under a different UID or home; the filename's pid is never checked against the payload's; a failed `disarm` leaves a completed marker; the remove-before-post window is deliberate and loses a recovery if the process dies inside it; and the chord half is covered by unit tests alone — no test kills a process, observes system modifier state, or verifies event delivery.

Issue #306, opened 2026-07-10
ADR-018: Third-party host parameters — backlog; prove one exact plug-in before generic expansion


MongLong0214:
# ADR-018 — Verified host-exposed third-party parameters

> **SSOT revision: 2026-09-21. OPEN / BACKLOG.** Some shared code and individual read experiments exist; a generic verified third-party write platform does not. #308 owns activation. This issue does not block #944's small read-only library reuse or existing stock AX correctness work.
>
> Audit baseline: `d067e91958ad3ff1a1bdfb1e6e39a3e23e786517`. Historical experiments are not a new all-vendor qualification at this commit.

## Decision and bounded objective

Control the exact plug-in instance hosted by Logic through a demonstrably target-faithful host parameter surface. A separate Audio Unit instantiated by the MCP process is not the Logic insert. Coordinates, OCR, parameter-name coincidence or an uncorroborated channel bank are not sufficient verified target identity.

When activated, start with **one explicitly named plug-in build, one exact insert and one required parameter**. Reuse existing catalog/identity/readback helpers. A vendor SDK, all-vendor manifest approval platform and new TestAU ecosystem are not prerequisites for this first proof.

## Implementation history and corrections

- [The initial insertion/read correction](https://github.com/MongLong0214/logic-pro-mcp/issues/306#issuecomment-5247240090) found that picking a plug-in name was not the same as selecting its Stereo/Dual Mono format leaf. Driving the proper menu path inserted Raum; its native editor exposed little AX, while Logic's Controls view exposed named values. This is a real third-party **read observation**, so 'no third-party surface has ever been measured' is incorrect.
- Readability and advertised settability did not establish a verified write. Later stock Controls-view slider/popup failures are scoped observations, not proof about every vendor or view.
- The old 'four modules, zero callers' statement counted filenames instead of declared symbols and omitted a fifth live helper. `HostParameterGate.swift` declares `HostWriteRejection`, and `HostParameterPlane.swift` declares `PluginParameterProvider`. `ControlsViewBooleanParameterWriter` is called from the existing verified AX path. Preserve it; do not delete this whole directory as unused.
- #846/#848 repaired probe/product label matching. #852 corrected a probe that toggled an already-open editor closed and called it not found.
[...]

Issue #305, opened 2026-07-10
ADR-017: Flex Pitch — backlog; validate a real analysed-region surface before expansion


MongLong0214:
# ADR-017 — Flex Pitch inspection and verified editing

> **SSOT revision: 2026-09-21. OPEN / BACKLOG. No public verified Flex Pitch editing is claimed.** Activate through #308 only for a selected user workflow. Existing models are not a delivered UI/data path; unmeasured access is not a universal platform prohibition.
>
> Audit anchor: `d067e91958ad3ff1a1bdfb1e6e39a3e23e786517`. The historical readings below were not rerun for this issue edit.

## Decision

Start with a monophonic, already Flex-analysed audio-region fixture and establish exact note identity and actual readable state in the Audio Track Editor. Only then consider one scalar edit and independent result verification. Do not first build a large F0 engine, a generic pitch-edit DSL or checkpoint orchestration around an unproven live input.

The long-term scope remains monophonic inspection, qualified scalar edits, and separately qualified timing/topology edits. Polyphonic support and coordinate-only automation are not initial public capabilities.

## History and corrected conclusions

- Early comments described the Arrange region subtree or partially prepared editor and generalized their empty results. Those snapshots must retain their exact surface, fixture, Flex mode and analysis-completion conditions.
- A later probe claimed Show/Hide Flex was inert. [The September 11 correction](https://github.com/MongLong0214/logic-pro-mcp/issues/305#issuecomment-5629188936) demonstrated a real 0 → 1 → 0 transition once the actual frontmost target and positive control were established. Showing Flex is not itself analysing audio.
- [The September 13 investigation](https://github.com/MongLong0214/logic-pro-mcp/issues/305#issuecomment-5653969409), associated with #881, could not obtain the required campaign fixture: no audio region was present and attempted audio import did not produce the expected panel within the observed window. That is a setup/instrument limitation, not proof that all analysed-editor note data is unavailable.
- Other comments record a small editor census. Those are not erased, but neither an element count nor a substring search proves that the correct analysed note surface was exhaustively inspected. Establish the target/analysis preconditions before relying on their negative result.
- `FlexPitchModel`, `FlexPitchEditGate` and `FlexPitchRefValidity` groundwork does not show an actual live snapshot producer. Check callers by symbol before retaining/removing scaffolding.

[...]

Issue #301, opened 2026-07-10
ADR-013: Channel EQ — finish truthful state readback and bounded verified-write scope


MongLong0214:
# ADR-013 — Channel EQ state readback and bounded verified writes

> **Execution SSOT r2 — 2026-09-21. OPEN / selected remainder; preserve the shipped writer.** #308 schedules work; #292 owns shared stock capability evidence. **#955 owns full-plan/reset/diff/batch/compensation/Automatic EQ.** Source checkpoint: `1b36b90bb8f339c82a3d606c2f44cab08ef18da6`. Historical observations below are not a fresh qualification of this source.

## 1. Decision and immediate deliverable

Use the existing native-editor AX plane, exact-instance acquisition and `SliderIncrementWalk`. Preserve `logic_plugins.set_eq_band_verified`; do not rename it to a prototype alias or implement C4 endpoints, `ChannelEQC4Adapter`, a virtual control surface or a new EQ doctor subsystem.

The first deliverable is **one real same-instance state reader**, followed by completion of the existing supported scalar-write contract. A whole-state reader is not all-band write support; one successful gain write is not full-plan application.

#292 and this issue share the existing catalog/AX contract, not a circular whole-epic dependency. One implementation worker edits shared catalog/verified-AX files at a time; parallel work may own isolated tests/review.

## 2. History and closure correction

1. [August 30 correction](https://github.com/MongLong0214/logic-pro-mcp/issues/301#issuecomment-5466538022): the older AU census instantiated a different object outside Logic and did not prove that the actual hosted EQ editor was opaque.
2. [Corrected census](https://github.com/MongLong0214/logic-pro-mcp/issues/301#issuecomment-5467110685): distinguish the 24 named band controls from total sliders/checkboxes. Multiplying bands by fields is not a measured inventory.
3. [Write measurement](https://github.com/MongLong0214/logic-pro-mcp/issues/301#issuecomment-5468417002): a successful AX set could move one step rather than reach the target; bounded observed convergence was required.
4. [PR #725](https://github.com/MongLong0214/logic-pro-mcp/pull/725), merge `434267f2c1a9aae46ea96d2d09260d7dc3bb80db`, added the named writer and closed #301 on 2026-08-31. It also addressed #724 acquisition/menu issues and #726 instance confusion. It did not deliver every original state/batch/rollback API.
[...]

Issue #299, opened 2026-07-10
ADR-011: Full Compressor and auto-setting — backlog; next step is one additional verified parameter


MongLong0214:
# ADR-011 — Compressor parameter expansion and later auto-setting

> **SSOT revision: 2026-09-21. OPEN / BACKLOG. Full Compressor and automatic setting are not implemented.** Existing individually supported parameter operations remain maintained under #292. Activate this expansion only through #308; do not treat the old release-train dates as a current instruction.
>
> Audit anchor: `d067e91958ad3ff1a1bdfb1e6e39a3e23e786517`. Historical observations below were not rerun for this body revision.

## Decision and scope

Reuse the current stock-plug-in capability catalog, exact-instance acquisition and verified parameter writer. The first expansion, when selected, is **one additional named parameter with correct physical-unit read/write/readback**. Do not begin by building the original whole auto-setting engine, every circuit model or a second capability registry.

This issue retains the original long-term Full Compressor scope: core parameters, model-dependent availability, full supported-state snapshots, batch compensation, recommendation and explicitly confirmed application. Those stages remain unimplemented/backlogged until individually selected; no narrow success silently closes the whole issue.

## History that an implementer must not repeat

- Existing verified insertion and limited `set_param_verified` work predate this expansion. Requests missing target corroboration, supported mode or project identity can correctly refuse before any parameter experiment begins.
- August comments recorded out-of-release decisions and later reopening. The current disposition is open backlog, not a completed feature and not abandonment.
- Native and Controls views are different providers. Some observations found parameter names and values on separate elements without a demonstrated non-positional association; readable numbers do not prove which parameter they represent.
- Controls-view sliders/popups and checkbox controls have different actuation behavior. The working `ControlsViewBooleanParameterWriter` is shared production code; an observation about that writer does not establish full native-editor support.
- [September 11 correction](https://github.com/MongLong0214/logic-pro-mcp/issues/299#issuecomment-5628734137): the valid association name is `AXLinkedUIElements` (plural); the singular probe could never match it. The `AXDescribedBy` row was withdrawn as the wrong platform attribute. Positive controls were added for other association reads. Earlier zero-count conclusions without those controls must not be recycled as evidence.
- Those corrected measurements still cover specific trees/builds/locales, not every possible Compressor route. Later Channel EQ association scans are not automatically measurements of Compressor.
[...]

PR #722, opened 2026-08-30
docs(roadmap): replace three inherited plug-in AX walls with measurements

MongLong0214:
Three roadmap rows (#301, #305, #306) each carried a blocker saying the AU parameter view is AX-opaque. Measured on live Logic 12.x today, all three were wrong, in different ways.

- **#301** — Channel EQ's native view exposes all 8 bands as **24 named, settable sliders** with `AXValueDescription` in Hz and dB. The opacity claim traces to `spike-channel-eq-au-census.swift`, which instantiates the Audio Unit *outside* Logic; ADR-018's own Important Boundary says that cannot speak for Logic's live instance.
- **#305** — Flex Pitch is in the Audio Track Editor, not a plug-in window. The blocker was inherited, never observed against this surface.
- **#306** — "same wall as #305" was wrong twice: different surface, and no wall. Tier B (Logic's Controls view) renders a stock plug-in as an `AXTable` with one named `AXRow` per parameter, addressed by row label rather than index or coordinate.

The wall that did hold is narrower and different in kind: the Compressor's *native* editor names 1 of its 22 sliders, which is why exactly one Compressor parameter is supported. That is an **identity** wall, not a reachability one.

Measurements: #301 https://github.com/MongLong0214/logic-pro-mcp/issues/301#issuecomment-5466538022 · #299 https://github.com/MongLong0214/logic-pro-mcp/issues/299#issuecomment-5466600291

No write was attempted anywhere, and third-party plug-ins stay unmeasured. Settability is a claim by the element, and AX settability lies.

### DIFF
--- a/Scripts/check-roadmap-blockers-cite-measurements.py
+++ b/Scripts/check-roadmap-blockers-cite-measurements.py
@@ -32,13 +32,28 @@ So a row that mentions a wall must carry, in the same cell, either an ISO date o
 comment that recorded the measurement. That is a low bar deliberately — it is the bar that would
 have caught all three, and a higher one would be a bar on prose that nothing can enforce.
 
-It applies in BOTH directions, and that is not an accident of substring matching. "No measured AX
-wall in front of it" is as much a claim about the world as "AX-opaque", and it is the claim nobody
-asked #305 to date. When did you check there was no wall? is exactly the question whose absence
-cost three rows. A row denying a wall answers it the same way a row asserting one does.
+It applies in BOTH directions, and that is not an accident of substring matching. The row for #304
+said "no measured AX wall in front of it", and *when did you check there was no wall?* is exactly
+the question whose absence cost the three rows above. A row denying a wall answers it the same way
+a row asserting one does.
 
 It says nothing about rows that describe work as unstarted, behind a dependency, or waiting on a
 decision without mentioning a wall at all. Those are not claims about a surface.
+
+WHAT IT MISSES, so nobody reads a pass as coverage
+--------------------------------------------------
+It matches literal phrases in the LAST cell of a row whose cells include exactly `OPEN`. So it does
+not see:
+
+  * a wall described in words this list does not carry — "unreachable", "cannot be read",
+    "the tree returns nothing"
+  * a wall claim in a cell other than the last, or a table whose columns are ordered differently
+  * a state cell spelled anything but `OPEN`, or a row split across lines
+  * a pipe inside backticks, which splits one cell into two
+
+And a date is a pointer, not proof: `2026-08-30` typed into a cell satisfies this guard and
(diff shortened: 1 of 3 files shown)


MongLong0214:
## Review found the guard passing things it should not, and prose ahead of the evidence. Both fixed.

A blind adversarial pass on this branch, asked to refute rather than confirm.

### One technical error, and it is the one worth fixing before anyone builds on it

I wrote that the Controls view is *"the shape `set_param_verified` steps 9-12 already drive"*. **It is not.** Step 9 matches an `AXSlider` by **that slider's own `AXDescription`**, and the Controls view's sliders have none — I measured `sliders=26 named=0` myself and then wrote the opposite. The name sits on an `AXStaticText` in the row:

```
AXRow > AXCell
          AXStaticText val=Threshold:      <- the name is HERE
          AXGroup val=-20.0 dB
            AXSlider val=60 vdesc=-20.0 dB <- and has none of its own
```

Reaching these needs a **new row-label locator**, which is a different and larger piece of work than "the path exists, add descriptors". Corrected in the roadmap, on #299, on #292, and in ADR-013's revised Decision.

### Three claims narrowed to the evidence

- **"Tier B measured viable"** → *addressable*. No write, no readback, so viability is not established.
- **"the wall is gone"** → *the recorded wall is not where it was said to be*. Elements reporting `settable=yes` is a claim by the element, and this project has AX settability lying on record.
- **"the reason only `threshold` is supported is now measured"** → the correlation is measured; the mechanism is **inferred from the code path**.

### The guard had two real defects, and the drive could not have caught either

- **`CITED_BY_DATE` was decoration.** Its fixture said `opacity`, which is in no vocabulary list, so the case passed no matter what date handling did. Deleting `DATE` entirely left it green. Fixed, and the fixture now asserts it contains a wall phrase so it cannot silently stop testing what it names.
- **`main()`'s zero-scan refusal was asserted by nobody** while the drive's docstring claimed both refusals were covered. Both refusal paths now run through `main()` — a missing file, an unparseable one, and a table that parses to zero OPEN rows.

Both fixes were checked by reintroducing the defect: deleting `DATE` acceptance fails three cases, and returning 0 instead of 2 on a zero scan fails two.

### And the guard now says what it cannot see

A wall in words its list does not carry (*"unreachable"*, *"cannot be read"*), a claim outside the last cell, a reordered table, a differently spelled state cell, a pipe inside backticks. Plus the honest one: **a typed date satisfies it and measures nothing.** It raises the floor from "a wall with nothing beside it" to "a wall with something a reader can go and check", and that is the whole of what it does.

One factual error in the guard's own docstring is fixed too: it attributed the undated denial to #305 when the row it quotes is #304.

### CI

`build`, `compile`, `roadmap`, `test` all pass.

PR #723, opened 2026-08-30
fix(#608): see the modal, and stop calling plug-in windows one

### LINKED ISSUE #608: Every mutating operation is refused as "blocking dialog present" if it is the first call after server start
Found while proving #606. Reproducible, and it refuses real work on a condition that is not true.

## What happens

In a **fresh** server process, the first tool call that consults the blocking-dialog preflight is
refused. Any later call — including an immediate retry — passes.

```
call 1   project.save_as   error=unsupported_state  blocking_dialog_present=true   ← refused
call 2   project.save_as   state=A  success=true                                   ← wrote the file
```

Same process, same screen, three seconds apart. Nothing was dismissed in between.

It is not specific to `save_as`. A fresh process doing `logic_tracks.list_library` first is refused
the same way, and doing anything at all first — `logic_system.health` will do — makes the next call
succeed:

```
fresh process → health → save_as     state=A  success=true   ← first save_as, no refusal
```

## It refuses on something it cannot name

`blockingLogicDialogResult` only reports `dialog_title`/`dialog_role` when `blockingDialogInfo()`
returns an identity. In every one of these refusals it returns **nil** — so the caller is told
`blocking_dialog_present: true` with nothing named.

I added a census to that path (in the #606 branch) so the refusal reports what the guard saw. On a
refusal:

```json
"dialog_presence_census": {
  "reason": "windows_read",
  "windows": [{ "title": "… - Tracks", "subrole": "AXStandardWindow",
                "is_blocking_dialog": false, "children": 10,
                "sheet_children": 0, "children_with_unreadable_role": 0 }],
  "dialog_present_recheck": false
}
```

**`dialog_present_recheck` re-asks `dialogPresent()` at census time and gets `false`** — the same
question, the same process, milliseconds after the refusal that answered `true`. One window, standard
subrole, no sheets, every attribute reading cleanly. There is no dialog and there was none.

## Where it comes from

`AXLogicProElements.dialogPresent` is fail-closed by design, and correctly so:

```swift
guard let app = appRoot(runtime: runtime) else { return true }
…
case .success(.absent), .success(.malformed), .failure:
    return true
```

`appRoot` resolves the PID, and one of the resolution paths (`logicProPIDViaSystemEvents`) runs an
AppleScript. A cold first call in a fresh process is the obvious candidate for the transient, which
matches the "any prior call fixes it" behaviour exactly. I have **not** proven which of the four
true-returning branches fires — the census is taken after the fact, by which time the condition is
gone — so treat the mechanism as the leading hypothesis and the behaviour as measured.

## Why it matters

Fail-closed is right. Fail-closed **on a transient** is a guard that refuses correct work at the worst
possible moment: the first thing a client does after connecting. And because the refusal names no
dialog, the operator is sent looking for a modal that is not on screen — which is exactly how this
cost me an afternoon on #606.

## What a fix has to do

1. Establish where the transient is, by recording the branch **at the moment the guard answers true**
   rather than reconstructing afterwards.
2. Not paper over it with a blind retry: a guard that retries until it likes the answer is not a
   guard. If the condition is "the app root was not resolvable yet", that is a distinct, nameable
   state — `logic_not_yet_addressable` — and it deserves its own refusal, separate from
   `blocking_dialog_present`, so a caller can tell "wait a moment" from "dismiss your dialog".
3. Keep the census. A fail-closed refusal that cannot name what it saw is the defect class this
   repository already refuses elsewhere.

MongLong0214:
Closes #608.

## What happened

A Logic alert about a missing audio device sat on screen for most of a measuring run and nothing said so. Under it:

```
File > 열기…      AXEnabled = false        <- Open is NEVER disabled in a healthy Logic
File > 별도 저장…  AXEnabled = false
pressing 별도 저장…  opened no panel        <- so the false was HONEST; the app was blocked
```

With the alert dismissed, every one of those reads `true`. I built a whole explanation about menu validation on readings taken through it and threw it away. The tell was in the data — `Open` reading false — and I walked past it.

## The two halves

**The harness could not see a modal.** `evidence.py` now records, at the moment *each check is written* rather than once per run, whether a Logic window sits at the CoreGraphics modal-panel level, and `is_clean` refuses a run that recorded anything while one did. Per-check because this alert appeared, was dismissed, and reappeared inside one sitting. The level is derived from `CGWindowLevelForKey(kCGModalPanelWindowLevelKey)` rather than written as `8`, and the read is CoreGraphics rather than AX for the same reason `logic_window` already is: the harness has to see the screen even when the AX tree is the thing under test.

**`isBlockingDialogWindow` called plug-in windows blocking.** That is #608. Measured in one instant:

| window | AX subrole | AXModal | CGWindowLayer |
|---|---|---|---|
| Logic alert | `AXDialog` | **true** | 8 |
| Plug-in window `Studio Grand` | `AXDialog` | **false** | 3 |
| Arrange window | `AXStandardWindow` | false | 0 |

`AXDialog` describes a window class, not whether it blocks. An unreadable `AXModal` is treated as **blocking**, because a caller told "nothing is blocking" on the strength of a read that did not happen is the shape this gate exists to refuse.

## One defect in my own guard, recorded rather than quietly fixed

It first read bounds behind `isinstance(b, dict)`. CoreGraphics returns an **NSDictionary proxy**, so that test is `False` on the real window list and the loop silently skipped a modal it had already identified. Every synthetic case passed while it did — the fixtures supplied real dicts, so **the fixture was more permissive than the thing it stood for and could not fail**. There is now a subscriptable non-dict fixture; reintroducing the `isinstance` test makes that case fail *and* makes the live read go blind again.

## Three measurements became harnesses

`live_301_channel_eq_bands_are_addressable`, `live_306_controls_view_labels_the_row_not_the_slider`, `live_369_export_menu_path_is_reachable`. Each makes the modal check its first precondition.

They refused four times before passing, and **every refusal was right**:

1. **five Compressor inserts** exist across the mixer — "one resolves" was false, and picking one by position is not allowed
2. two elements answer to a track name — the arrange **track header** and the **mixer strip**
3. two `AXLayoutArea`s answer to the mixer label — the Mixer pane and the **Inspector's channel strip for the selected track**, same role, same description, both real. Ancestry separates them
4. `Absolute Zero` carries **two** Compressors; `Studio Grand` carries one

And they corrected four numbers I had written into issue comments. The Channel EQ group exposes **26** sliders, not 24 — 24 band parameters plus two output `Gain` sliders — and **11** checkboxes, not 8. The Compressor census is **per-instance**: 22/11 on one strip, 20/10 on another, in the same project. Four assertions went red on readings that were **right**. They assert names and invariants now, because a total arrived at by arithmetic is not a measurement.

## Results

```
live_301  14 / 14      live_306  16 / 16      live_369  10 / 10
checks_recorded_under_blocking_modal: 0      restorations_failed: 0
```

Ship gate PASS at `66a093e1`, including live evidence for this head.

**Not covered, stated rather than implied:** `isBlockingDialogWindow` lives in a standalone script with no test harness, so the Swift half is verified by typecheck and by the measurement above, not by a case that has been watched fail.

### DIFF
--- a/Scripts/check-falsifiable-adoption.py
+++ b/Scripts/check-falsifiable-adoption.py
@@ -22,10 +22,11 @@ promise: the count may rise and may not fall. Raising FLOOR is a reviewed edit i
 is where a ratchet's teeth are — a number that can be lowered by whoever is inconvenienced is a
 suggestion.
 
-It does not check that a counterexample is a GOOD one. `falsifiable` establishes that a predicate
-can distinguish the observation from one stated alternative, and no more; whether that alternative
-is the one that matters is the author's claim and a reviewer's job. Said here because a guard named
-"adoption" invites being read as "quality".
+It does not check that a counterexample is a GOOD one. It does reject one mechanically detectable
+hollow form: a constant dictionary counterexample paired with a lambda that only subscripts a
+boolean wrapper. That form proves lookup distinguishes true from false, not that the predicate
+examines its stated observation. Beyond that narrow case, whether the alternative is the one that
+matters remains the author's claim and a reviewer's job.
 """
 import ast
 import glob
@@ -37,8 +38,42 @@ REPO = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
 # The measured floor. Raise it when a harness converts; never lower it to make a branch pass.
 FLOOR = 4
 
-def _calls_falsifiable(text):
-    """Whether the source contains an actual CALL to `falsifiable`, not a mention of the word.
+def _call_argument(call, position, keyword):
+    if len(call.args) > position:
+        return call.args[position]
+    return next((arg.value for arg in call.keywords if arg.arg == keyword), None)
+
+
+def _constant_literal(node):
+    if isinstance(node, ast.Constant):
+        return True
+    if isinstance(node, (ast.List, ast.Tuple, ast.Set)):
+        return all(_constant_literal(item) for item in node.elts)
+    if isinstance(node, ast.Dict):
+        return all((key is None or _constant_literal(key)) and _constant_literal(value)
+                   for key, value in zip(node.keys, node.values))
+    return False
+
+
+def _is_hollow_falsifiable(call):
+    """The known boolean-wrapper pattern, not a claim to judge arbitrary counterexamples."""
+    predicate = _call_argument(call, 1, "predicate")
+    counterexample = _call_argument(call, 3, "counterexample")
+    if not isinstance(predicate, ast.Lambda) or not isinstance(counterexample, ast.Dict):
+        return False
+    if not _constant_literal(counterexample):
+        return False
+    if len(predicate.args.args) != 1 or predicate.args.vararg or predicate.args.kwarg:
+        return False
+    parameter = predicate.args.args[0].arg
+    body = predicate.body
+    return (isinstance(body, ast.Subscript)
+            and isinstance(body.value, ast.Name)
+            and body.value.id == parameter)
+
+
+def _falsifiable_calls(text):
+    """`(has_real_call, hollow_lines)` for actual calls, never comments or strings.
 
     Parsed, because a regex counts `# falsifiable(` in a comment and a `"falsifiable("` in a
     string. Found by review, 2026-08-29, reproduced: a harness whose only occurrence was a comment
@@ -50,22 +85,33 @@ def _calls_falsifiable(text):
     try:
         tree = ast.parse(text)
     except SyntaxError:
-        return False
+        return False, []
+    has_real_call = False
+    hollow_lines = []
     for node in ast.walk(tree):
         if not isinstance(node, ast.Call):
             continue
         func = node.func
         name = func.attr if isinstance(func, ast.Attribute) else getattr(func, "id", None)
         if name == "falsifiable":
-            return True
-    return False
+            if _is_hollow_falsifiable(node):
+                hollow_lines.append(node.lineno)
+            else:
+                has_real_call = True
+    return has_real_call, hollow_lines
 
 
-def adoption(paths=None):
-    """`(adopters, total)` — harnesses calling `falsifiable(`, and how many there are."""
+def _calls_falsifiable(text):
+    """Whether this source has an actual, non-hollow call to `falsifiable`."""
+    return _falsifiable_calls(text)[0]
+
+
+def adoption(paths=None, include_hollow=False):
+    """Adopting harnesses and total; optionally include separately reported hollow call locations."""
     files = paths if paths is not None else sorted(
         glob.glob(os.path.join(REPO, "Scripts", "livekit", "live_*.py")))
     adopters = []
+    hollow = []
     for path in files:
         try:
             with open(path, encoding="utf-8") as fh:
@@ -74,13 +120,18 @@ def adoption(paths=None):
             # Unreadable is not adopted. Counting it either way would let a permissions accident
             # move the number, and this number is the whole point of the guard.
             continue
-        if _calls_falsifiable(text):
-            adopters.append(os.path.basename(path))
+        has_real_call, hollow_lines = _falsifiable_calls(text)
+        name = os.path.basename(path)
+        hollow.extend((name, line) for line in hollow_lines)
+        if has_real_call:
+            adopters.append(name)
+    if include_hollow:
+        return adopters, len(files), hollow
     return adopters, len(files)
 
 
 def main():
-    adopters, total = adoption()
+    adopters, total, hollow = adoption(include_hollow=True)
     if total == 0:
         print("-> FAIL: no live harnesses found — an empty set satisfies any floor")
         return 1
@@ -88,6 +139,8 @@ def main():
     print(f"   falsifiable() adoption: {len(adopters)} of {total} harnesses (floor {FLOOR})")
     for name in adopters:
         print(f"     uses it  {name}")
+    for name, line in hollow:
+        print(f"     hollow   {name}:{line} (constant counterexample + boolean-wrapper predicate)")
 
     if len(adopters) < FLOOR:
         print(f"-> FAIL: adoption fell to {len(adopters)}, below the recorded floor {FLOOR}")
--- a/Scripts/check-falsifiable-adoption.py
+++ b/Scripts/check-falsifiable-adoption.py
@@ -23,10 +23,11 @@ is where a ratchet's teeth are — a number that can be lowered by whoever is in
 suggestion.
 
 It does not check that a counterexample is a GOOD one. It does reject one mechanically detectable
-hollow form: a constant dictionary counterexample paired with a lambda that only subscripts a
-boolean wrapper. That form proves lookup distinguishes true from false, not that the predicate
-examines its stated observation. Beyond that narrow case, whether the alternative is the one that
-matters remains the author's claim and a reviewer's job.
+hollow form: a constant dictionary counterexample paired with a lambda that only reads one
+attribute or subscript from its parameter, including `.get(...)`. That form proves lookup
+distinguishes true from false, not that the predicate examines its stated observation. Beyond that
+narrow case, whether the alternative is the one that matters remains the author's claim and a
+reviewer's job.
 """
 import ast
 import glob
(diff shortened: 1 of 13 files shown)


MongLong0214:
## Review found this shallower than it claims. Not merging.

A blind adversarial pass, asked to refute. It did, and the sharpest finding is about the evidence mechanism itself rather than any one harness.

### The counterexamples are not counterexamples

> Many `falsifiable` calls first compute the real condition into `{"ok": value}` and then use `lambda observation: observation["ok"]` against a hand-authored `{"ok": False}`. That proves only that dictionary lookup distinguishes true from false; the underlying assertion can be replaced with constant `True` and the counterexample still passes.

That is correct and it undercuts what these harnesses claim to be. `falsifiable` exists so a predicate is shown **rejecting a counterexample**. Reducing the observation to a boolean before handing it over means the predicate never sees the measurement, and the counterexample never exercises it. The counterexample has to be a **raw observation shape** the real predicate rejects — the actual slider list with a description stripped, not `{"ok": False}`.

### `blocking_modal` detects a window level, not modality

Two misses, both backed by this repository's own code:

- **Attached sheets.** Logic's New Track blocker is an `AXSheet` whose host reports `AXModal=false`, and `AccessibilityChannel+ModalReconcile.swift:603` already documents that asking only about top-level modality misses it. A document-modal sheet blocks the document while `File > Open` stays enabled — so #369's negative control passes while its target is blocked.
- **Off-screen and other Spaces.** `kCGWindowListOptionOnScreenOnly` returns `None` for a running modal on an inactive Space.

And in the other direction, any modeless Logic panel assigned that level is a false positive.

### `isBlockingDialogWindow` rejects a genuinely blocking window before it reads `AXModal`

Logic 12.3's **Go To Position** window is recorded in this tree as `AXFloatingWindow` with `AXModal=true` (`AccessibilityChannel+Transport.swift:1246`). The new predicate only admits `AXDialog`/`AXSystemDialog`, so it never reaches the `AXModal` read. My "unreadable is fail-closed" claim is also inconsistent: an unreadable **subrole** returns harmless, and an unreadable window list leaves `blocking_dialog_present=false`.

### The modal is sampled after the observation, and a missing field reads as clean

`check`/`falsifiable` snapshot the modal *when the record is written*, not when the AX probe ran. A modal that covers the probe and closes before the record yields a clean count. Worse, `summarize` treats a record with **no** `blocking_modal` field exactly like an explicit `None`, so a schema gap counts as clean rather than invalidating the run.

### The witness widens silently

`menuItems(for:)` searches below the requested View button and, if that is empty, **falls back to every menu item under the application**. An identically named item in another plug-in window can be selected and the receipt does not disclose the fallback. That is the exact silent widening I removed from the mixer scoping, still present one function away.

It also erases AX status everywhere: `attr` drops the return code, `children` turns an error into `[]`, `description` turns an error into `""`, `valueIsSettable` turns an error into `false`. So **26 sliders whose description reads all FAILED satisfy "no slider has an AXDescription"** — #306's central assertion can pass on broken reads.

### And the health check passes on transport failure

`Driver._body(None)` returns `{"_transport_error": None}`, which is truthy, so "the release artifact answers a read-only request" is green when the transport is dead.

### Prose still carrying stale numbers

The 301 harness's opening docstring still says 24 sliders while its own note says 26. The 306 prose says both Compressor instances have 11 settable while the measurements are 22/11 and 20/10. The roadmap still describes 8 enables and 24 sliders. `evidence.py` says a modal blocks Logic "as a whole", which a document-modal sheet does not.

### What I am fixing

Raw-observation counterexamples across all three harnesses; sheets and window-list scope in `blocking_modal`; `AXFloatingWindow` in the Swift predicate and a consistent unreadable policy; sampling the modal at observation time and treating a missing field as unclean; removing the menu fallback; preserving AX status in the witness so a failed read is distinguishable from an absent value; a health predicate that rejects `_transport_error`; and the stale numbers.

MongLong0214:
## All six findings fixed, and one of them changed a measurement

Head `ee928f53`. Ship gate **PASS** with live evidence bound to that head.

### The counterexamples are counterexamples now

Fifteen calls rewritten to the idiom `live_290` already used: raw observation in, real predicate, and a counterexample **derived from the real observation with one field mutated**. **24 counterexamples across the three harnesses, all rejected.**

### The ratchet had certified the defect

`check-falsifiable-adoption.py` counted syntactic occurrences, so raising its floor from 1 to 4 this morning was raising it on hollow calls. It now refuses to count a call whose counterexample is all constants and whose predicate is a boolean-wrapper lookup, and reports it with file and line:

```
hollow   live_369_export_menu_path_is_reachable.py:227 (constant counterexample + boolean-wrapper predicate)
```

Verified by injecting one hollow call and watching it get named, then removing it.

### A measurement changed, and the finding got stronger

Preserving AX status turned up something the old helper had erased. The Controls-view sliders do **not** return an empty `AXDescription` — **the read fails**, in two ways:

```
kAXErrorAttributeUnsupported  -25212   ×15   the element has no such attribute
kAXErrorNoValue               -25205   ×11   the attribute exists and holds no value
```

Not one of the 26 returns an empty string. Both codes are evidence of absence and both are accepted; **any other failure code is refused**, because a read that went wrong proves nothing either way. The conclusion is unchanged and better founded: no Controls-view slider carries a name of its own, so step 9 of `set_param_verified` cannot reach them.

This is precisely the hole the review named — 26 failed reads used to satisfy "no slider has a description".

### The rest

- **`blocking_modal`** gains an AX sheet signal beside the CoreGraphics level, and a third state for *cannot tell* distinct from *found none*. A check recorded with a missing or unknown snapshot now makes the run **unclean**, where before it counted as clean. The on-screen-list limitation is documented with its consequence rather than papered over.
- **`isBlockingDialogWindow`** treats `AXModal == true` as blocking regardless of subrole, so Go To Position (`AXFloatingWindow`) is caught. Unreadable means blocking at **every** read in the predicate and its caller, not one of them.
- **The witness** no longer falls back to every menu item in the application when its local search is empty, and `AXRead<Value>` carries `value` / `absent` / `failed` instead of collapsing all three.
- **The health assertion** rejects a body carrying `_transport_error`, which used to be truthy and green on a dead transport.
- Stale prose fixed: the 301 docstring's 24-vs-26, the 306 prose's "11 settable on both", `evidence.py`'s "blocks Logic as a whole".

### Results

```
live_301  14 / 14      live_306  16 / 16      live_369  10 / 10
counterexamples 24, none unrejected
recorded under a modal: 0   missing snapshot: 0   unknown: 0
```

Repo guards 15/15. Full suite 4010 tests.

MongLong0214:
## Correction — I had the two AX error constant names reversed

Verified against `AXError.h` and the Swift runtime rather than taking either side's word:

```
kAXErrorAttributeUnsupported = -25205
kAXErrorNoValue              = -25212
```

I wrote them the other way round in the comment above, in the harness, and in the head commit. **The numeric rule was right and every name attached to a number was wrong.** The measured Controls-view split is therefore:

```
kAXErrorNoValue               -25212   ×15   the attribute exists and holds no value
kAXErrorAttributeUnsupported  -25205   ×11   the element has no such attribute
```

The finding is unchanged: not one of the 26 Controls-view sliders returns an empty `AXDescription` — every read fails, in one of two ways, and both are absence. What changes is which name goes with which count.

## A second review round refuted five more of the six responses

- **Two counterexamples still degenerate.** The window-open checks in #301 and #306 blank `pressed_children` and `opened_windows`, so the predicate rejects on a `len(...) == 1` guard **before it ever tests the button's role or name**. A cheap structural guard doing the rejecting is the hollow pattern wearing a costume. They will be derived from the real observation with one meaningful field changed.
- **The ratchet recognises one spelling.** `lambda o: o.get("ok")` is the same defect and it counts, because the rule matches only a bare subscript AST. It also counts a call that cannot run — and the test I added asserts that as correct.
- **`_first_ax_sheet` swallows an unreadable role** and reports a sheet on any Logic window as blocking the observed one. An unreadable role in a sheet enumeration has to be cannot-tell, not a skip.
- **`isBlockingDialogWindow` is still inconsistent.** An absent `AXModal` is treated as blocking only for `AXDialog`/`AXSystemDialog`, so an `AXFloatingWindow` like Go To Position with an absent or wrong-typed value returns harmless — and `axAttribute` maps a successful-but-wrong-type read to absent, so that path includes an unreadable payload.
- **The witness still loses status and still widens.** `bypass_before`/`bypass_after` carry no status sibling, the failure log records the attribute and code but not which element failed, and `newWindows(since:)` accepts any sole new Logic window as the one the slot opened.

Plus stale numbers in the guard, the roadmap, and the 306 prose — which says "most" native sliders are settable where both measurements are exactly half.

This is the third round on this branch where the code held up better than what I wrote about it.

Issue #724, opened 2026-08-30
plugins.set_param_verified succeeds once per process, then a stuck plug-in menu wedges every AppleEvent


MongLong0214:
## Measured 2026-08-30

`plugins.set_param_verified` succeeds **once per process** and then poisons the application. Measured against the release artifact, from a clean screen:

```
popups at start: 0
call0  Compressor threshold = 55   state=A  observed=55   -> Logic popup windows after: 1
call1  Compressor threshold = 60   state=C  window_open_failed
```

A **successful** write leaves an open popup menu behind. Every later call then fails, and it gets worse: once that popup is up, **every AppleEvent to Logic times out with `-1712`**, so the front-document path cannot be read and calls degrade from `window_open_failed` to `project_identity_mismatch`. One stuck menu disables every verified write in the process, and any other operation that needs AppleScript.

The same sequence reproduces on the Channel EQ band write built on the same acquisition path, so this is the shared plug-in window opener, not one operation.

## Where it comes from

`openPluginWindowFromTargetSlot` builds its attempt ladder as:

```
rank 0 controls   (the open button)
the slot itself
every OTHER button in the slot
```

`rankedPluginSlotOpenControls` excludes only the bypass control. So the fallback tier contains the slot's list button, and that button's own advertised action says what it does:

```
AXButton «목록»  actions = ["AXPress", "Name:Legacy 플러그인으로 플러그인 메뉴 열기\n…"]
```

It opens the **plug-in chooser menu**. Pressing it can never open a plug-in window, and it leaves a modal popup on screen.

## What identifies the popup, without AppleScript

AppleEvents are exactly what the popup blocks, so it cannot be detected by asking Logic. It is a Logic-owned CoreGraphics window at the **popup-menu level**:

```
layer 101 = CGWindowLevelForKey(kCGPopUpMenuWindowLevelKey)
230 x 593
```

Escape closes it and AppleEvents recover immediately — verified by sending Escape and getting a document name back on the next try.

Note this is **not** at the modal-panel level (8), so the modal detector added in #723 does not see it. Blocking comes in at least three shapes, at three different levels: alerts at 8 with `AXModal=true`, popups and menus at 101 with no entry in the AX window list, and sheets attached to a host whose own `AXModal` reads false.

## What a fix has to do

1. **Never press a control that opens a menu.** Exclude any control advertising `AXShowMenu` or a named menu-opening action. Match on the action, not the label — the label is localised and the action is not.
[...]

PR #725, opened 2026-08-30
feat(#301): verified Channel EQ band control by walking the slider

### LINKED ISSUE #301: ADR-013: Channel EQ — finish truthful state readback and bounded verified-write scope

MongLong0214:
Closes #301.

## The recorded wall was not where it was said to be

ADR-013 said *"Live census showed the Channel EQ Editor/Controls view can be AX-opaque, so extending the existing AX slider path is not reliable enough"*, and built a whole Decision on it: a virtual Mackie C4 control surface, two virtual MIDI endpoints, a `ChannelEQC4Adapter`, a doctor capability.

That census is `Scripts/spike-channel-eq-au-census.swift`, which instantiates the Audio Unit **outside Logic**. ADR-018's own Important Boundary says why that cannot answer the question: *"the MCP server is not the host attached to Logic's live AU instance."* It measured a different object, and the answer had been blocking this issue since.

Measured on live Logic 12.x: Channel EQ's native view exposes **8 named band enables and 24 named band-parameter sliders**, each reporting settable and each carrying an `AXValueDescription` in engineering units. The ADR's Decision was revised on the issue; no control surface is needed.

## The write is a walk, not a set

This is the part that changed the design, and it is why the catalog stayed empty until now.

```
actions on a band slider:  ["AXIncrement", "AXDecrement"]      ← no direct-set action
original 262 (+2.2 dB), requested 222
AXUIElementSetAttributeValue → status 0 (success), readback 261 (+2.1 dB)
```

**It reports success and moves one step.** Six consecutive sets to the same target walked 262 → 256, exactly one per call. So `AXValue` on these sliders is a direction, not a destination, and the single set Compressor `threshold` uses — where `set 60` lands on 60 — cannot be reused. `AXIncrement` is not one step either: thirty of them drove the control to its maximum.

`SliderIncrementWalk` is the pure core, free of AX, driven by an injected read and nudge. Ten edge cases, each with a case that can fail: already at target nudges nothing; saturation, oscillation and an unchanged display all report `noProgress` rather than looping; a rejected nudge stops immediately; a lost readback is its own outcome; crossing the target and moving away is `overshot`; the budget is exact; a step larger than one is fine because the loop is driven by the readback; a non-positive budget never nudges. **An out-of-range target is not clamped** — it terminates on saturation, because clamping would report arrival at a value nobody asked for.

## Engineering units without inventing a mapping

The raw-to-Hz relationship is non-linear — 250 renders `100 Hz`, 850 renders `5000 Hz` — so a Hz or dB target cannot be computed. A display request instead walks until **Logic's own `AXValueDescription`** equals the string asked for. Logic renders it; the walk converges on what it rendered. Arrival is exact string equality.

Choosing the direction needed a second pass. The first version always probed upward and then froze on whatever that probe did, so a target below the current value walked 179 steps into the rail. It now probes once, compares the two renderings against the target, reverses if the step moved away, and keeps re-evaluating. Ordering comes from the leading signed number with the unit text required to match on both sides; where that cannot be established the walk returns `noProgress` rather than guessing.

## Live results

```
raw   302 requested → 302 observed, "+6.2 dB"      40 steps   state A
dB    "+2.2 dB" from 302  (downward)                42 steps   state A
dB    "+6.2 dB" from 262  (upward)                  40 steps   state A
restore 262 → "+2.2 dB"                             40 steps   state A
```

Harnesses at this head, each from a closed-window start:

| harness | |
|---|---|
| `live_301_channel_eq_bands_are_addressable` | **15/15** |
| `live_301_eq_band_write_lands_and_rolls_back` | **18/18** |
| `live_290_selectors_resolve_by_identity` | **13/13** |

All `is_clean`, every counterexample rejected.

## A shipped defect this exposed — see #724

While proving the walk, `plugins.set_param_verified` turned out to succeed **once per process**. A successful write left an open plug-in chooser menu, the next call failed `window_open_failed`, and once that popup was up **every AppleEvent to Logic timed out with -1712**, so later calls degraded to `project_identity_mismatch`. One stuck menu disabled every verified write in the process.

The attempt ladder's fallback tier contained the slot's list button, whose own advertised action says it opens a menu. Controls advertising a menu-opening action are excluded now, matched on the action rather than the localised label, and acquisition leaves no popup on either path — detected through CoreGraphics at the derived popup-menu level, because AppleEvents are exactly what a stuck popup blocks.

Verified after the fix: three consecutive `set_eq_band_verified` calls all reach State A with zero popups, and so do three consecutive `set_param_verified` calls.

## Scope, stated plainly

- Band **write and readback** for Channel EQ, addressed by band and parameter **name**. No ordinals, no coordinates.
- The catalog declares the measured ranges — Frequency 0…1050, Gain 0…480, Peak/Cut Q 0…127, **Shelf Q 0…52**, Order 0…5. The shelf max genuinely differs and flattening it would put a valid request outside the control's travel.
- Registering the verb moved seven pinned counts across five test files, which is what pinning is for.
- Not covered: other EQ plug-ins, third-party plug-ins, and any locale beyond the Korean labels measured here — another locale resolves nothing and the operation refuses rather than acting on a guess.

## The #724 fix took three wrong guesses, and the measurements are worth keeping

Each guess was refuted by a measurement, and the third one was the answer.

**Guess 1: "demotion cannot work live."** My probe cleared `AXMain` and `AXFocused`, saw the window
still front, and concluded Logic ignores the demote. Logic *does* ignore `AXFocused` — but the probe
skipped the `AXRaise` fallback, which is what actually demotes:

```
clear AXMain / AXFocused   -> isFront still true
AXRaise(arrange window)     -> isFront FALSE
```

A probe that reproduces only part of the production path is not a statement about production. I had
written that conclusion into a worker packet as "measured"; the worker built on it and correctly
stopped rather than inventing a discriminator, which is the only reason a wrong fix did not land.

**Guess 2: "a `.unique` match is safe to reuse directly."** Both fail-closed tests went red
immediately. The planted wrong window carries a `Threshold` slider too, so `pluginWindowMatch`
returns `.unique` for it. **An element's presence is not proof of identity.** Reverted.

**Guess 3, and the answer: the open control is a TOGGLE.**

```
press 열기          window opens, isFront true
demote              isFront false
press 열기 AGAIN    the window is GONE from AXWindows — it CLOSED
```

Acquisition demoted an already-open window and then pressed the slot to prove it came back, which
closed it instead. So the fix is narrower than either guess: after a demote succeeds on a `.unique`
match the window is already the right one and is **raised**, not re-opened, and a raise that does
not front it falls through to the unchanged ladder. The `.none` and `.ambiguous` paths — which is
what both fail-closed tests exercise — are untouched, and demotion stays as the discriminator it
always was.

Live acceptance, four consecutive calls each, from a closed-window start:

```
set_eq_band_verified   272 → 282 → 292 → 262    all State A, window open throughout
set_param_verified      55 →  65 →  58 →  60    all State A   (the shipped operation)
```

---

## Also in this branch: #726, a wrong-insert write that reported success

Filed and fixed while verifying the above. `plugins.set_param_verified` could write to a
different insert than the one requested and still return State A with the requested insert
in `target_identity`. Reproduced on a real project: `Absolute Zero` carries Compressor at
inserts 0 and 2; writes addressed to insert 0 and insert 2 both landed on slot 0.

Undetectable downstream — readback comes from the same wrong window, so it agrees with the
request and `verified` stays true.

Two independent causes, both fixed here.

### 1. Two editors for the same plug-in are indistinguishable

Every AX attribute of the two Compressor editors agrees: AXTitle, AXSubrole, AXParent,
AXSections, header texts. Only screen geometry differs, and geometry is forbidden here.
`plugin_insert_ref` does not bridge it either — it is an inventory-side fingerprint over
(track, insert, identity) with no link to a window.

So the duplicated case refuses with `ambiguous_plugin_instance` before any window is opened
or pressed, naming the colliding inserts. It runs after the inventory-completeness gate, so
an unreadable slot cannot hide a duplicate.

### 2. The window match never checked the plug-in

It keyed on track name plus one slider AXDescription. Compressor, Noise Gate, Expander,
DeEsser and Limiter all expose a `Threshold`, so a Compressor write could match a Noise Gate
editor on the same track. A plug-in editor does name its plug-in, in its direct AXStaticText
children — measured across four editors on one track:

```
ChromaVerb  ["보기:", "ChromaVerb", "Studio Grand"]
Compressor  ["보기:", "사이드 체인:", "Compressor", "Studio Grand"]
Channel EQ  ["보기:", "Channel EQ", "Studio Grand"]
Piano       ["보기:", "Studio Piano", "Studio Grand"]     <- slot says `Piano`
```

That last row is why the check resolves through `VerifiedPluginCatalog` rather than comparing
strings. The name survives the Controls view, stays English while surrounding labels are
localized, and sits at no fixed index (13 in one window, 42 in another). A window with no
readable static text is refused; absence is not identity evidence.

### What is proven live, and what is not

Live: the duplicated track refuses with `write_attempted: false` and no window opened (0
plug-in editors before and after); the single-instance path still reaches State A.

Not live: no track in the test project has two plug-ins sharing a slider AXDescription, so
`plugin_window_plugin_mismatch` is pinned by unit tests with injected AX only. Third-party
editors are unmeasured.

## And #724, the acquisition defect

`set_param_verified` succeeded once per process, then wedged. Three stacked causes: the
acquisition ladder pressed the menu-opening list button; the resulting popup sits at
CGWindowLayer 101 and blocked every AppleEvent with -1712; and the slot's open control is a
toggle, so re-pressing it closed the window instead of raising it. All three fixed, with the
popup cleanup reporting `window_count_unavailable` rather than treating an unreadable count
as a clean screen.

### DIFF
--- a/Package.resolved
+++ b/Package.resolved
@@ -1,15 +1,6 @@
 {
-  "originHash" : "6c926ace4eed5db8b503d612fccba4dff0e0122b4c622d86bdc4e8e972219044",
+  "originHash" : "556ef8452b5e29e7d89a39cb4ccd602017b20d5779c03e16152b41fa1d5339a6",
   "pins" : [
-    {
-      "identity" : "async-http-client",
-      "kind" : "remoteSourceControl",
-      "location" : "https://github.com/swift-server/async-http-client.git",
-      "state" : {
-        "revision" : "3a5b74a58782c3b4c1f0bc75e9b67b10c2494e8f",
-        "version" : "1.33.1"
-      }
-    },
     {
       "identity" : "eventsource",
       "kind" : "remoteSourceControl",
@@ -17,33 +8,6 @@
       "state" : {
         "revision" : "a3a85a85214caf642abaa96ae664e4c772a59f6e",
         "version" : "1.4.1"
-      }
-    },
-    {
-      "identity" : "swift-algorithms",
-      "kind" : "remoteSourceControl",
-      "location" : "https://github.com/apple/swift-algorithms.git",
-      "state" : {
-        "revision" : "87e50f483c54e6efd60e885f7f5aa946cee68023",
-        "version" : "1.2.1"
-      }
-    },
-    {
-      "identity" : "swift-asn1",
-      "kind" : "remoteSourceControl",
-      "location" : "https://github.com/apple/swift-asn1.git",
-      "state" : {
-        "revision" : "a9a5efd40eaf558a2bcd48d64b1d1646be686008",
-        "version" : "1.7.1"
-      }
-    },
-    {
-      "identity" : "swift-async-algorithms",
-      "kind" : "remoteSourceControl",
-      "location" : "https://github.com/apple/swift-async-algorithms.git",
-      "state" : {
-        "revision" : "3da39bbc4e687d4192af7c9cf4eab805745a0b9c",
-        "version" : "1.1.5"
       }
     },
     {
@@ -56,57 +20,12 @@
       }
     },
     {
-      "identity" : "swift-certificates",
-      "kind" : "remoteSourceControl",
-      "location" : "https://github.com/apple/swift-certificates.git",
-      "state" : {
-        "revision" : "89fbc3714264cce8db8e4ec51b64e01c3e28c6c5",
-        "version" : "1.19.3"
-      }
-    },
-    {
       "identity" : "swift-collections",
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/apple/swift-collections.git",
       "state" : {
         "revision" : "6675bc0ff86e61436e615df6fc5174e043e57924",
         "version" : "1.4.1"
-      }
-    },
-    {
-      "identity" : "swift-crypto",
-      "kind" : "remoteSourceControl",
-      "location" : "https://github.com/apple/swift-crypto.git",
-      "state" : {
-        "revision" : "47d3869a7291f085c1fb9fb1e6d3b97a793f45c6",
-        "version" : "4.5.1"
-      }
-    },
-    {
-      "identity" : "swift-distributed-tracing",
-      "kind" : "remoteSourceControl",
-      "location" : "https://github.com/apple/swift-distributed-tracing.git",
-      "state" : {
-        "revision" : "dc4030184203ffafbb2ec614352487235d747fe0",
-        "version" : "1.4.1"
-      }
-    },
-    {
-      "identity" : "swift-http-structured-headers",
-      "kind" : "remoteSourceControl",
-      "location" : "https://github.com/apple/swift-http-structured-headers.git",
-      "state" : {
-        "revision" : "933538faa42c432d385f02e07df0ace7c5ecfc47",
-        "version" : "1.7.0"
-      }
-    },
-    {
-      "identity" : "swift-http-types",
-      "kind" : "remoteSourceControl",
-      "location" : "https://github.com/apple/swift-http-types.git",
-      "state" : {
-        "revision" : "db774a277f60063a32d854f2980299caf06da041",
-        "version" : "1.6.0"
       }
     },
     {
@@ -128,75 +47,12 @@
       }
     },
     {
-      "identity" : "swift-nio-extras",
-      "kind" : "remoteSourceControl",
-      "location" : "https://github.com/apple/swift-nio-extras.git",
-      "state" : {
-        "revision" : "88a51340f59cf181ebde888bd1b749296b3ec029",
-        "version" : "1.34.3"
-      }
-    },
-    {
-      "identity" : "swift-nio-http2",
-      "kind" : "remoteSourceControl",
-      "location" : "https://github.com/apple/swift-nio-http2.git",
-      "state" : {
-        "revision" : "61d1b44f6e4e118792be1cff88ee2bc0267c6f9a",
-        "version" : "1.44.0"
-      }
-    },
-    {
-      "identity" : "swift-nio-ssl",
-      "kind" : "remoteSourceControl",
-      "location" : "https://github.com/apple/swift-nio-ssl.git",
-      "state" : {
-        "revision" : "3f337058ccd7243c4cac7911477d8ad4c598d4da",
-        "version" : "2.37.0"
-      }
-    },
-    {
-      "identity" : "swift-nio-transport-services",
-      "kind" : "remoteSourceControl",
-      "location" : "https://github.com/apple/swift-nio-transport-services.git",
-      "state" : {
-        "revision" : "67787bb645a5e67d2edcdfbe48a216cc549222d5",
-        "version" : "1.28.0"
-      }
-    },
-    {
-      "identity" : "swift-numerics",
-      "kind" : "remoteSourceControl",
-      "location" : "https://github.com/apple/swift-numerics.git",
-      "state" : {
-        "revision" : "0c0290ff6b24942dadb83a929ffaaa1481df04a2",
-        "version" : "1.1.1"
-      }
-    },
-    {
       "identity" : "swift-sdk",
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/modelcontextprotocol/swift-sdk.git",
       "state" : {
         "revision" : "a0ae212ebf6eab5f754c3129608bc5557637e605",
         "version" : "0.12.1"
-      }
-    },
-    {
-      "identity" : "swift-service-context",
-      "kind" : "remoteSourceControl",
-      "location" : "https://github.com/apple/swift-service-context.git",
-      "state" : {
-        "revision" : "d0997351b0c7779017f88e7a93bc30a1878d7f29",
-        "version" : "1.3.0"
-      }
-    },
-    {
-      "identity" : "swift-service-lifecycle",
-      "kind" : "remoteSourceControl",
-      "location" : "https://github.com/swift-server/swift-service-lifecycle.git",
-      "state" : {
-        "revision" : "9829955b385e5bb88128b73f1b8389e9b9c3191a",
-        "version" : "2.11.0"
       }
     },
     {
--- a/Package.resolved
+++ b/Package.resolved
@@ -1,6 +1,15 @@
 {
-  "originHash" : "556ef8452b5e29e7d89a39cb4ccd602017b20d5779c03e16152b41fa1d5339a6",
+  "originHash" : "6c926ace4eed5db8b503d612fccba4dff0e0122b4c622d86bdc4e8e972219044",
   "pins" : [
+    {
+      "identity" : "async-http-client",
+      "kind" : "remoteSourceControl",
+      "location" : "https://github.com/swift-server/async-http-client.git",
+      "state" : {
+        "revision" : "3a5b74a58782c3b4c1f0bc75e9b67b10c2494e8f",
+        "version" : "1.33.1"
+      }
+    },
     {
       "identity" : "eventsource",
       "kind" : "remoteSourceControl",
@@ -8,6 +17,33 @@
       "state" : {
         "revision" : "a3a85a85214caf642abaa96ae664e4c772a59f6e",
         "version" : "1.4.1"
+      }
+    },
+    {
+      "identity" : "swift-algorithms",
+      "kind" : "remoteSourceControl",
+      "location" : "https://github.com/apple/swift-algorithms.git",
+      "state" : {
+        "revision" : "87e50f483c54e6efd60e885f7f5aa946cee68023",
+        "version" : "1.2.1"
+      }
+    },
+    {
+      "identity" : "swift-asn1",
+      "kind" : "remoteSourceControl",
+      "location" : "https://github.com/apple/swift-asn1.git",
+      "state" : {
+        "revision" : "a9a5efd40eaf558a2bcd48d64b1d1646be686008",
+        "version" : "1.7.1"
+      }
+    },
+    {
+      "identity" : "swift-async-algorithms",
+      "kind" : "remoteSourceControl",
+      "location" : "https://github.com/apple/swift-async-algorithms.git",
+      "state" : {
+        "revision" : "3da39bbc4e687d4192af7c9cf4eab805745a0b9c",
+        "version" : "1.1.5"
       }
     },
     {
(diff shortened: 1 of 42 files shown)


MongLong0214:
## Blind review found a reachable wrong-insert write. Not merging.

I asked a blind adversarial pass to refute eight claims. It refuted most of them, and one is a defect I introduced and then **measured on this machine**.

### W4 — a write can land on the wrong plug-in instance

> A reachable wrong-window acceptance exists when one track has two instances of the same plug-in with the same named parameter. If insert A's editor is the sole matching open window while insert B is targeted, the new `.unique` branch demotes and raises A, then accepts it. The later identity check proves only that B's slot still contains the expected plug-in; it does not prove that the raised editor belongs to B.

`pluginWindowMatch(forTrackName:matchingSliderDescription:)` identifies a window by **track name + slider description**. Two Compressors on one track produce the same match for both, so the reuse branch I added cannot tell them apart.

This is not hypothetical. The project open here has exactly that shape:

```
Absolute Zero:  insert 0 Compressor    insert 2 Compressor
```

Measured, targeting each insert in turn and then reading both back directly through AX:

```
set_param_verified insert=0 value=40  -> state A, target_identity.insert = 0
set_param_verified insert=2 value=70  -> state A, target_identity.insert = 2

read back:   slot 0 Threshold = 0 %      ← every write landed here
             slot 1 Threshold = 56 %     ← never touched
```

**Both writes went to the first slot while the envelope reported the requested insert.** That is a silent wrong-target write reported as State A, which is the failure class this project's fail-closed discipline exists to prevent. Filed separately; this PR does not merge until it refuses instead.

My commit message also claimed the two fail-closed tests exercise the untouched `.none`/`.ambiguous` paths. The review says they begin at `.unique`, so that sentence is wrong and the tests do not cover the acceptance path I added.

### Also real, and being fixed here

- **W5** — the popup cleanup treats a nil CoreGraphics count as `noPopupObserved`, whose outcome reads clean. Failure conflated with absence, again. It also cancels every visible Logic menu and posts Escape without confirming Logic is frontmost, so it can dismiss a menu a user opened and send Escape somewhere else.
- **W6** — the catalog records `no_write_round_trip_claimed` while the entry declares `parameterWriteReadback`, and the test named `...WithoutClaimingAWriteRoundTrip` asserts that capability. Units are accepted case-insensitively but reused verbatim to build the exact target string, so an accepted `db` can never equal Logic's `dB`.
- **W3** — the walk requests `current.value ± 1`, so the comment claiming no one-step assumption is inaccurate. The loop is driven by readback, but the request is not.
- **W8** — "State A only after an actual AXValue write" is contradicted by the zero-step path, where a target already at value returns State A with no nudges. And this PR's own title claims more than the evidence: one parameter path was proven live, not all 24, and not exact unit rendering.
- **`testConsecutiveEQBandWritesReuseTheVerifiedOpenEditor` is decoration.** Its fake raise never fronts the editor and slot pressing does not toggle it closed, so deleting the raise path leaves it green. The #724 fix is currently pinned only by the live run.

### One correction to the review

It reports the addressability harness proves only non-empty `AXValueDescription`, and that is right — the declared `Q` unit and exact formatting are not substantiated by it. I will narrow the catalog's unit claims rather than widen the harness's, because the rendering was measured for Hz and dB and inferred for Q.

Issue #736, opened 2026-09-01
Live QA: MCU channel never detects feedback + duplicate virtual MIDI ports created on every launch (v3.13.0)


ElDavichin:
### Preflight

- [x] I checked whether this is already covered by an existing issue.
- [x] I separated product behavior from demo/rendering artifacts.

### Priority

P1 - public claim, core workflow, or verified-state blocker

### Live context

- Date/time: 2026-09-01
- Logic Pro version: 12.3 (desktop, com.apple.logic10)
- Project/session state: "Untitled - Tracks" project open, 1 track
- Capture or E2E script: Manual testing via MCP tool calls (logic_system.health, logic_midi.list_ports)
- Transcript/provenance path: N/A (interactive session)
- Related media artifact: N/A

### Observed outcome

Mismatch between response and live UI/resource

### Observed evidence

Repeated `logic_system health` output (unchanged across all attempts):
"mcu": { "connected": false, "registered_as_device": false, "feedback_stale": false, "port_name": "LogicProMCP-MCU-Internal" }

Confirmed correct Logic-side wiring: Control Surfaces > Setup shows Mackie Control installed, both Input and Output ports set to LogicProMCP-MCU-Internal.

Tried with no change: Control Surfaces > Rescan All MIDI Drivers, full Logic Pro quit/reopen, restarting the logic-pro MCP connector, full macOS restart.

Also observed via `logic_midi.list_ports`: every LogicProMCP virtual port (KeyCmd-Internal, MCU-Internal, Scripter-Internal, MIDI-Internal) appears TWICE in `sources`, immediately after every fresh process start — including right after a full computer reboot. This caused intermittent delivery failures when using `send_cc`/`send_note` with port:"keycmd" during manual MIDI Learn (some sends never reached Logic's Learn listener).

### Claim boundary

Until this is fixed, do not claim MCU-backed mixer verification (e.g. set_master_volume's State A readback) is achievable on desktop Logic Pro 12.3 — registered_as_device/connected never transitions to true even with correct Control Surfaces setup. Also do not claim reliable delivery of keycmd-port MIDI sends (send_cc/send_note with port:"keycmd") for MIDI Learn workflows, since duplicate virtual MIDI port creation on every server start causes intermittent, non-deterministic delivery.

### Done when

- [ ] Root cause of MCU registered_as_device staying false is identified or reproduced.
- [ ] Root cause of duplicate virtual MIDI port creation on startup is fixed.
[...]

Issue #291, opened 2026-07-10
ADR-008: Verified routing graph and exact bus/output/send repairs — selected session-repair foundation


MongLong0214:
# ADR-008 — Observe and repair exact routes, not labels that resemble routes

> **Execution SSOT r4 — 2026-09-22. OPEN / SELECTED for the session-repair scenario. R0 is READY; later writes require their actual observation prerequisites.** The owner requested the real-world routing/cleanup feature, activating R0–R3 of this existing ADR rather than creating a duplicate. #308 schedules; #971 owns end-to-end acceptance. Source checked at `ffcf0bf8ac0f93a7615fdce6b57d0b2fd5370575`. Historical measurements remain scoped, not a fresh live qualification. No current #964/v3.17.0 release dependency is added.

## 1. Existing implementation and responsibility boundary

`logic://mixer` already publishes a partial `routing_graph` through `RoutingGraphPublication`. `RoutingGraph`, `RoutingDiff` and `RoutingWriteGate` exist. Reuse them and the current target registry, trace, UI ownership and resource readers. Do not create another graph/registry to satisfy the new scenario.

First fix the demonstrated source-reference read-order defect. Then provide real track/strip/bus/aux/physical endpoint relationships and exact output/input/send writers. A model, guard or temporary menu experiment is not a public verified writer.

| Boundary | Owner |
|---|---|
| Source/project call-order defect, graph/endpoint/edge readback and scalar routing operations | **This issue** |
| Whole-session track/strip population, association and layout scope | #965 / ADR-020; share one actual observation and reference model |
| Intended topology, ambiguity questions and approved minimal repair plan | #966 / ADR-021 |
| Bus allocation, explicit aux creation/reuse and cleanup of owned new objects | #967 / ADR-022 |
| Naming, layout and color | #968, #969, #970 |
| Approved multi-step composition and independent final report | #971 / ADR-026, using existing Saga/trace machinery |

The old no-aux-creation initial boundary remains true **for these primitive writers**. The newly requested authorized lifecycle is explicitly added under #967. No implicit aux creation may be ignored as a side effect. R4 sidechain-source mutation remains later backlog; existing sidechain relationships must still be covered or marked unknown when assessing affected-path safety.

## 2. Implementation history that must survive

| Evidence | Finding and limit |
|---|---|
| #740 | Source identities reuse the existing target registry. Reading infrastructure is not entirely unimplemented. |
[...]

PR #738, opened 2026-09-02
fix(#736): stop creating a second endpoint that already answers to our name

MongLong0214:
Fixes the duplicate virtual MIDI endpoints reported in #736, and makes health say which situation it is in.

### What the reporter established

Two live server processes is the **normal** state for a host that opens one subprocess per client connection to a single stdio config entry. Nothing in the connector configuration can prevent it, and a reconnect on the host side relaunches a subprocess — so killing one instance produced a fresh duplicate seconds later. That path, not a manual double-launch, is what real users hit.

Both instances created virtual endpoints with the same fixed names, unconditionally, with no census and no report. Logic binds a control surface by name; with two endpoints answering to `LogicProMCP-MCU-Internal` it binds one of them, and if that is not the one the answering server reads from, health correctly reports `connected: false` while the wiring looks right.

Four data points across two of the reporter's sessions, both directions:

```
duplicate present  →  connected: false   (x2)
single endpoint    →  connected: true    (x2)
```

— arrived at after a control-surface reconfiguration, MIDI driver rescans and a full reboot had all failed to help, every one of them performed while unknowingly duplicated.

### Changes

- Endpoints get a deterministic, direction-aware `kMIDIPropertyUniqueID` derived from the port name, instead of the random one CoreMIDI hands out per launch. A collision fails loudly rather than falling back to a random identity.
- Before creating, census CoreMIDI for endpoints carrying one of our names. If a foreign one exists, **do not create a twin**. The later instance runs degraded and the first keeps working.
- `logic_system health` publishes `mcu.port_census.endpoint_count` and `has_foreign_endpoint`, so `connected: false` finally distinguishes "Logic is not set up" from "another process owns a port with my name". That ambiguity is what cost the reporter a reconfiguration, a rescan and a reboot.

### Verified live, with a negative control

```
two OLD binaries   SOURCES 9   every LogicProMCP name duplicated x2
two NEW binaries   SOURCES 5   each name exactly once
both exited        SOURCES 1   cleaned up

unique IDs across three separate restarts: identical every time
  MCU source 553522772, MCU destination 1265798057 (distinct per direction)
```

### Known limitation, stated rather than glossed

The first instance to start owns the ports, and if it exits the degraded one does not take over — it never created anything to take over with. The reporter's topology makes that the wrong instance: their stable connection starts first while the churny one is what actually sends MIDI (MIDI Learn via `send_cc`, and MCU is two-way by definition).

The fix for that is to defer port creation until an instance issues its first MIDI operation. Their from-zero measurement showed it is safe — a port appearing while Logic is already running, with a control surface previously configured against that name, binds with no rescan. That is a follow-up; this change is the half that is correct under either ownership rule.

**Not claimed:** this does not fix MCU registration for a setup that has never configured the control surface. It removes the duplicate that was breaking one that had.

Refs #736


MongLong0214:
**Do not merge yet.** An adversarial review refuted three of this PR's four claims and found six defects, two of them High. Recording them here so the PR body is not left standing as if it were true.

**The claim that a second instance never creates a twin is false in a race.** The census completes at `MIDIPortManager.swift:119` and creation happens separately at `:122`/`:169`; `MIDIEngine.swift:135`/`:141` has the same gap. Two processes can both observe zero, both pass the guard, and both create. With two *new* binaries the deterministic ID usually makes one lose and dispose — but the duplicate exists in between. Against an **older** process holding a random ID there is no collision at all, so both endpoints persist permanently. That is precisely the owner-harming duplicate this change is for.

**An enumeration failure is published as an observed zero.** `MIDIGetNumberOfSources` and `MIDIGetNumberOfDestinations` both return zero on error — CoreMIDI documents this — and `VirtualMIDIEndpointIdentity.swift:31` turns the result straight into a range. Unreadable endpoint names are silently filtered at `:53`. So a CoreMIDI service failure becomes `endpoint_count: 0, has_foreign_endpoint: false`: we attempt a twin, and health reports a zero nobody observed. The census type has no validity field.

This is the same defect class as the absent-vs-failed work in #369, arrived at from a different direction: **"I could not read" must never become "there is none."** An unknown census has to take the degraded path, because refusing to create is safe and creating a twin is not.

**Health republishes a startup snapshot.** `MCUChannel.swift:374`/`:389` record the census only at startup, and `MIDIPortManager.swift:204` deliberately returns the previous non-empty census when a later enumeration comes back empty. A degraded instance therefore keeps reporting `endpoint_count: 2, has_foreign_endpoint: true` after the owner exits and nothing is left; an owner that cached a clean census never reports a twin created afterwards. The claim that health "distinguishes" the two situations holds at exactly one instant.

**A degraded operation still does not name the conflict** — the thing this change was supposed to fix. `MIDIEngine.start()` throws the conflict but stores no cause, so `logic_midi send_cc` comes back as State C `channels_exhausted` with a hint saying only that CoreMIDI was not initialized. A caller reads that as broken MIDI hardware.

**"Mine" means "in this actor's reference set", not "owned by this process."** `MIDIEngine` and `MIDIPortManager` keep separate sets, so a same-process `midi.create_virtual_port` sees this server's own endpoint as foreign. And a disposal whose error is ignored at `:219` is still recorded as released at `:226`, so our own leaked endpoint is classified foreign on the next start.

**The unique-ID derivation is not injective, with concrete collisions.** It is 32-bit FNV-1a masked to 31 bits over an unbounded name domain. Reproducing the function found real pairs:

```
source kRPI1eQeasjZ   and   source Y65lWHfsr5sr   ->  257011563
```

The answer is not a better hash — you cannot make one injective over an unbounded domain — but making the `kMIDIIDNotUnique` refusal reachable and verifying the process is left sane rather than half-initialised. Those two names are going into the tests so the detection is pinned by an actual collision.

The live proof in the PR body still stands as far as it goes: two old binaries duplicate, two new ones do not, IDs are stable across restarts. What the review shows is that it was one arrangement, and the failure modes it does not cover are reachable. Fixes in progress; I will re-run the two-instance proof on the new head.

MongLong0214:
All six review findings are closed. Live proof re-run on the new head, with the negative control still failing as it must:

```
two OLD binaries   SOURCES 9   every LogicProMCP name duplicated x2
two NEW binaries   SOURCES 5   each name exactly once, destinations clean
both exited        SOURCES 1   cleaned up
```

**The race** — census and creation are now serialised with the advisory file lock this repository already uses elsewhere, held across both. The previous version could have two processes both observe zero and both create; against a peer holding a random unique ID there was no collision to lose, so both endpoints persisted.

**The enumeration failure published as an observed zero** — `MIDIGetNumberOfSources` and `MIDIGetNumberOfDestinations` both return zero on error, and unreadable names were silently filtered out, so a CoreMIDI service failure became `endpoint_count: 0, has_foreign_endpoint: false`. The census has an explicit unknown state now, an unreadable name is an unknown rather than a nonexistent endpoint, and an unknown census takes the degraded path — refusing to create is safe, creating a twin is not.

That is the same defect class as #369's absent-versus-unreadable work, reached from a different direction, and #291 met the third instance of it before it could become a defect. "I could not read" must never become "there is none."

**Health republished a startup snapshot** — it censuses at read time now, so a degraded instance stops claiming a foreign endpoint after the owner exits.

**A degraded operation did not name the conflict** — `send_cc` came back as `channels_exhausted` with a hint about initialization, which reads as broken MIDI hardware. The startup cause survives into the response now.

**"Mine" meant "in this actor's reference set"** — ownership is process-wide, and a disposal whose error was ignored no longer counts as released.

**The unique ID is not injective**, and the review proved it with real collisions rather than an argument: `kRPI1eQeasjZ` and `Y65lWHfsr5sr` both map to `257011563`. A hash over an unbounded domain cannot be made injective, so the answer is detection — `kMIDIIDNotUnique` is reachable and loud, and those two names are the test case, so the guard is pinned by an actual collision rather than a hypothetical one.

The known limitation in the PR body still stands and is unchanged: the first instance to start owns the ports, and deferring creation to first MIDI use is the follow-up that fixes which instance wins.

Issue #448, opened 2026-07-25
Track layout request — shipped sort preserved; remaining readback/color/reorder transferred to ADR-020/024/025


computerkoalas:
# Track-layout request — historical proposal and explicit requirement transfers

> **Maintainer reconciliation — 2026-09-22. This combined implementation queue is closed as transferred, not as full feature completion.** The original proposal and all comments remain historical evidence. Existing `tracks.sort_verified` is real delivered work; arbitrary exact reordering, full layout observation and verified color are not made complete by this issue's old `completed` state. Current implementation owners are below; #308 schedules them and #971 verifies the whole session-repair scenario.

## Current execution owners — every original requirement retained

| Original requirement | Current owner and boundary |
|---|---|
| Complete track/strip references, order, type, stack parent/depth/header/type/collapse and provenance | #965 / ADR-020. Whole requested population, not visible rows misrepresented as all tracks. |
| Actual palette/color reading on the intended track/strip | #970 / ADR-025 C1, exposed through #965's observation model only when measured. Unknown stays unknown. |
| Verified color assignment and expected-before check | #970 C2/C3. Actuation without independent target-color readback is not verified coloring. |
| Anchor-based move before/after, preserve parent/stack by default | #969 / ADR-024 L2/L3, including the still-unqualified non-coordinate provider. |
| Shipped menu-based verified sort | Preserve existing `tracks.sort_verified`; #969 L1 integrates it rather than rebuilding it. It is not arbitrary reorder. |
| Duplicate-name safety, hidden/collapsed/nested tracks, topology generation and reference rebind/remap | #965 observation plus #969 movement; naming continuity remains #285/#353 and #968. |
| Folder/summing-stack boundaries, region placement and unchanged routing under cosmetic permission | #969 with #965/#291 observation. Summing-related routing changes require explicit #966 plan and #291/#967 operations. |
| Naming alongside layout without losing references | #968 / ADR-023, reuse existing rename. No speculative deletion/merging. |
| Dry run, exact consent, multi-step execution, safe compensation and final before/after report | #966 planning and #971 composition, using current #287 Saga only at its actual supported scope. |
| State A/B/C, stale-target checks, live duplicated-vocal/dry-wet/master/stack fixture and public limits | Each scalar owner plus #971 end-to-end fixture. No action acknowledgement or mock-only result closes the actual requested feature. |

[...]

PR #746, opened 2026-09-02
feat(#448): verified track sort — the after-read joins by name, and it reaches State A live

MongLong0214:
## Do not merge until this has been driven live

This operation reorders the tracks of a real project and its happy path has never run against Logic. What blocks the drive on this machine is recorded below, not hand-waved. Everything else here is done: the design, the fixes from an adversarial review, and a live check of the fail-closed path.

## What it adds

`tracks.sort_verified` drives `Track > Sort Tracks by` and checks the effect by re-reading the arrangement order. The caller supplies `criterion`, `expected_order` (a complete list of already-issued `trk_` references) and `confirmed: true`.

The design decisions worth naming:

- **"The order changed" is not proof.** A sort by the wrong criterion also changes the order, so the after-order must equal the caller's complete expected order exactly.
- **An already-sorted project is State B, not State A.** A correct sort changes nothing there, and "nothing changed" cannot be distinguished from "the command did nothing".
- **`expected_order` takes references, not names.** Duplicating a track in Logic produces identically named tracks; an identity that cannot name two of the things it orders is not an identity.
- Criterion labels live in `AXLocalePolicy` with what was measured and when. An unmeasured locale refuses by name rather than guessing a translation.

Anchor-based reorder ("move track A before track B") stays out: it is drag-only as recorded, and a sort is not a substitute.

## What an adversarial review found, and what changed

Four ways State A could be awarded for something unproven, all closed:

- **A wrong criterion could pass.** The verifier discarded the requested criterion. The verdict is now bound to the menu leaf actually pressed: its title is read from that element, mapped to a criterion, compared with the request, and emitted as `actuated_menu_item_label` / `actuated_criterion`. The menu exposes no persistent selected-sort attribute and the comment says so.
- **A stale reference to a same-named track was silently rebound** to whichever header sat at its old index. `TargetDescriptor` carries only index and name, so a duplicate-name project now refuses before any AX work.
- **"Complete order" was not complete.** A collapsed stack takes the header rail from 44 rows to 21 by this repository's own measurement, so hidden tracks could go unrepresented and be reordered anyway. A sort whose scope cannot be enumerated is refused.
- **A failed press claimed nothing was written.** The action had already been sent. The order is now read back even on a failed press; when it cannot be read the response says the state is unknown, invalidates the affected references, and is not safe to retry.

With them: four verdict-driving read sites now keep `kAXErrorNoValue` and `kAXErrorAttributeUnsupported` as absence and name every other status as a read failure; the single read taken 150 ms after the press became a settle witness requiring two consecutive identical observations; a comment that claimed a live measurement was corrected; and three tests that could not fail were rewritten.

## What was driven live

The fail-closed path, against a release binary built from this head:

```
sort_verified criterion=track_name expected_order=[trk_0…trk_4] confirmed=true
  -> State C stale_target_reference, write_attempted: false
     "expected_order names track references that are unavailable or stale: trk_0, trk_1, …
      Re-read logic://tracks and retry with its current track_ref values."
```

Nothing was mutated. That is the behaviour the contract promises for unknown references, observed rather than asserted.

## Why the happy path is still unproven, precisely

Two conditions on this machine, both measured:

1. `logic://tracks` currently answers `readable: false`, `reason: track_names_synthesised_from_project_file`, with placeholder names, after an explicit `system.refresh_cache`. A caller therefore cannot obtain the real `trk_` references the operation requires.
2. The available fixture project has three tracks named `Studio Grand`, and duplicate names are exactly what this change now refuses before touching AX.

A first live drive needs a project whose track resource reads live and whose track names are unique. It has to check what the unit tests cannot: that the menu leaf title read back after the press really identifies the criterion, that the settle witness sees a stable order rather than a transient one, and that Logic does not recreate header objects in a way that turns every real sort into State B.

Full suite and repository guards pass; live evidence for this head is the export-menu harness, which proves the artifact answers a live Logic and is not about the sort.

Refs #448

### DIFF
--- a/Sources/LogicProMCP/Accessibility/AXLocalePolicy.swift
+++ b/Sources/LogicProMCP/Accessibility/AXLocalePolicy.swift
@@ -327,6 +327,63 @@ enum AXLocalePolicy {
         rationale: "Top-level menu titles expose no stable AXIdentifier in Logic."
     )
 
+    /// #448 — Track > Sort Tracks By. Measured on 2026-09-02 on Logic Pro
+    /// 12.3 with a Korean UI as `트랙 › 트랙을 다음으로 정렬`. The menu bar is
+    /// only partly localized, so this policy deliberately has no inferred EN/JA
+    /// form: a locale without this measurement must refuse.
+    static let sortTracksByMenuItem = LabelSet(
+        canonical: "트랙을 다음으로 정렬",
+        variants: [],
+        rationale: "Measured 2026-09-02 on Korean Logic Pro 12.3: Track > Sort Tracks By submenu title. No other locale is measured; never translate this menu label."
+    )
+
+    static let sortTracksMenuPath = MenuPath(
+        bar: trackMenuBar,
+        item: sortTracksByMenuItem
+    )
+
+    static let sortTracksByMIDIChannelMenuItem = LabelSet(
+        canonical: "MIDI 채널",
+        variants: [],
+        rationale: "Measured 2026-09-02 on Korean Logic Pro 12.3 as a Track > Sort Tracks By leaf; no other locale is measured."
+    )
+
+    static let sortTracksByAudioChannelMenuItem = LabelSet(
+        canonical: "오디오 채널",
+        variants: [],
+        rationale: "Measured 2026-09-02 on Korean Logic Pro 12.3 as a Track > Sort Tracks By leaf; no other locale is measured."
+    )
+
+    static let sortTracksByOutputChannelMenuItem = LabelSet(
+        canonical: "출력 채널",
+        variants: [],
+        rationale: "Measured 2026-09-02 on Korean Logic Pro 12.3 as a Track > Sort Tracks By leaf; no other locale is measured."
+    )
+
+    static let sortTracksByInstrumentNameMenuItem = LabelSet(
+        canonical: "악기 이름",
+        variants: [],
+        rationale: "Measured 2026-09-02 on Korean Logic Pro 12.3 as a Track > Sort Tracks By leaf; no other locale is measured."
+    )
+
+    static let sortTracksByTrackNameMenuItem = LabelSet(
+        canonical: "트랙 이름",
+        variants: [],
+        rationale: "Measured 2026-09-02 on Korean Logic Pro 12.3 as a Track > Sort Tracks By leaf; no other locale is measured."
+    )
+
+    static let sortTracksByUsedMenuItem = LabelSet(
+        canonical: "사용 여부",
+        variants: [],
+        rationale: "Measured 2026-09-02 on Korean Logic Pro 12.3 as a Track > Sort Tracks By leaf; no other locale is measured."
+    )
+
+    static let sortTracksByCreationDateMenuItem = LabelSet(
+        canonical: "생성일",
+        variants: [],
+        rationale: "Measured 2026-09-02 on Korean Logic Pro 12.3 as a Track > Sort Tracks By leaf; no other locale is measured."
+    )
+
     /// #519: File > Save As…
     ///
     /// The English label was MEASURED on 2026-08-19 by enumerating the File menu on a live Logic
@@ -1501,6 +1558,14 @@ enum AXLocalePolicy {
         editMenuBar,
         navigateMenuBar,
         trackMenuBar,
+        sortTracksByMenuItem,
+        sortTracksByMIDIChannelMenuItem,
+        sortTracksByAudioChannelMenuItem,
+        sortTracksByOutputChannelMenuItem,
+        sortTracksByInstrumentNameMenuItem,
+        sortTracksByTrackNameMenuItem,
+        sortTracksByUsedMenuItem,
+        sortTracksByCreationDateMenuItem,
         saveAsMenuItem,
         savePanelWindowTitle,
         savePanelPackageRadio,
--- a/Sources/LogicProMCP/Accessibility/AXLogicProElements+Menu.swift
+++ b/Sources/LogicProMCP/Accessibility/AXLogicProElements+Menu.swift
@@ -1,4 +1,4 @@
-import ApplicationServices
+@preconcurrency import ApplicationServices
 import Foundation
 
 
@@ -7,6 +7,22 @@ extension AXLogicProElements {
 
     private static let englishTopLevelMenuTitles: Set<String> = ["File", "Edit", "Track"]
     private static let koreanTopLevelMenuTitles: Set<String> = ["파일", "편집", "트랙"]
+
+    /// The legacy menu helpers intentionally flatten AX failure to `nil` for
+    /// best-effort callers. A mutation verifier needs the stronger distinction:
+    /// -25205/-25212 say a menu node is absent; every other status means that
+    /// node was not read and must refuse before an action is sent.
+    enum MenuItemRead: Error {
+        case found(AXUIElement)
+        case absent
+        case unreadable(stage: String, status: String)
+    }
+
+    enum LogicUILocaleRead {
+        case locale(String)
+        case absent
+        case unreadable(stage: String, status: String)
+    }
 
     /// Get the menu bar for Logic Pro.
     static func getMenuBar(runtime: Runtime = .production) -> AXUIElement? {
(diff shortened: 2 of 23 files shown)


MongLong0214:
## Driven live at `7da95686`. The happy path cannot reach State A, and the reason is an assumption live Logic falsifies

Release binary built from this head, Logic Pro 12.3, Korean UI, a project prepared with three plain
audio tracks and no collapsed stack. Two findings; the first is a defect in this branch.

### 1. `after_order` never moves, so a correct sort is reported as `readback_mismatch`

```
sort_verified criterion=creation_date expected_order=[Yankee, Bravo, Zulu] confirmed=true

  state: "B"   verified: false   reason: "readback_mismatch"   write_attempted: true
  actuated_criterion: "creation_date"   actuated_menu_item_label: "생성일"

  before_tracks: ['Bravo', 'Yankee', 'Zulu']
  after_tracks : ['Yankee', 'Bravo', 'Zulu']      ← the sort WORKED, and matches expected_order
  before_order == after_order:  True              ← the reference list did not move
```

The names moved and the references did not. `trackSortAfterOrder` recovers which reference each row
now holds by matching the after-headers to the before-headers with `CFEqual` on `AXUIElement`
(`AccessibilityChannel+Tracks.swift:390-395`). That assumes Logic reorders the element objects.

It does not. Measured from outside the product, across a sort that demonstrably moved the rows:

```
BEFORE   [0] «1개의 ‘Bravo’ 트랙»   [1] «2개의 ‘Yankee’ 트랙»   [2] «3개의 ‘Zulu’ 트랙»
AFTER    [0] «1개의 ‘Yankee’ 트랙»  [1] «2개의 ‘Bravo’ 트랙»    [2] «3개의 ‘Zulu’ 트랙»

CFEqual map:  after[0] -> before[0]   after[1] -> before[1]   after[2] -> before[2]
```

Element 0 held "Bravo" before and holds "Yankee" after. Logic keeps the same `AXUIElement` objects at
the same rail positions and swaps their content. So the join is always the identity permutation,
`after_order` is always `before_order`, and the comparison against `expected_order` can only fail.
`sort_verified` can never award State A for any criterion.

The honest half is that it stays State B rather than claiming success — the contract held. But the
operation is unusable as written.

A fix is close at hand: this branch already refuses a project with duplicate track names before any
AX work, precisely because `(name, ordinal)` is not an identity. With that precondition already
enforced, names ARE unique for the duration of a run and can carry the join that element identity
cannot.

### 2. `Track > Sort Tracks by > Track Name` performs no operation on Logic 12.3

Not a defect in this branch, but it is the criterion the contract leads with. Measured with Logic's
own undo entry as the witness — a real sort leaves «트랙 정렬 실행 취소», a no-op leaves whatever was
there before:

```
order Yankee, Bravo, Zulu  (not name-sorted)
click 트랙 이름
  undo before: «이름 변경 중 실행 취소»
  undo after : «이름 변경 중 실행 취소»      ← Logic recorded nothing
  order after: Yankee, Bravo, Zulu           ← unchanged
```

Reproduced twice, from two different starting orders, through both `AXPress` and a System Events
menu click. The full submenu on the same three tracks:

| leaf | undo entry created | order moved |
|---|---|---|
| MIDI 채널 | no | no |
| 오디오 채널 | no | no |
| 출력 채널 | no | no |
| 악기 이름 | **yes** | **yes** — Zeta,Bravo,Zulu → Bravo,Zeta,Zulu |
| 트랙 이름 | no | no |
| 사용 여부 | no | no |
| 생성일 | **yes** | **yes** — Bravo,Zeta,Zulu → Zeta,Bravo,Zulu |

All seven leaves read `AXEnabled = true`, so the branch's disabled-item guard cannot catch this. The
three channel criteria plausibly tie on these tracks; `트랙 이름` does not, and still does nothing.

Worth recording: on plain audio tracks `악기 이름` sorted them alphabetically by track name — the
result `트랙 이름` was asked for and did not deliver.

### What this run DID confirm

- Every `AXLocalePolicy` label added by this branch is exactly right. Read live from the menu:
  `트랙을 다음으로 정렬`, and the leaves `MIDI 채널`, `오디오 채널`, `출력 채널`, `악기 이름`,
  `트랙 이름`, `사용 여부`, `생성일`.
- `actuated_menu_item_label` / `actuated_criterion` report the leaf that was actually located, and
  matched the request on every call.
- The collapsed-stack refusal is not hypothetical. Logic's own default new-project template ships an
  `Absolute Zero` track stack, collapsed, so a freshly created project trips that refusal until the
  stack row is deleted. Two of two new projects behaved this way.

### Blocking

Finding 1 blocks merge. Finding 2 is Logic's behaviour and belongs in the contract's documentation
and, probably, in a refusal for a criterion measured to be inert.

The earlier note that #748 blocked this drive was wrong — `trk_` references are obtainable, and this
run used them.

MongLong0214:
Addendum, measured so whoever fixes finding 1 does not have to search for a join key.

A track header exposes no stable identity. Its full attribute set on Logic 12.3:

```
AXDescription, AXParent, AXSize, AXFocused, AXChildren, AXRole, AXTopLevelUIElement,
AXHelp, AXPosition, AXWindow, AXRoleDescription, AXSelected, AXFrame
```

No `AXIdentifier`, no `AXValue`, no per-track handle of any kind. `AXHelp` is the same generic
sentence on every row. The only content that distinguishes one row from another is `AXDescription`,
which on a Korean Logic reads `«N개의 ‘이름’ 트랙»` — and `N` is the track NUMBER, so it moves with
the sort and cannot be the key either.

That leaves the name, which this branch already guarantees is unique: it refuses a project with
duplicate track names before any AX work, for its own reasons. So the precondition the fix needs is
already enforced, and the join `trackSortAfterOrder` cannot do by element identity it can do by name.

Stated as a limit rather than left implied: this makes the sort verification depend on names being
unique for the duration of one call, which is exactly what the existing refusal buys and no more. A
rename racing the sort would defeat it, the same way it would defeat any name-keyed identity.

MongLong0214:
One more thing about finding 1, and it explains how it survived review.

`trackSortAfterOrder` is `private` and appears nowhere under `Tests/`. Grepping the branch returns
exactly two hits, both in `AccessibilityChannel+Tracks.swift` — its definition at `:378` and its one
call at `:215`.

So the function that decides the verdict has **no unit coverage at all**, and could not have any as
written: it takes `[AXUIElement]`, which the deterministic suite never produces.
`Issue448TrackSortVerifiedTests.swift` exercises the contract and dispatcher layers above it, and
every one of them passes while the join underneath returns the identity permutation.

That is why this needed a live drive to find, and it is the part of the fix worth getting right:
changing the join from element identity to track name does not just correct the behaviour, it moves
the decision onto values a test can construct. The seam comes free with the fix.

Suggested coverage once it is testable, both directions:

- an after-list that is a permutation of the before-list, built from the SAME element objects with
  only the names moved — the shape live Logic actually produces — yields the permuted references;
- two rows sharing a name yields no order at all, rather than an arbitrary one.

The second matters because name uniqueness is a precondition this operation enforces elsewhere, and
a join that silently picks one of two identical keys would quietly re-introduce the identity problem
in a different disguise.

MongLong0214:
## The happy path reaches State A. Driven live, five times, and one of them failed — both are below

Fix is on `fix/448-name-join` (`29d89565`), built to a release binary and driven against Logic Pro
12.3, Korean UI. A three-track project whose creation order (`Zulu, Alpha, Bravo`) disagrees with its
rail order, so `creation_date` has to move something.

```
run 1: state=A  moved  Alpha,Bravo,Zulu -> Zulu,Alpha,Bravo   before_order==after_order: False
run 2: state=A  moved  Alpha,Bravo,Zulu -> Zulu,Alpha,Bravo   before_order==after_order: False
run 3: state=A  moved  Alpha,Bravo,Zulu -> Zulu,Alpha,Bravo   before_order==after_order: False
run 4: state=A  moved  Alpha,Bravo,Zulu -> Zulu,Alpha,Bravo   before_order==after_order: False
```

`before_order != after_order` is the whole point: the reference list now follows the tracks. Before
the fix that comparison was `True` on every sort, which is why State A was unreachable.

### The fifth drive, which returned State B on a correct sort

The first attempt after the fix produced this:

```
state: "B"   reason: "readback_mismatch"
before_tracks: ['Alpha', 'Bravo', 'Zulu']
after_tracks : ['Zulu', 'Alpha', 'Bravo']      ← the sort landed, and matches expected_order
before_order == after_order: True              ← but the settled reference list is the pre-sort one
```

I ran it while the machine was under a full ship-gate suite. Read against the settle loop, that is
what you would expect:

```swift
for attempt in 0..<4 {
    if attempt > 0 { usleep(75_000) }
    ...
    if previousOrder == resolvedAfterReferences { return .read(resolvedAfterReferences) }
```

The first read is taken with no delay, and two equal consecutive reads are accepted as settled. If
the sort has not landed within ~75 ms, reads 0 and 1 both return the PRE-sort order and the loop
reports that as settled. `afterTracks` is reassigned on the later iteration, which is why the receipt
shows sorted names beside an unsorted reference list — the two came from different reads.

This is not new to the fix, and the comment above that loop already says so: *"remains an unmeasured
assumption until a live sort establishes Logic's settling distribution."* It is now measured: one in
five, under load. **Two equal consecutive reads cannot distinguish "settled" from "not started yet"
when the first read is immediate.**

I am not changing it in this pull request. Widening a blocker fix into the settle policy is how a fix
grows a second, unreviewed change, and the loop's failure is conservative — it under-reports a real
sort as State B rather than awarding State A to something unproven.

### Still true from the earlier drive, and not fixed here

`Track > Sort Tracks by > Track Name` performs no operation on this Logic build — no undo entry, no
reorder, from two starting orders and through two actuation paths — while `악기 이름` and `생성일`
both work. All seven leaves read `AXEnabled = true`.

Worth noting for whoever touches the enabled-state guard: this repository already recorded, in a
CommitLore warning on `ResourceHandlers+StateReaders.swift`, that *"menuItemEnabledForActuation is
strict on purpose — an unreadable AXEnabled must not authorise a pick."* Today's measurement is the
other half of the same lesson: a READABLE `AXEnabled = true` does not authorise one either.

### What is in the fix

`trackSortAfterOrder` joins by track name instead of `AXUIElement` identity, and is now `internal`
taking `[String]` rather than `private` taking `[AXUIElement]` — which is what makes it testable at
all. Six new unit tests; injecting the old identity join reddens three of them, including the one
named for this regression. The existing 34 `Issue448` tests still pass.

PR #752, opened 2026-09-03
fix(#749): hoist two string concatenations out of argument literals, and guard the shape

### LINKED ISSUE #749: main can fail to build: a long chained-+ string literal inside a merging(...) trailing closure times out the type-checker
Reported by @ElDavichin while testing #738 (see https://github.com/MongLong0214/logic-pro-mcp/issues/736#issuecomment-5514082367): `main` does not compile on their machine.

> `AccessibilityChannel+MIDIImport.swift:640` timed out the type-checker on a long chained-`+` string literal inside a dictionary-literal + trailing-closure expression (`extras.merging(["hint": "..." + "..." + "\(x)" + ...]) { _, new in new }`). Extracted the hint into a separate `let` before the call — mechanical fix, no behavior change, but it blocked the build entirely until patched locally.

## Why it is toolchain-dependent

The expression is a dictionary literal, containing a five-way `+` chain mixing literals and interpolation, passed as an argument to `merging(_:uniquingKeysWith:)` with a trailing closure. Swift type-checks that as one expression; the search space grows quickly with each `+`, and whether it finishes inside the limit depends on the compiler version. It builds here and does not build there, which is the worst shape for a contributor pulling cold: the failure is a timeout, not a diagnosable error in their code.

## Where

Two sites in `Sources/` have a `merging([...]) { _, new in new }` whose dictionary literal contains three or more `+` concatenations:

```
Sources/LogicProMCP/Channels/AccessibilityChannel+MIDIImport.swift:637   5 concatenations   <- the reported one
Sources/LogicProMCP/Saga/SagaJournal.swift:793                          3 concatenations
```

## Fix

Hoist each message into a `let` before the call. That is behaviour-preserving by construction: the same string is built, once, and the dictionary literal then holds a single identifier.

Worth adding a repository guard for the shape, so it cannot come back — the same message text is easy to re-inline while editing, and nobody on a toolchain that happens to finish will notice.

MongLong0214:
A contributor could not build `main` at all. Reported by @ElDavichin while testing the merged #738 fix:

> `AccessibilityChannel+MIDIImport.swift:640` timed out the type-checker on a long chained-`+` string literal inside a dictionary-literal + trailing-closure expression. Extracted the hint into a separate `let` before the call — mechanical fix, no behavior change, but it blocked the build entirely until patched locally.

## Why it is toolchain-dependent

Swift type-checks one expression at a time. A `+` chain of literals mixed with interpolation, inside a collection literal, passed to a generic call with a trailing closure, is all one expression, and whether the solver finishes inside its limit depends on the compiler version. It built here and did not build there. A timeout is not a diagnosable error in the reader's own code, so the whole cost lands on whoever pulls the repository cold.

## What changed

Two sites had the shape — the reported one with five concatenations, and the saga wedge-timeout hint with three. Both now bind the string to a `let` before the call, which is behaviour-preserving by construction: the same string is built once and the literal holds an identifier.

`Scripts/check-typechecker-heavy-literals.py` refuses the shape from now on, at a threshold of three rather than the five that actually broke — a guard that only refuses what already broke cannot prevent the next one. It is discovered by `run-repo-guards.py` like every other guard (16 of 16 pass), and it was verified both ways: it fails when the reported concatenation is put back, and passes when it is removed.

Concatenations outside a call argument are deliberately left alone: a `let` holding a long chain is exactly the shape this asks people to write.

Full suite: 4,343 tests.

Closes #749


MongLong0214:
## Not merging yet — an adversarial review found three real defects, and I reproduced two of them by running the guard

CI is green on every job. That is not enough here: the thing this PR adds is a guard, and a green
tree proves only that the guard is silent, never that it is right.

### 1. The guard blocks ordinary code — concatenations are summed across the whole literal

`offenders()` counts `+` over the entire matched collection body, so unrelated entries add up.
Run against the guard as it stands on this branch:

```swift
report(["a": "one" + "!", "b": "two" + "!", "c": "three" + "!"])
```

```
1 argument literal(s) carry 3+ string concatenations:
  Sources/X/Ordinary.swift:3  concatenations=3
exit 1
```

Three separate entries, one concatenation each, no long chain anywhere. The threshold is supposed to
mean "a single value built from three or more concatenations". Counting per literal makes it mean
something else, and what it blocks is exactly the ordinary code #749 exists to keep unblocked.

### 2. Comments are scanned as code

```swift
    // Example only, not compiled:
    //   report(["hint": "a" + "b" + "c" + "d"])
    report(["hint": staticHint])
```

Same output: `concatenations=3`, exit 1. Writing down the shape you are warning about is enough to
trip the warning.

### 3. The guard has no fixture test, against this repository's own convention

`Scripts/test-menu-literal-guard.sh` and `Scripts/test-roadmap-table-guard.sh` both exist. There is
no `Scripts/test-*typecheck*`. So nothing shows this guard failing on input that should fail — the
"46 of 46 all green" shape that is not evidence.

### 4. The roadmap row overstates what is covered

The review also found the detector misses labeled or non-first arguments, a later chain after a
nested call, parenthesised pieces, and raw strings. Those gaps do not have to close in this PR. The
row does have to say what the guard covers and name what it does not.

### What the review could NOT refute, recorded because it is worth as much as the findings

- **The two Swift hoists are byte-preserving.** Every fragment, space, punctuation mark, Unicode
  character and the MIDI interpolation is unchanged and in the same order; the interpolation reads a
  static constant, so only the point of evaluation moved and there are no side effects.
- **The guard is reachable from CI.** `Scripts/run-repo-guards.py` discovers top-level
  `Scripts/check-*.py`, the new filename is discovered, and both `compile` and `test` invoke that
  runner while the required `build` job depends on both.

### One finding outside this PR's scope, noted so it is not lost

`build.needs` lists only `compile` and `test`, not `roadmap` (`.github/workflows/ci.yml:303`). The
roadmap drift check runs, but it does not gate the required context, so the update rule does not
actually block a merge. That is a workflow question, not this branch's.

Fixes for 1-4 are in progress on this branch.

MongLong0214:
## Fixed, and checked against the same inputs that caught it — not against the fix's own tests

The guard was rewritten to count per VALUE rather than per literal, and to strip Swift comments with
a real lexer that understands raw strings (`#"…"#`), multiline `"""…"""` and nested block comments.

I re-ran my two reproductions against the new guard, outside its own fixtures:

```
three unrelated entries, one concat each   accepted (exit 0)     was: REFUSED
commented-out example                      accepted (exit 0)     was: REFUSED
the real merging([...]) shape              REFUSED (exit 1)      concatenations=4
```

The fixture test is new and asserts BOTH directions — `Scripts/test-typechecker-heavy-literal-guard.sh`
returns exit 0 with `ok` on the offending shape being refused and on four ordinary shapes being
accepted. It is wired into `compile` and `test`, the two jobs `build` requires, so it gates a merge
rather than merely reporting.

## Coverage is now stated rather than implied

The review's false negatives were measured again against the new guard:

```
labeled / non-first argument                   still accepted
chain after a nested call in the same literal  still accepted
parenthesised pieces                           still accepted
raw strings                                    still accepted
line-split chain with a comment between        now REFUSED
```

Four of those five gaps remain, and the roadmap row now names each one instead of claiming the shape
is refused. That is the right trade: the guard covers the shape that actually broke a contributor's
build, and says what it does not cover, rather than growing a regex until it starts blocking ordinary
code again.

## Remaining before merge

This branch is behind `main` and its table is missing the #754/#755 rows, so its roadmap job would be
red as it stands. It rebases once #756 lands. Ship gate on the current head: PASS.

Issue #304, opened 2026-07-10
ADR-016: Smart Tempo and tempo-map automation — backlog; preserve existing tempo fixes


MongLong0214:
# ADR-016 — Tempo-map inspection and later Smart Tempo automation

> **SSOT revision: 2026-09-21. OPEN / BACKLOG.** General tempo control exists; full Smart Tempo/mode/map/analysis automation does not. Activate a specific stage through #308. Existing shipped tempo defects remain maintenance work and are not deferred just because this expansion is backlogged.
>
> Audit anchor: `d067e91958ad3ff1a1bdfb1e6e39a3e23e786517`. Historical experiments are evidence for their own conditions; no new live run is claimed here.

## Decision

Separate four domains: project tempo mode, region Flex & Follow, Smart Tempo analysis, and the Tempo List/map. The first new deliverable is a **truthful read-only Tempo List snapshot**. Do not start with a project-wide automatic adaptation engine, beat-analysis framework, multi-track workflow or generic async job platform.

`set_tempo` is not Smart Tempo support. The original single-tempo correctness issue #278 is separate from this expansion. Existing named transport operations and their safety fixes must keep working.

## History and corrections

- Early probes inspected the Smart Tempo editor and graphical tempo track and did not find event data. Those observations do not rule out the numeric Tempo List, which is the relevant map surface.
- Later records show numeric tempo events and a sub-bar placement experiment: a probe created a row at `5.3.2.60`, read the same position from the list and verified its removal by rereading after Undo. The row inherited the current BPM. This is reachability evidence, **not a public create/map-write feature**.
- A false verified tempo result was reproduced when the control-bar field held newly typed text but Logic refused the edit with a modal on a multi-event project. Subsequent fixes narrowed verification to complete, clean modal observations and truthful write/readback outcomes.
- [The final issue correction](https://github.com/MongLong0214/logic-pro-mcp/issues/304#issuecomment-5655196074) withdraws the claim that Undo requires the Arrange window: a stuck modal was the relevant condition. Do not bake the retracted window claim into a preflight rule.
- Deleting a tempo event by routes other than Undo was not established in those experiments. Failures observed under a blocking modal are not universal evidence that every delete route is impossible.
- The later roadmap review records additional after-write uncertainty and modal-timing limitations. A count of other handlers with similar shape is not a verified list of defects in all those handlers.

## Current source and observation entry points

[...]

PR #753, opened 2026-09-03
feat(#304): read the tempo map, and write it by converging

MongLong0214:
The Tempo List surface, measured in both directions, plus the fixes an adversarial review asked for. `TempoMapAX` is an internal reader and a converging writer; it is wired to no dispatcher or registry entry, and its loop has not been driven live.

## What was measured

The roadmap row said nothing had ever been measured against this surface. It has now.

**Reading is solid.** `편집 › 템포 › 템포 목록 보기` opens a fully classified table — zero elements without a role. Each row is three cells and the whole value of each sits on the cell's group:

```
AXCell → AXGroup AXDescription=[1 1 1 1 ]         position (bar beat div tick)
AXCell → AXGroup AXDescription=[120.0000]         tempo
AXCell → AXGroup AXDescription=[01:00:00:00.00]   SMPTE
```

So a tempo event reads back exactly, as a number, with no per-digit parsing. The event count from the `항목 수` text is checked against the row count; a disagreement is an incomplete read, not a shorter list.

**Writing works, but a single set does not land the target.**

```
118.0000 → set 120 → 119.0000    changed, did NOT land
119.0000 → set 120 → 120.0000    changed, landed
120.0000 → set 120 → 120.0000    no change, already there
```

One unit per write toward the target. So a verified tempo write is a bounded converging loop with an exact readback, not a set.

**A caching mistake worth recording.** The first probe held an `AXUIElement` across the write, read `nil` afterwards, and reported the value had changed — the row re-renders and invalidates the reference. Reading nil from a stale reference is not an observation of change. Every read re-finds its element from the window, and a test pins that by making a cached reference go bad.

## What the review changed

- **Write authority rested on reads that could vanish.** The table census, row and cell reads, group roles, descriptions, the count witness and window discovery were all best-effort, so `cannotComplete`, `invalidUIElement`, `failure` and `apiDisabled` collapsed into the same nothing as `noValue` and `attributeUnsupported`, and an unreadable competing table could disappear. Twelve sites checked, twelve changed: only the two absence codes mean absent, and an invalidated element is re-read during settlement rather than treated as a verdict.
- **A valid extreme input crashed.** `Double(Int.max)` rounds to 2^63, outside `Int`, so that target passed the guard and trapped the conversion; slightly smaller ones authorized billions of iterations. The writer now takes only integer BPM inside `TransportDispatcher.supportedTempoRange` (5–999) and carries a write cap and a wall-clock deadline. Fractional targets refuse: the measurement is one unit per write at integer tempi, and nothing has measured whether a fractional target can land at all.
- **There was no settlement protocol.** One read of a stale value became "did not move" and one transient rebuilding error became terminal, with rollback starting against the same rebuilding tree. Reads settle before they are judged, and the fixture delays its render so a test can fail the way Logic fails.
- The two count witnesses must agree twice before the reading is accepted; a crossing of the target is refused rather than counted as progress; a rollback that drifts records its write count and the tempo it finally observed.

## Scope

Out of scope and refused rather than attempted: the six disabled tempo-menu items (they need an audio region carrying tempo information), the Smart Tempo editor (disabled without an audio selection), and creating or deleting tempo events. Rollback uses the same converging loop, because the stepping behaviour means a careless restore drifts.

The ADR and roadmap describe this as what it is — an internal, unwired reader plus a converging writer that is not live-qualified — rather than as a callable capability.

Focused suites: 182 tests. Full suite and repository guards pass.

Refs #304


MongLong0214:
## CI is red again on the same test, and the previous fix addressed the wrong half

```
✘ the practical write cap stops a long valid walk and reports the rollback receipt
  TempoMapAXTests.swift:336  Expectation failed:
    refusedOutcome.rollback
      → .failed(failure: Tempo write reached its wall-clock deadline at 7.0, not 5.0.,
                writes: 62, finalObserved: Optional(8.0))
    == .restored(writes: 64, finalObserved: 5.0)
  TempoMapAXTests.swift:337  fixture.currentTempo → 7.0  ==  5.0
```

Read carefully, the numbers say which half broke. The FORWARD walk behaved: the first expectation,
`failure == .attemptBudgetExhausted(observed: 69, target: 999, budget: 64)`, is not among the
failures, so `convergenceDeadlineOverride: 600` did its job and the budget stopped the walk.

What timed out is the **rollback**. `deadlineExceeded` renders as "at 7.0, not 5.0" where those are
tempo values, not seconds: the restore was walking 69 back down to 5, reached 7.0, had spent 62
writes, and hit a wall clock.

The cause is that the override never reaches that path. `setExistingTempo` applies it once, at

    deadline: Date().addingTimeInterval(convergenceDeadlineOverride ?? convergenceDeadline)

and then calls `rollbackToInitialTempo(at:initial:in:localeIdentifier:runtime:fallbackObserved:)`,
whose signature has no deadline parameter at all, so the restore converges against the production
3 s default. On this machine 64 writes fit inside it; on the runner they do not.

So the earlier change pinned one bound of two and left the third clock unpinned. The test's own
comment — "the deadline is pushed far out so the budget is the only thing that can stop this walk" —
describes an intent the code only half implements.

Fix: thread `convergenceDeadlineOverride` through `rollbackToInitialTempo` into its converge call, so
a test that says it is measuring the write cap is measuring the write cap in both directions.
`wallClockDeadlineBoundsLongWalk` keeps pinning the other bound and should be unaffected.

Worth noting for whoever takes it: this is the second CI-only failure on this one test, and both
times the local gate was green. The bound that fires is decided by how fast the host is, which is
exactly the property a test must not have.

MongLong0214:
CI went red on a test this branch does not touch, and I re-ran rather than changed anything. Recording why, so the retry is not a blind one.

```
✘ switchThatNeverChangesStructureRefusesWithinDeadlineAndLeavesEntryView()
  ControlsViewBooleanParameterWriterTests.swift:993
  Expectation failed: completedWithinBoundedRestore
  ...failed after 0.791 seconds
```

That is `#299`'s Controls-view test, not `#304`'s. Line 990 asserts the whole operation finishes
inside a **0.75 s wall-clock budget**, and the runner took 0.791 s. There is no causal path from a
tempo-map change to a Controls-view switch, and `main` is green on the same test, so this is the
budget being marginal on a loaded runner rather than a regression here.

It is the same class of defect as the one this branch just fixed — a test whose verdict depends on
how fast the host is. The difference is that this one is not mine to fix inside this pull request:
changing a neighbouring suite's timing assumption while landing a blocker is how a fix grows a
second, unreviewed change.

Re-run of the failed jobs is in flight. If it goes red again on the same assertion it is
deterministic on that runner, not marginal, and then it needs a real fix on its own branch — the
budget replaced by something that does not measure the machine.

Issue #292, opened 2026-07-10
ADR-009: Stock plug-in capability contract — complete current verified-parameter scope


MongLong0214:
# ADR-009 — Verified stock plug-in capability contract

> **SSOT revision: 2026-09-21. OPEN; current-surface completion, not a new plug-in platform.** This body is the executable brief. #308 owns queue order. The prior 'no remaining surface can reach ten parameters' conclusion is superseded by the later EQ measurements.
>
> Source baseline: `d067e91958ad3ff1a1bdfb1e6e39a3e23e786517`. Recorded live results below belong to their own fixtures/heads; this editorial reconciliation is not a fresh qualification.

## Goal and scope decision

Make the existing stock-plug-in capability data, public claims and actual verified writers agree. Reuse `StockPluginCatalog`, the current operation registry and existing AX read/write paths. Do not create a competing `VerifiedParameterCapabilityRegistry` merely because that name appeared in the proposal.

The current completion slice covers capability/readback accuracy, the supported parameter set, target/window behavior, unsupported/drift outcomes and current evidence. **Batch snapshot application, partial-batch compensation and automatic-EQ application are explicitly deferred to #955.** They are not silently marked done and do not block the bounded single-operation slice.

The existing `logic://stock-plugins` resource remains the capability-list authority. The proposal's additional list alias is not required. A new public generic parameter-read/snapshot API belongs to the scoped #955 follow-up when needed; existing readback must already be correct here. No new public API is implied by editing this issue.

## Current implementation history

| Evidence | Meaning and limit |
|---|---|
| Existing verified insertion/acquisition and `set_param_verified` | Foundation already present; do not rebuild insertion or the target registry. |
| PR #725 / #301 | Added named Channel EQ band-parameter writes through the AX plane. Controls-view-only conclusions cannot describe this separate path. |
| [2026-09-12 measurement](https://github.com/MongLong0214/logic-pro-mcp/issues/292#issuecomment-5644404963) | Eleven named parameters reached different displayed values; also corrected a stale executable-image mistake and unit formatting. |
| [2026-09-13 Q investigation](https://github.com/MongLong0214/logic-pro-mcp/issues/292#issuecomment-5650371845) | Raw Q movement and engineering-unit movement were different questions. Do not infer all units from one successful raw-value path. |
[...]

Issue #757, opened 2026-09-03
The sort's settle witness accepts two immediate equal reads, so a slow sort reports its own pre-sort order as settled


MongLong0214:
Measured on `fix/448-name-join` while driving `tracks.sort_verified` live (Logic Pro 12.3, Korean UI).
Split out of #746 so it is tracked after that branch merges.

## The witness cannot tell "settled" from "not started yet"

`AccessibilityChannel+Tracks.swift`, the `after:` closure of the sort transaction:

```swift
for attempt in 0..<4 {
    if attempt > 0 { usleep(75_000) }
    guard case .read(let afterRead) = strictTrackSortStateRead(runtime: runtime) else { return .unavailable }
    afterTracks = afterRead.tracks
    guard let resolvedAfterReferences = trackSortAfterOrder(...) else { return .unavailable }
    afterReferences = resolvedAfterReferences
    if previousOrder == resolvedAfterReferences { return .read(resolvedAfterReferences) }
    previousOrder = resolvedAfterReferences
}
```

The first read is taken with **no delay**, and two equal consecutive reads are accepted as settled. If
the menu action has not landed within ~75 ms, reads 0 and 1 both return the PRE-sort order and the
loop reports that as the settled answer.

## Measured, not inferred

Five live drives of a correct `creation_date` sort on a three-track project:

```
4 of 5   state=A   before_order != after_order   (idle machine)
1 of 5   state=B   reason readback_mismatch      (machine under a full ship-gate suite)
```

The failing receipt is self-contradictory in a way that identifies the cause exactly:

```
before_tracks: ['Alpha', 'Bravo', 'Zulu']
after_tracks : ['Zulu', 'Alpha', 'Bravo']     ← the sort landed and matches expected_order
before_order == after_order: True             ← the settled reference list is the pre-sort one
```

`afterTracks` is reassigned on a later iteration than the one whose references were accepted, so the
two halves of the receipt came from different reads.

The code already anticipated this. The comment above the loop says the bound *"remains an unmeasured
assumption until a live sort establishes Logic's settling distribution."* This is that measurement:
one in five under load.

## Why it is not urgent, and why it should still be fixed

The failure is conservative — it under-reports a real sort as State B rather than awarding State A to
something unproven, so the Honest Contract is not violated. What it costs is a correct operation
being unusable on a loaded machine, with a receipt that reads as contradictory.

## Not a design, just the constraint any fix has to meet

[...]

PR #765, opened 2026-09-04
feat: a measured map of Logic's UI, and three ADR blockers re-measured

MongLong0214:
Read every open ADR row against live Logic instead of against itself. **Five of the six examined were wrong about their own blocker**, and the pattern is always the same: a row describing a wall that nobody re-measured after the code moved. So this PR does two things — corrects the rows, and adds the machinery that makes the next drift visible instead of silent.

## The rows, and what the readings say

| ADR | the row said | measured |
|---|---|---|
| **ADR-008** #291 | node identity needs a decision | the decision shipped in #740; the live wall is reading the **destination** |
| **ADR-009** #292 | apply-back waits on a locator | the locator exists and is wired; what is missing is any second route to the ADR's own criterion — sliders report success and do not move (2026-09-02), popups open and expose only the current choice or several identically-titled options (2026-09-04) |
| **ADR-010** #293 | the public provider stays closed | both notes round-trip correctly. The closed half is sharper: `collect` needs a `RegistryResolvedIdentityProof` whose only construction sits inside `#if QUALIFICATION_FAULT_SEAM`, scoped to debug — the shipped binary exports no mint, and nothing registers the provider either |
| **ADR-011** #299 | "no write attempted anywhere" | writes were attempted; slider and popup are both walls and the checkbox path is the one that works |
| **ADR-014/015/017/018** | various | corrected, each with the reading |
| **ADR-001** #284 | the `R-SEM` split covers 109 operations | the registry declares 113, each with exactly one spec row and each in the allow table. Read-only is still exactly 23, so the four additions are all mutating |

## The machinery: `docs/observations/`

13 records, one per measurement, each a claim bound to a host build with the method to re-run it. Two guards run in CI — `run-repo-guards.py` globs `Scripts/check-*.py`, so they are enforced rather than optional:

- **`check-observation-records.py`** — schema, and *every number in a conclusion must appear in a reading*. It rejected my own records three times: a poll interval, a PR number, and a count of dependent sites, each asserted in prose with nothing behind it.
- **`check-observations-cover-live-walls.py`** — a roadmap row claiming a dated measurement needs a record for **that row's own issue**.

Both were hardened by finding holes in them:

- The coverage trigger matched too few wordings. Five escaped it (`live-verified 2026-09-04`, `observed live on`, `drove this live`, …), and widening it exposed two genuinely unrecorded claims.
- The coverage rule read every issue number on a line. ADR-011's row claims about #299 and mentions "behind #292", so a #292 record was being accepted as coverage. Attribution now comes from the row's identity cells.
- **`depends` validated the file and not the symbol.** Renaming `LocatorFailure.popupUnmeasured` left the path resolving, the build passing, and the record pointing at a name the file no longer contained. Every component of a dotted symbol is checked now.
- **The `host` block was typed, not measured.** All nine records claimed `macOS 26.6` on a machine that has run `macOS 26.3` since February — the first block was written by hand and every later record inherited it by copy, so the error spread exactly as fast as the records did. `Scripts/observation_host.py` generates it; `host.os` is now required and is a drift axis.

Each guard was shown to fail on the defect it exists for and to pass once it was gone.

## One behaviour change

`set_param_verified` on a Controls-view popup answered *"selection has not been measured, so no menu action is guessed"*. True when it shipped on 2026-09-02, false after 2026-09-04. The refusal is right either way; what was wrong is what it told the caller to expect. `LocatorFailure.popupUnmeasured` is now `popupNotAddressableByName` and its prose states the wall. `PluginSetParamVerifiedTests` pinned the old sentence and would have gone on passing over stale text.

## Three issues opened, each with its reading

- **#766** — `inferTrackType` answers `audio` for every track. Three tracks of three types returned one `observed_track_type`; the Input Monitoring help present on every header names an audio track and a software instrument track in the same sentence, so both label sets match on all 7 headers and no branch order repairs it.
- **#767** — AppleScript `entire contents` returns an empty list *without raising*, for every application on this host (0 against 464 by a manual descent of the same window). Two documented AX-opacity conclusions cite an empty result from it as their evidence; those are marked here rather than quietly left.
- **#768** — 20 of 138 `LabelSet`s match exactly one language, eight of them Korean-only. An empty `variants` list cannot say whether the string is invariant or whether nobody looked.

## Review

No cross-provider review was available for this branch — one provider is at its usage limit until 2026-09-08 and the other has no tokens. Stated rather than papered over: CI and the ship gate are not a second reader. The design decisions with no second opinion are the 19-surface taxonomy, keeping `observations-status.py` out of CI, and the contents of the grandfather list.

`Scripts/observations-status.py --coverage` reports 7 of 19 surfaces measured, 12 untouched.


MongLong0214:
## Independent review could not run, and what I did instead

Both providers are out: `codex` is at its usage limit until 2026-09-08, and `grok` has no tokens. So
this had no second reader, and I am saying that rather than letting the ship gate and CI stand in for
one — they check different things.

What I did instead was aim the adversarial pass at my own work, and it found a real hole.

### I refuted my own coverage guard

Its trigger required `measured` or `driven` within 24 characters of a date. Five wordings from rows I
might actually write walked straight through:

```
live-verified 2026-09-04: …                    caught=0
observed live on 2026-09-04 that …             caught=0
we drove this live and it does not work        caught=0
measured on a live Logic yesterday             caught=0
2026-09-04 measurement: …                      caught=0
```

A pattern that narrow reads as a guard and is a hole — the exact failure this directory exists to
stop, in the thing built to stop it. The trigger is now two independent halves, either sufficient: a
measurement verb near a date in the spellings people write, or `live` beside a measurement word with
or without a date. An undated live claim is a weaker citation, not a weaker claim. All five are
caught now.

### Widening it found two more gaps

- **#300** — ADR-012, closed 2026-08-28, predates the records. Grandfathered with its reason, like
  the others.
- **#757** — measured **today**, and I had not written it down. The settle-witness record now lands
  here: five drives, one under load, with the receipt whose `after_tracks` were sorted beside an
  unsorted `after_order`.

And the schema guard immediately caught that new record citing a `75 ms` poll interval that appeared
in no observation. The reading moved into `observations`, which is where it belongs. A guard that
catches its author on the first record it sees is the outcome I wanted from this.

### What still has no second reader

The design decisions, as opposed to the mechanics: whether `--stale` staying out of CI is right,
whether the 19-surface taxonomy is the correct cut, and whether grandfathering by issue number is a
list that will actually shrink. Those are judgement calls I made alone and they are worth a look when
a provider is available.

Guards: 18 pass. Records: 4, schema-clean. Coverage: 3 of 19 surfaces.

MongLong0214:
## Scope grew: three ADR rows re-measured, not one

Since the PR was opened this branch also went after ADR-014 and ADR-015 the same way. The results
differ from each other, which is the point of doing it rather than reading the rows.

| ADR | what the docs said | measured |
|---|---|---|
| 009 #292 | "a locator for row-labelled controls is new work" | the locator exists and is wired; the real wall is that popups cannot be selected by name |
| 014 #302 | issue header: `kAXColumns` is absent · roadmap: present with 8 | **both wrong** — status 0, **6** columns, and the names are on the `AXHeader` sort buttons |
| 015 #303 | the seam is compiled out of release | **correct** — zero seam symbols in the shipped binary |

### ADR-014 is the useful one

R2 was described as blocked for want of a column map. It has one. The columns carry no identity —
every `AXTitle` and `AXDescription` empty — and the names sit on the header's sort buttons
(`L`, `M`, `위치`, `이름`, `트랙`, `길이`), which is the route `EventListReadbackCollector.sortButtonTitles`
already takes and matches the labels already in `AXLocalePolicy`.

Recorded with the limit that matters: no region was selected, so the table had no rows. This measures
column STRUCTURE and says nothing about cell content or the row-to-field mapping — which is the part
R2 actually needs.

### ADR-015 held

I set out to refute it the way the other two were refuted and it survived.
`IndependentExpectedSeam` sits inside `#if QUALIFICATION_FAULT_SEAM` (lines 59–121), `Package.swift`
scopes that define to `.debug`, and the shipped release binary carries **zero** seam symbols — so
`independentPayload` is always nil there and `verifyRegion` can only answer `incompleteCannotVerify`.

That is recorded as a confirmation rather than left as an untested row. Otherwise "nobody checked"
and "checked and true" look identical in the table.

### Records

6 records, 5 of 19 surfaces. Every one carries its host build, a reverify command, and the code
standing on it, so a Logic update names what it invalidated.

Ship gate: PASS on `0849bea2`. 18 guards pass.

Refs #292 #302 #303 #306

Issue #798, opened 2026-09-06
The documented way to regenerate ui-labels.json reformats all 2858 lines of it


MongLong0214:
`docs/locale/ui-labels.json` says how to regenerate itself, and running that command rewrites every
line of the file without changing a single value.

## Measured, 2026-09-07, at `b3e4d27b`

```
$ python3 Scripts/locale_labels.py --write
$ git diff --numstat docs/locale/ui-labels.json
2858    2858    docs/locale/ui-labels.json
```

The content is identical — `json.load` of both sides compares equal. Only the indentation moves:

```
committed file   1-space indent   74163 bytes
--write output   2-space indent   85339 bytes
```

`Scripts/locale_labels.py:408` and `:419` both pass `indent=2`, and so does
`Scripts/observations/locale-propose.py`'s `--apply`. The file in the tree is indented with one
space and round-trips through `json.dumps(..., indent=1)` byte for byte, so it was written by
something else and every tool the repository documents disagrees with it.

## Why it is not cosmetic

`Scripts/check-locale-labels-json.py` compares `on_disk["labels"]` with `build()["labels"]` — parsed
objects. Formatting is invisible to it, so nothing in CI can see the drift, and it survives
indefinitely.

What it costs is review. A campaign that legitimately adds two provenance blocks, followed by the
regeneration the file itself asks for, produces a 2858-line diff in which those two blocks are not
findable. The instruction and the check disagree about what the file should look like, and the
person in the middle pays for it — measured today, while adding an 8-line field to two labels.

## Shape of the fix

Either direction closes it; they differ only in how much churn lands once.

* Make the writers emit `indent=1`, matching the tree. Two lines in `locale_labels.py`, one in
  `locale-propose.py`, zero churn, and every in-flight branch keeps applying cleanly.
* Or run `--write` once, commit the 2858-line reformat, and let the generator stay authoritative.

Whichever is chosen, the thing that keeps it closed is a check rather than the choice: a test that
reads the file on disk, feeds it back through the writer, and requires the bytes to be identical.
Comparing parsed objects is what let this drift in the first place, so a test that compares parsed
objects would not detect it — it has to compare bytes.

MongLong0214:
Merged to `main` in #800 — one renderer for ui-labels.json, and a test that compares its bytes.

The roadmap row is updated with what the fix actually established, including what it does NOT close. Closing here rather than through a `Closes` line because the work already landed; this records the state rather than causing it.

Issue #797, opened 2026-09-06
A locked screen is invisible to the live harness: two AX paths disagree and the failure names a depth limit


MongLong0214:
Every live harness reads Logic through the accessibility API. With the screen locked that API does
not fail — it answers differently on each of the three paths the repository uses, and none of the
answers says *the screen is locked*. The run then fails, or hangs, naming something else.

## Measured, 2026-09-07, screen locked, Logic Pro running as pid 3574

```
CGSessionCopyCurrentDictionary()['CGSSessionScreenIsLocked']   True
                                ['kCGSSessionOnConsoleKey']    True

direct AXUIElement, AXWindows of pid 3574                      1   titled 'Logic Pro'
System Events, `count of windows` of process "Logic Pro"       0
CGWindowList, on-screen, owner Logic Pro                       1   name '', 260x282 at (830,222)

Scripts/livekit/evidence.py  logic_window()                    None
Scripts/livekit/evidence.py  _production_ax_modal_signals()    raises _ModalReadError(
                                 'AX sheet descendant search reached depth 32')
```

Three readings of the same application at the same moment: one window, no windows, and one
untitled window. `logic_window()` uses `CGWindowList` deliberately — so the harness can see windows
even when the AX tree is what is under test — and it still returns nothing, because the one window
on screen carries no name to match.

## Why this costs a session rather than a run

The precondition does not report the condition it hit. `_ModalReadError('AX sheet descendant search
reached depth 32')` names a recursion guard, so the reading is *"a Logic window has a pathological
descendant tree"*. That is a plausible product defect, and it is not what happened.

Read through `System Events` instead — which is what an ad-hoc diagnosis reaches for — the answer is
`0 windows`, i.e. *"Logic has no windows"*. Also a plausible product defect. On 2026-09-06 this
reading was taken at face value, Logic was force-restarted on the strength of it, and the restart
changed nothing because the application was never the problem.

## What is not claimed

Whether a locked screen is what wedged a ship gate for 34 minutes on 2026-09-06 with no output past
`full suite`. The gate was killed rather than diagnosed, so the cause of that hang is not
established and it should not be folded into this issue as though it were.

Whether the AX window count under a lock is stable. One window titled `Logic Pro` was seen once;
whether that is the lock's doing or whatever Logic happened to have open is not measured.

## Shape of the fix

`evidence.py` already has the right vocabulary for this: `is_clean` refuses on a dimension whose
[...]

Issue #793, opened 2026-09-05
The campaign proposer silently resurrects a retracted provenance claim


MongLong0214:
A provenance block retracted by hand comes straight back the next time
`Scripts/observations/locale-propose.py --apply` runs against the same census, because nothing in
the ledger records *why* it was retracted in a form the proposer can read.

## What happened, 2026-09-06

`markerListDeleteMenuItem` means the Delete button in Logic's **Marker List window**. It had been
backed by a provenance block citing the ja-JP arrange-menus census:

```
'削除': record=2026-09-05-ja-JP-arrange-menus-census  role=AXMenuItem  attribute=title  observed='削除'
```

That census walks the **application menu bar**, so the `削除` it saw is the Edit menu's Delete —
a different element with the same role and the same string. The claim was retracted for exactly
that reason, and the ratchets carry a dated entry saying so:

> `roles` cannot see the difference — it names a kind of element and Logic puts the same kind in
> many containers.

Running the proposer to attach provenance for two *unrelated* labels re-added the retracted block
verbatim and flipped `coverage['ja-JP']` back to `measured`. Nothing failed. The ratchet even
reported it as an improvement — `1 item(s) closed` — and invited me to tighten the ceiling around
a claim that had been withdrawn as false. It was caught only because the ratchet output named a
label I recognised.

## Why the existing machinery does not stop it

`path_contains` exists and does the right thing, but it is a field **inside a provenance block**:
it constrains evidence that has already been written. The proposer never reads it, and a retracted
label has no block for it to live in. So the constraint that would prevent the bad match can only
be expressed after the bad match has been made.

`roles` is the only label-level filter, and it is too coarse by construction — an `AXMenuItem`
named `削除` is an `AXMenuItem` named `削除` wherever Logic puts it.

## Shape of the fix

A label-level declaration of where its evidence may come from — a surface, a path fragment, or
both — read by the proposer when it matches and by the guard when it validates. Then a retraction
is expressible as a fact about the label ("this one lives in the Marker List window") rather than
as an edit that the next tool run undoes.

The two labels retracted on 2026-09-05 are the test cases: `markerListDeleteMenuItem` must not
match a row whose path is under `AXMenuBar/`, and `pluginWindowEditorViewMenuItem` must not match
the application's View menu.

## Not claimed

Whether any *other* retraction has already been silently resurrected. I checked these two because
the ratchet named one of them; the ledger has 264 variants and I did not audit the rest.

[...]

PR #799, opened 2026-09-06
fix(#793): a retraction is a fact about the label, not an edit the next run undoes

MongLong0214:
Adds `evidence_scope` — a label-level declaration of where a locale label's evidence may come from —
and makes the proposer and both halves of the guard read it.

### The failure this closes

`markerListDeleteMenuItem` means the Delete in Logic's **Marker List window**. The arrange-menus
census walks the **application menu bar**, where `Edit > Delete` carries the same string on the same
`AXMenuItem` role in all three locales, so the claim backed by it was withdrawn on 2026-09-05.

On 2026-09-06, running `locale-propose.py --apply` to attach evidence for two *unrelated* labels put
the retracted block back verbatim and flipped `coverage['ja-JP']` to `measured`. Nothing failed. The
ratchet reported it as `1 item(s) closed`.

`path_contains` would have refused the match, but it lives **inside** a provenance block — it can
only constrain evidence that already exists, and a retracted label has no block to hold it. `roles`
is the only label-level filter and cannot help: an `AXMenuItem` named `削除` is one wherever Logic
puts it.

### What is added

`evidence_scope` carries `surfaces` (checked against `docs/observations/SURFACES.md`),
`path_contains`, and a `reason` that is **required and read** — a scope with no reason cannot be
told from a guess, and this is how a retraction gets written down.

Three sites honour it:

* the proposer skips census rows outside the scope *before* comparing any string;
* `provenance_problems` refuses a citation whose record was measured on another surface, or whose
  row sits outside the path;
* `coverage_problems` does the same — of the four wrong-element claims measured on 2026-09-05, two
  were coverage, so enforcing one half only would have left the other able to repeat it.

A block's own `path_contains` is **ANDed** with the label's rather than substituted for it, so the
hole does not reappear one level down. `locale_labels.py` carries the field through `--write`: a
retraction a regeneration erases is one the next campaign silently undoes.

### Measured

Both labels retracted on 2026-09-05 now declare a scope. Against the three navigation-free censuses
they are refused in every locale that matched them, and **every other label's proposals are byte-
identical**, so the scope is the only thing that moved:

```
                        proposals with scope   without   refused
ja-JP                          42                43      markerListDeleteMenuItem
ko-KR                          91                93      + pluginWindowEditorViewMenuItem
en-US                          82                84      + pluginWindowEditorViewMenuItem
```

### The declarations state their own limit

`AXWindow[` says the element is inside a window, which is as much as a locale-independent path
fragment can say — Logic localizes window titles and titles a plug-in window by its track. It
refuses the menu bar, which is where the false matches came from; it would **not** refuse a
same-named item in another window. Both `reason` strings say so.

### Second commit

Generalising `path_contains` into a tuple changed what a malformed value does: `5` and `["a", 2]`
used to raise on the `in` comparison, and afterwards filtered to nothing and imposed no constraint.
A loud failure had become a false green, in the field whose whole job is to refuse a reading taken
somewhere else. Found reviewing the change before it shipped; one predicate now answers for all
three sites that carry such a fragment.

### Not claimed

Whether any *other* retraction has already been silently resurrected. Two were checked because a
ratchet named one of them; the ledger has 264 variants.

### Verification

34/34 repo guards. 110 guard self-test cases and 60 proposer cases. Mutation-tested: disabling the
provenance surface check, dropping the label fragment from either half, neutering the shape
validation, removing the proposer filter, deleting the malformed-value refusal, or dropping the
carry-forward each turns the suite RED.

Ship gate PASS. For #793.

The third commit adds the roadmap rows and observation records for #797 and #798, both filed today —
`roadmap-table-matches-github.py` and `check-observations-cover-live-walls.py` both required them.


MongLong0214:
## Adversarial review at `419e8030`, and what it found

Run blind against a worktree checked out at the exact pushed head, asked to break the change rather
than summarise it. It found **six defects — four holes and two tests that passed for the wrong
reason.** All six are fixed in `fb1ea502`, each reproduced through the guard's real entry points and
mutation-tested against its own case.

The reviewer could not execute the self-test scripts (its sandbox forbade creating their temporary
directories), so it read the code and exercised the guard against in-memory mutations of the real
ledger and real census records instead. That is worth stating: its coverage of the tests is a
reading, not a run.

### The two that mattered

**Surface scopes could be laundered through a shared evidence file.** `_rows` walks a record's own
observations *and* every file it lists as `evidence`, and a census evidence file holds rows from
more than one surface. Checking only the record's header was therefore not a scope at all wherever
an evidence file is shared:

> I set `markerListDeleteMenuItem.evidence_scope.surfaces` to `["arrange.track_headers"]`, marked
> en-US measured, and cited the real en-US track-header record. The only matching `Delete` row […]

A row's own `surface` is now checked too. Both gates stay — a record whose header is outside the
scope is still refused before its rows are read — and the new cases are built so that **only the row
check can be what refuses**: two records with identical headers, identical roles and identical
strings, differing solely in which surface the row inside their evidence file belongs to.

**`path_contains` was skipped for measured absence and identifier coverage.** Those two branches
select rows themselves instead of calling `sighting`, so a scope added at the `sighting` call sites
left both exempt:

> Using the real `"path_contains": "AXWindow["` declaration […] both of these passed the full guard:
> measured absence citing the application-menu record and locating the Edit menu's `Paste` item;
> identifier coverage citing the application-menu `delete:` item. […] `main_rc=0`.

This is the "fixed one half" shape `coverage_problems` already carries a comment about, one branch
further in.

### The two smaller ones

`"path_contains": null` was read as an absent key, so a declaration that reads as a constraint
imposed nothing. And the proposer read the scope's fields straight out of the object — a scope
written wrongly permitted everything, and `surfaces: 5` crashed the tool. It now fails closed
through the guard's own `scope_problems`.

### The two tests

`a list of fragments is accepted` only asked `scope_problems` whether a list is well formed; nothing
checked that every fragment is *required*. The malformed-scope cases called the helper directly,
which is exactly why none of them could see either hole above. Both now run through
`provenance_problems` and `coverage_problems`.

### The one not fixed

Renaming the Swift symbol drops the scope, because metadata is recovered by exact label name — the
same for `roles` and `retired`. Doing better needs the generator to be told which old name became
which new one, and nothing in the projection knows that. It is written down at the site rather than
implied away.

### What the reviewer tried and could not break

Worth as much as the findings, and quoted rather than summarised:

> - Provenance correctly ANDs the block fragment with the label fragment: satisfying either one
>   alone was refused; satisfying both passed.
> - Positive measured coverage likewise enforced both `coverage_paths[locale]` and label scope.
> - `5`, mixed-type lists, empty lists, dictionaries, and blank strings were reported as malformed.
>   `null` is the exception.
> - Direct record-surface mismatches were refused in provenance and coverage. **The bypass requires
>   a matching record header whose shared evidence contains another surface.**
> - Provenance-derived measured coverage could not be made clean while its provenance remained
>   invalid.

The three-locale control is unchanged after the fixes: 42/43, 91/93, 82/84, the same two labels
refused, every other label's proposals identical.

PR #829, opened 2026-09-08
fix: a record that cites nothing was an uncounted gap, and the executable's provenance was a timestamp

MongLong0214:
The second and third findings from the practices audit. The first — CommitLore — landed in #826.

## A record that cites nothing was invisible

Measured: **36 of 59** records carry no `evidence` entry at all. Their readings live only in their
own `observations` array, which the schema permits — and which **nothing counted**. So a new uncited
record cost nothing, and the ledger could learn strictly less about a newer claim while every
ceiling held.

`records_without_evidence` is now a ratchet dimension, held as a **set** for the reason every
dimension here is: a count is defeated by a swap, and a set names which record appeared. The 36 are
grandfathered with the reason in `raised` — retro-fitting captured output to readings taken months
ago would mean **inventing artefacts**, which is the opposite of what this directory is for.
Counted separately from `schema_v1_records` on purpose: upgrading a record's schema must not close
this gap.

Stated in the code rather than implied: this does **not** check that a cited file *supports* the
claim. Existence is checked elsewhere; whether the bytes say what the conclusion says is a heavier
rule, named rather than pretended away.

Mutations: dropping the dimension, and computing it as always-empty. Both red.

## The executable's provenance was a timestamp

`evidence.py` records `built_from` — the worktree's head — beside the binary's sha256 and compares
neither to the other, which is why it honestly sets `built_from_is_measured: false`. The substitute
is an mtime comparison, and on 2026-09-08 it produced a **false positive**: a rebase reapplied
identical content, moved a source's mtime, SwiftPM decided by content and relinked nothing, and a
binary that was exactly what those sources produce read as stale. The live gate then refused a
document whose every reading was clean.

Timestamps cannot settle this in either direction. `Scripts/build-provenance.sh` hashes the inputs
the compiler consumed:

```
content identical, mtime moved     mtime: STALE      binding: bound
a real source change               mtime: stale      binding: MISMATCH (inputs)
the executable replaced            mtime: silent     binding: MISMATCH (artifact)
no binding at all                  —                 cannot-tell, never a pass
```

The third is the one mtime could never see.

**A design correction the tests found rather than reasoning.** The first version digested
`Package.resolved` from disk and the binding flapped — every `swift build` rewrites it, so two emits
either side of one gate restore disagreed while the executable's hash was identical. It now digests
the **committed** pins, consistent with how the rest of the repository treats that file.

Not wired into the live gate here: making it required turns every existing evidence document without
a binding into a refusal, and that transition should land on its own where it can be watched.

## Also: `docs/tickets/TICKET-STANDARD.md`

A ticket is finished when the implementer needs **no** judgement. Two demonstrations from this
session: a ticket drafted on a reading that was accurate but about the wrong pane would have
re-implemented working code against another pane's schema; and a design question posed as a choice
between two options had a third answer that neither contained. Both were caught by checking the
premise, which a ticket specific enough makes possible.

Ship gate PASS: 4440 tests, 38 repo guards.


MongLong0214:
Superseded by what has since landed on `main`. Same check as #830, which was stacked on this one: every file exists on `main` in an equal or newer version.

- `Scripts/build-provenance.sh` / `Scripts/test-build-provenance.sh` — `main` captures and checks every reading and returns cannot-tell; this branch predates that.
- `Scripts/commitlore-capture.sh` / its test — `main` prints WHY a capture was rejected instead of `outcome=rejected staged=False` with no reason.
- `docs/observations/RATCHETS.json` — `main` already carries the `records_without_evidence` set this branch added, with a current membership rather than the one frozen here.
- `docs/observations/2026-09-08-the-region-origin-is-the-bridge-not-a-label.json` — `main`'s conclusion is the narrower one: it says the origin is absent from THAT SURFACE rather than absent, because the tempo and signature maps were never looked for. This branch still states the stronger claim.
- `Scripts/check-observation-ratchets.py`, `docs/tickets/TICKET-STANDARD.md` and the evidence files are identical.

Its `roadmap` job was red for a row that has since been fixed on `main`.

The branch is kept.

Issue #821, opened 2026-09-07
The release path cannot publish a tag once the release commit is merged — the stamp is unobtainable on main


MongLong0214:
`Scripts/release-stable.sh` published v3.16.0's tag locally and then could not push it:

```
pre-push REFUSED: no passing-preflight stamp for tree f4796bb2 (commit 79dc93a4)
```

The refusal is correct behaviour. What is broken is that **no honest route exists to satisfy it for
a merged release commit**, so the repository's own release path stops at the last step.

## Why the stamp cannot be obtained

`preflight-stamp.sh` keys on **(tree, HEAD commit object, BASE sha)** — deliberately, because an
amend can change what the preflight read. `lpm-ship.sh` records it, and only after every check
passes, because "a failed gate must not hand out its own bypass".

On `main`, every route to a passing gate is closed:

| attempt | outcome |
|---|---|
| `lpm-ship.sh .` with the default base | `FAIL: LPM_BASE=origin/main resolves to HEAD — every check would compare the branch with itself` |
| `LPM_BASE=<pre-release commit>` | gate runs and passes 4440 tests, then `LIVE-GATE FAIL: LPM_GATE_RULE_REF=origin/main resolves to HEAD — same commit, same judge` |
| `LPM_BASE=… LPM_GATE_RULE_REF=<same commit>` | `LIVE-GATE FAIL: LPM_GATE_RULE_REF=586783ec is not a remote-tracking ref` |

And the stamp recorded for the release **branch** does not transfer:

- linked worktrees have their own git dir, so the stamp lands in
  `.git/worktrees/<name>/lpm-preflight` and the main checkout looks in `.git/lpm-preflight`
- even in one checkout it would not match, because the key includes BASE, and after the merge the
  BASE the checker computes is the merge commit itself

## What is actually true of the tree being tagged

Everything the stamp attests was measured, twice, at `c054432b`:

```
public-surface preflight                      ok
repo guards (the set CI discovers)            ok
roadmap table matches GitHub                  ok
full suite                                    ok — 4440 tests passed
live evidence bound to c054432b               live_534 7/7, live_755 3/3, every dimension zero
```

Plus the PR's own CI: compile, roadmap and test all green.

So the content is verified and the token is unobtainable. That gap is the bug.

## Shape of a fix, not a decision

Whatever it is, it has to keep the property the stamp exists for — a RED gate must not mint one.
Candidates, in the order I would try them:

[...]

Issue #766, opened 2026-09-04
inferTrackType returns audio for every track: the header aggregate carries no type signal


MongLong0214:
`inferTrackType` classifies every arrangement track as `audio`. Measured live on Logic Pro 12.3 (build 6674), ko-KR, macOS 26.6 — three tracks created back to back, three different types, one answer:

```
logic_tracks create_instrument -> observed_track_name "Deluxe Classic"  observed_track_type "audio"   state A
logic_tracks create_audio      -> observed_track_name "오디오 4"         observed_track_type "audio"   state A
logic_tracks create_drummer    -> observed_track_name "SoCal"           observed_track_type "audio"   state A
```

## Why it happens

`AXValueExtractors.inferTrackType` (`Sources/LogicProMCP/Accessibility/AXValueExtractors.swift:737`) joins `AXDescription` / `AXTitle` / `AXIdentifier` / `AXHelp` / name of the header **and every descendant to depth 4** into one lowercased string, then runs substring tests in a fixed order (lines 768-776).

Every track header carries the Input Monitoring button, whose help text is:

> 입력 모니터링 버튼. 녹음 활성화가 되지 않은 **오디오** 또는 **소프트웨어 악기** 트랙에서 수신 신호를 들을 수 있습니다.

That one sentence satisfies `trackTypeAudio` (`오디오`) **and** `trackTypeInstrument` (`소프트웨어`, `악기`) on every header of every type. `.audio` is tested first, so `.audio` always wins.

## The signal carries no type information

Dumped the same aggregate the classifier reads, for all 7 headers in the project:

```
            Zulu   audio=1  instrument=2  drummer=0  external=0
  Deluxe Classic   audio=1  instrument=2  drummer=0  external=0     <- software instrument
         오디오 4   audio=3  instrument=2  drummer=0  external=0     <- audio
           SoCal   audio=1  instrument=2  drummer=0  external=0     <- drummer
           Alpha   audio=1  instrument=2  drummer=0  external=0
           Bravo   audio=1  instrument=2  drummer=0  external=0
    Studio Grand   audio=1  instrument=2  drummer=0  external=0
```

Identical token sets on every header. The only variation is `오디오 4`, and that is its **name**, not a type signal — a track called "Piano" and a track called "오디오 4" would classify differently for that reason alone.

[...]

PR #830, opened 2026-09-08
docs(#766): the header has no type signal in English either, and the strip that does needs no Mixer

MongLong0214:
Stacked on #829 — it carries that PR's two commits as ancestors. #829's own CI is red on the `roadmap` job for a row this branch fixes, so merging this one lands both.

## What was measured

#766 established in ko-KR that every track header classifies as `.audio`, and marked English as inference. Measured now on Logic 12.3 (6674), UI language `en`:

```
headers censused                     21
fields carried by EVERY header       23
fields NOT shared, per header          2   the name, and Track N “name”
the one exception                      4   a Track Stack's disclosure arrow
```

The method is what makes this a measurement rather than a restatement: the field sets are **intersected** across all 21 headers, so anything present on every header is disqualified as a discriminator no matter what it says, and only what is left can be a type signal. What is left is the track's name.

The English sentence is the Korean one — *"Input Monitoring button. Hear incoming signals on audio or software instrument tracks that aren't record enabled."* — on all 21 headers, satisfying `trackTypeAudio` and `trackTypeInstrument` together. The shipped classifier's `.unknown` is confirmed, not changed.

## The correction

`inferTrackType`'s own comment says the type signal lives in the channel strip and that reading it *"needs the Mixer revealed"*. It does not:

```
Audio 1           Input slot,  no MIDI Effect slot
19 instruments    MIDI Effect slot,  no Input slot
Absolute Zero     neither                            <- a Track Stack
both slots        0 strips
```

Read from the **Inspector's left strip**, main window, Mixer closed, at `window > AXGroup > AXList > AXGroup > AXLayoutArea > AXLayoutItem`, identified by its own `AXHelp` prefix. It follows the track selection in one or two 50ms polls on all 21 tracks. The zero in the last row is the clause that makes this a discriminator rather than a coincidence, and the Track Stack showing neither slot means a rule built on these two slots needs a third outcome, not a default.

## Both checks were made to fail first

Neither reverify script is trusted because it printed PASS. Each was run against mutated copies of the captured census:

| mutation | result |
|---|---|
| a header gains `AXDescription=Audio Channel Strip` | FAIL, names the header and the field |
| a strip shows both slots | FAIL, "the rule no longer discriminates" |
| no audio-shaped strip remains | INCONCLUSIVE, not PASS |
| a strip never agreed with its header | FAIL, "cannot support this rule" |
| unmutated control | PASS, exit 0 |

## Two holes found on the way

**`host.locale` was unvalidated** while `RATCHETS.json` groups coverage by it, matching the string exactly. A record naming anything else was counted in no locale and nothing complained — both records here said `en`, Logic's own preference string, where the axis says `en-US`. All 62 existing records already comply, so the rule lands with nothing grandfathered, and `known_locales()` reads the axis from the same file the ratchet does so the two cannot drift.

**`commitlore-capture.sh` printed `outcome=rejected` and nothing else.** The CLI had said which rule fired and on which record, in stderr and in its JSON; the wrapper dropped both, so a rejection was indistinguishable from a malformed draft. It reports them now, and the new case fails without the change.

## Scope

`inspector.channel_strip` is a new surface rather than a Mixer reading, which opens ja-JP and ko-KR gaps recorded with a dated raise. The slot names are English strings, so closing those by arguing that a slot's presence is locale-independent would be an argument standing in for a measurement.

Not claimed: this does not implement the strip-based classifier. It establishes that the signal exists, where it is, and what it costs to read — and that the header will never supply it.

The last commit corrects the #821 roadmap row, which the guard added earlier today refused to let past while the table said OPEN and GitHub said CLOSED.


MongLong0214:
Superseded by what has since landed on `main`, not abandoned. Every file this branch touches now exists on `main` in an equal or **newer** version, so rebasing it would replace corrected code with the drafts it came from.

Checked file by file against `origin/main`:

- `Scripts/livekit/ax_inspector_strip_type_census.swift` — `main` has the `kidsOrNil` split that keeps "the read failed" apart from "there are none", which this branch does not. That distinction is the reason the census can make an absence claim at all.
- `Scripts/build-provenance.sh`, `Scripts/test-build-provenance.sh` — `main` has the "a failed measurement must not become a measurement" hardening (every reading captured and checked, `return 2` on cannot-tell).
- `Scripts/observations/reverify-*.sh` — `main` has the aimed-instrument fix; this branch still has the version whose `else` reported an absence by failing to find a type.
- `docs/roadmap/README.md`, `docs/observations/*.json` — `main`'s rows and records are ahead.
- `docs/tickets/TICKET-STANDARD.md` and the evidence files are byte-identical.

Nothing here is lost: no file on this branch is absent from `main`. The measurement it carried — the English header census, and the correction that the strip read does **not** need the Mixer revealed — is on `main` and is what today's #766 work was built on.

The branch is kept.

Issue #856, opened 2026-09-11
ADR-009 history: Scripter target/readback investigation answered; MCU remainder is #862


MongLong0214:
# Scripter target/readback investigation — historical answer

> **SSOT revision: 2026-09-21. CLOSED / investigation answered, not generic plug-in control delivered.** Current stock-plug-in work is #292/#301, the concrete MCU targeting work is #862, and broader third-party expansion is backlog #306. #308 owns programme scheduling.

## What the investigation established

The existing Scripter route emits MIDI controller data into a host-side assignment configured by the operator. Validating or echoing a track/insert field inside the server does not establish that the transmitted controller identifies that hosted plug-in instance. A parameter that changes somewhere is not proof that the requested parameter on the requested insert changed.

[The final 2026-09-11 answer](https://github.com/MongLong0214/logic-pro-mcp/issues/856#issuecomment-5635985505) records the concrete request/wire mismatch: the insert field was constrained to the one supported shape, the track field did not select a different bus target, and the transmitted parameter was a controller number rather than a product-owned hosted parameter identity. This is a limitation of that measured route, not a universal statement that Logic cannot expose parameter values.

## The separate MCU result and its limit

[The earlier MCU experiment](https://github.com/MongLong0214/logic-pro-mcp/issues/856#issuecomment-5634754399) demonstrated a plug-in parameter movement and displayed feedback. The subsequent repeated-target investigation found re-banking and target-identity instability; the proposed verified writer was withdrawn before merge. The final answer explicitly transfers those findings, including delta-only LCD freshness, to #862.

Do not select the favorable first experiment and omit the failed repetition. Conversely, do not use the failed Scripter binding to block the independently working stock AX/EQ path.

## Current disposition

- Keep this investigation closed; it is not a second implementation queue for #862.
- Any future Scripter verification claim must first establish actual product-owned destination identity and independently observed result, not merely add another value reader or rename a provenance field.
- A new MCU operation requires its actual bank/page/target/freshness contract and repeatable positive/negative evidence under #862 and the selected owning feature.
- #292/#301 continue on their existing AX path without waiting for a generic MCU parameter provider.
- #306's broad vendor platform remains backlogged; #944's read-only library reuse is unrelated to completing this entire expansion.

[...]

PR #860, opened 2026-09-11
fix(#856): the MCU display reaches the parser

MongLong0214:
The UMP converter skipped SysEx7 whole, so every MCU display write was counted as "not read" and
discarded. That is where the surface says **what it is controlling, by name** — and `parseBytes` has
handled `.sysEx` all along, so the gap was only in the conversion.

Through the shipped path, with no spike:

```
rx_frames          111 -> 160
sysex_frames         0 -> 49
unconverted_words   49 -> 0
lcd  "DelCls DelCls DelCls StdGrn StdGrn ...  0  0  0  0"
```

## How this was found

#856 asked whether the MCU plane could carry a plug-in parameter with a readback. Chasing it turned
up something larger, recorded in `docs/observations/2026-09-11-the-mcu-plane-was-never-connected-and-plug-in-mode-works.json`:
**this host had no control surface installed at all.** Logic's Control Surface Setup listed zero
devices and its Input Port read `Invalid Port`. Four MCU button classes were verified sent and
produced no effect and no echo, while an AX selection readback moved in the same run — which is what
makes those zeros readings rather than silence.

With a `Mackie Control` installed and both ports bound, plug-in assignment mode remaps the V-Pots and
the surface reports it two ways: ring values (`86 -> [1,1,1,5,5,5,5,5] -> 86`) and the LCD naming
what it controls (`Cha EQ` per strip). The LCD is SysEx. That is this change.

## What it does

The `F0`/`F7` framing is implied by the SysEx7 status rather than carried in the data, so it is added
back. Two things are **counted rather than emitted**: a continue whose start this packet never saw,
and a message still open when the packet ends — half a display write is not a display write, and the
function is called per packet with no state between calls. An over-long `numBytes` is refused rather
than clamped, because a packet that lies about its length is not one to half-believe.

Eight new rows, and each branch shown able to fail: dropping the terminator, accepting an over-long
count, emitting an orphan continue, and no longer counting an unterminated message each killed the
row aimed at it.

## Two #736 rows pinned the old contract

They asserted SysEx7 is *skipped*, which this change makes false. Their subject is the **stride** — a
64-bit message costs two words, and a converter advancing one reads its tail as a header and invents
an event no device sent. That is kept: the witness moved to MIDI 2.0 channel voice, which is still
skipped, and SysEx7's own stride is asserted alongside the sweep rather than dropped from it.

Shown rather than claimed: breaking the SysEx7 stride to one word and breaking the MIDI 2.0 stride to
one word each fail the rows aimed at them. The coverage survived the edit.

## Not covered

Cross-packet assembly on real traffic. Every display write observed on this host arrived inside one
packet, so the multi-packet path has synthetic rows and no live evidence, and the source says so.

**#856 stays open.** The issue asks whether `.mixerSetPluginParam` can have a State A, and
that needs the `assignPlugin` route and a V-Pot write wired as well. SysEx was the critical path and
the piece that was silently discarding evidence.

### DIFF
--- a/Sources/LogicProMCP/MIDI/MIDIFeedback.swift
+++ b/Sources/LogicProMCP/MIDI/MIDIFeedback.swift
@@ -56,6 +56,12 @@ enum MIDIFeedback {
         var out: [UInt8] = []
         var unconverted = 0
         var i = 0
+        // SysEx7 assembly state, for a message split across several words of THIS packet. A message
+        // that starts here and ends in a later CoreMIDI packet is not assembled — this function is
+        // called per packet and holds no state between calls — and its words are counted as
+        // unconverted at the end rather than emitted as a truncated frame.
+        var sysExPending: [UInt8] = []
+        var inSysEx = false
         while i < words.count {
             let word = words[i]
             let messageType = UInt8((word >> 28) & 0xF)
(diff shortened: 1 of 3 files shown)


MongLong0214:
## The readback was already built end to end — one converter gap kept it from ever running

Checked after this change, and it reframes what the fix is worth. The whole path already existed:

```
MCUFeedbackParser.swift:65   case .sysEx(let bytes):
                               if let lcd = MCUProtocol.decodeLCDSysEx(bytes) {
                                   await cache.updateMCUDisplayRow(...)
StateCache.swift:701         func updateMCUDisplayRow(upper:text:offset:)
ResourceHandlers+StateReaders  MCUStateDTO { connection, display }   ->  logic://mcu/state
```

Parser, decoder, cache write and resource exposure: all present, none of it ever reached, because
the UMP converter discarded SysEx7 before `parseBytes` could see it.

`logic://mcu/state` now:

```json
"display": {
  "upperRow": "DelCls DelCls DelCls StdGrn StdGrn StdGrn StdGrn StdGrn ",
  "lowerRow": "                                                        "
}
```

Those are the eight strip names the surface is displaying, abbreviated the way an MCU LCD does it
(`DeluxeClassic`, `StudioGrand`).

### Why it looked healthy

`MCUDisplayState` initialises both rows to 56 spaces. So an empty display and a display that has
never once been written are the same value, and the resource has been serving the initial value as
though it were a reading. That is the absence-as-success shape this repository names in its own
review classes, sitting in a wire field.

It also means a caller could not have told the difference, which is why this went unnoticed while
the surface was unbound as well — two independent reasons for the same empty string.

### Scope, unchanged

This PR is still only the converter and its rows. Making the *display* a verified readback — a
caller asking "what is the surface showing" and getting an answer bound to a moment — is a separate
question, and `lowerRow` staying blank here is not evidence either way about whether Logic writes it.

Issue #864, opened 2026-09-11
logic_edit undo reports success and does not undo — it sends a CC nothing is assigned to, and the cgEvent fallback is never reached


MongLong0214:
## What happens

```
logic_edit undo {}
-> {"state":"B","success":true,"reason":"readback_unavailable",
    "operation":"edit.undo","method":"midi_key_command","cc":30,"channel":16,
    "raw":"Key command triggered: edit.undo (CC 30 CH 16)"}
```

Nothing is undone. Measured 2026-09-12 on Logic 12.3 (6674): two `insert_plugin` calls left a Compressor on tracks 1 and 19; `logic_edit undo` was driven twice and both plug-ins were still there on a refreshed re-read of `logic://mixer`. The same two undos performed by hand (`Cmd+Z` with Logic frontmost) removed both, so Logic's undo stack had exactly the entries the command claimed to be popping.

## Why

`RoutingTable.swift:198` routes it `[.midiKeyCommands, .cgEvent]`. The MIDI key-command channel sends **CC 30 on channel 16** — a controller number that only does anything if the operator has bound it in Logic's *Controller Assignments*. This project has no such binding, and the product can neither create one nor enumerate it, so the send lands nowhere.

The `.cgEvent` fallback — which would post the actual `Cmd+Z` and does work — is **never reached**, because the MIDI send "succeeds" at the wire level. A send-only channel that cannot fail cannot fall through.

## Why `success: true` is the sharp end

State B with `readback_unavailable` is the honest label for a send-only write, and in isolation this response is not lying. The problem is what it is used FOR: **undo is the rollback primitive.** `AccessibilityChannel.undoLastLogicAction` is the `rollback()` that `insert_plugin` calls when a verified insert fails its readback, and a live harness that restores the project by driving `logic_edit undo` records a restoration that did not happen.

That is not hypothetical — it is how this was found. `live_855_the_insert_leaf_is_the_strips_configuration` used `logic_edit undo` for its restore step, and the run reported `restorations_failed: 2` and refused to pass. The harness now clicks Logic's own `Edit ▸ Undo` menu item instead and verifies by re-reading the strip.

## Acceptance

- `logic_edit undo` either undoes, or fails closed — it must not answer `success: true` for a write whose effect it cannot establish and whose destination is an assignment the operator never made
[...]

PR #868, opened 2026-09-11
fix(#851): health published MCU staleness twice, so one payload could contradict itself

### LINKED ISSUE #851: health publishes MCU staleness twice from two cache reads, so one payload can contradict itself
## What is wrong

`logic_system health` publishes the MCU staleness answer in two places:

- `mcu.feedback_stale` — the machine field
- `channels[].detail` — the word `stale` or `active` in the MCU line

They are one *rule* as of the staleness unification (`MCUConnectionState.isFeedbackStale(now:)`),
but they are read from **two different cache snapshots** in the same response:

```
Sources/LogicProMCP/Dispatchers/SystemDispatcher.swift:371   let report = await router.healthReport()   // MCUChannel.healthCheck reads the cache
Sources/LogicProMCP/Dispatchers/SystemDispatcher.swift:386   let mcu = await cache.getMCUConnection()  // a second, later read
```

Inbound MCU feedback arriving between those two reads — or a `stop()` — re-stamps
`lastFeedbackAt` and flips `isConnected`. A single payload can then say `feedback_stale: true`
next to a channel line reading `feedback active`, or the reverse. Both halves are individually
correct; the payload as a whole describes a state that never existed.

This is the snapshot-and-certificate-are-not-atomic shape, not a duplicated-constant shape. The
duplicated `5.0` literal was a separate defect and is already fixed; unifying the rule did not and
could not close this one, because the gap is between the two *reads*, not between two spellings.

## How it was found

Named by an independent reviewer while answering an unrelated question about the duplicated
threshold. It was not hypothetical for long: the live harness written for that fix originally
asserted the two fields match, and had to be rewritten, because a green from that assertion would
have been luck rather than proof. That rewrite is recorded in
`Scripts/livekit/live_849_one_staleness_answer_reaches_both_surfaces.py`, which now deliberately
compares the bit only against its own snapshot and says why.

## Why it has not bitten

The window is microseconds and only closes when Logic sends MCU feedback in it. Nothing observed
it in the wild. That makes it low-frequency, not absent — and a reader of a health payload has no
way to tell an unlucky read from a real disagreement.

## What would fix it

Read the MCU connection **once** and give both surfaces the same value, rather than letting the
channel and the dispatcher each reach for the cache. That means threading a snapshot through
`healthReport()`, which is a wider change than the staleness unification wanted to make, which is
why it is filed rather than folded in.

## Acceptance

- one `logic_system health` response derives `mcu.feedback_stale` and the MCU `channels[].detail`
  word from a single cache read
- a test that fails when they are taken from two reads, not merely one that passes when they agree
- the live harness above may then assert the two fields match, and its docstring's exclusion note
  is removed rather than left describing a gap that closed

MongLong0214:
Closes #851.

`mcu.feedback_stale` and the word in `channels[].detail` shipped in the same document and came from **two separate reads of the cache** — `router.healthReport()`, which reached `MCUChannel.healthCheck` and let it derive the word from its own read, and `cache.getMCUConnection()` a few lines later for the wire field.

Feedback arriving between those two reads puts `detail: "…feedback stale"` beside `feedback_stale: false`, with neither field wrong at the instant it was taken and the document wrong as a whole.

### This is the check `live_849` refused to write

Comparing those two fields was that harness's first shape, and it was rejected for exactly this reason — green by luck, red on a race, *"a flake wearing the shape of a proof"* — and filed as #851 instead of being written. The new harness is that check, now that it is a statement about the product rather than about timing.

### The fix

The rule is evaluated **once per payload**. `MCUChannel.healthCheck` renders no staleness at all and reports what it can answer without a clock: whether the port carries and whether Logic registered the device. The dispatcher takes one snapshot, computes staleness once, and the same boolean produces both the wire field and the clause appended to the MCU row. Agreement is structural — there is no second evaluation left to disagree, and no second clock to disagree by.

A disconnected port gets **no clause in either direction**. It is not stale under the rule and it is not active either, and a renderer that read only the boolean would print `feedback active` on a dead port — which is the mutation the live harness catches deterministically.

### On what the live run can and cannot show

It does **not** show that the race used to happen. Nobody caught the two fields disagreeing on a live server; the defect was read out of the two call sites, and a run against the old build would very probably have been green. The harness says so in its own docstring rather than implying otherwise. What carries the structural load is `MCUFeedbackStalenessRuleTests`, where the renderer is pinned as a function of its argument and the channel is pinned to render no staleness word at all.

What the live run adds is the half a unit test cannot reach: that the **assembled** payload a real client receives is self-consistent, across both answers, on a live server — and the sweep is required to contain both, because a surface that is permanently stale and permanently says so agrees with itself forever.

### Tests

The two call sites that read the word off `healthCheck` now read it where it lives. They were asserting the second derivation, which *is* the defect; `testLogicProCrashDetection` keeps its subject by asserting the cached state the dispatcher renders from.

### Gate

```
full suite                    ok
live evidence for this head   ok
repo guards                   ok
SHIP GATE PASS
```


MongLong0214:
Roadmap check is red for a sequencing reason, not a content one: #864 was closed while this branch was already open, so this branch's table still lists it as OPEN. It will be resolved by rebasing onto main once #869 lands — the row comes from there. Holding this PR until then rather than adding a row it does not own.

MongLong0214:
Rebased onto main now that #869 has landed, which resolves the roadmap failure: #864's row comes from main rather than being carried here. The branch was force-pushed with lease after comparing content — the difference against the stale remote was exactly #869's merged work plus this branch's own roadmap row and ratchet commit.

#851 is now closed with its row flipped, which the gate required: it refuses a PR that carries a closing keyword while the table still says OPEN.

Re-gated at 6895afe9 — all checks pass, full suite included.

PR #869, opened 2026-09-11
fix(#864): an undo that reports success now has to have undone something

### LINKED ISSUE #864: logic_edit undo reports success and does not undo — it sends a CC nothing is assigned to, and the cgEvent fallback is never reached

MongLong0214:
Closes #864.

`edit.undo` routed `[.midiKeyCommands, .cgEvent]` and the MIDI rung sent **CC 30 on channel 16** — a controller number that does nothing unless the operator bound it in Controller Assignments, which this product can neither create nor enumerate. Because a send-only channel succeeds at the wire, the `.cgEvent` rung that would post a real `Cmd+Z` was **never reached**. Two inserted plug-ins survived two `logic_edit undo` calls, each answering `success: true`.

Undo is the **rollback primitive**, so a rollback that did not roll back could report that it did.

### The fix

AX leads the chain. The Edit-menu entry is pressed and read back: State A means the entry named *before* the press is not the entry named after, so the stack demonstrably moved. Matching names is State B `noop_unobservable`, not a claim — Logic can legitimately word two entries the same way. A disabled entry is refused with `write_attempted: false` before anything is pressed.

### The row is found by its shortcut, and the first attempt proved why

The wording is **localized and state-dependent**. With an empty stack Logic writes `Can't Undo`, which does not *start* with `Undo`, so a prefix match skips it and lands on `Undo History…` — the one title that does. Widening to containment is worse, not better: three Edit-menu titles carry the word, and pressing `Undo History…` opens a window instead of undoing anything.

That is not hypothetical. It is exactly what my first version did: both reads named that row, so every undo reported `noop_unobservable`.

Measured in `docs/observations/2026-09-12-the-undo-entry-is-identified-by-its-shortcut-not-its-wording.json` — undo is ⌘Z at mask 0, redo mask 1, `Undo History…` mask 6. A host that remapped ⌘Z matches no row and is refused rather than having a different row pressed on its behalf. **No `Redo` label set was added**: that would be a second authority claiming the row can be found by its wording, which the measurement disproves.

The menu open is confirmed by looking for the `AXMenu` child rather than by `AXPress`'s return value — this operation opens the menu three times and `AXPress` on a menu-bar item *toggles*, so a press arriving while the previous Escape was still settling would close it, reported as "the entry vanished": a confident false statement about Logic caused by our own pacing.

Unit tests are built from the live reading, `Undo History…` included, and were mutation-tested: dropping the modifier comparison kills four of them.

### A second defect, fixed here because the rules required it

`live_778_step_input_toggle_reaches_its_menu_item` claims `AccessibilityChannel+Editing.swift`, so touching that file obliged the harness — and it was **red**, three of five checks. `edit.toggle_step_input` answered `Window > Step Input Keyboard was not found` on a menu that plainly carries the item: the path matched `.exact` against `Step Input Keyboard`, and Logic offers `Show Step Input Keyboard`.

Two further measurements shaped the fix rather than a guess: the title reads `Show …` in **both** states (it never becomes `Hide …`), and clicking it while open still closes it — so the verb describes nothing and cannot be relied on. The path matches by containment now. Live: both directions reach State A with the window state observed.

### Not done, and not claimed

`undoLastLogicAction` — the rollback `insert_plugin` calls — still posts a bare `Cmd+Z` CGEvent and returns `true` unconditionally. It is a separate write path and deserves its own change rather than being folded in. Called out on the issue too.

### Live evidence

Three harnesses at this head. `live_778_japanese_ax_reads_resolve` is obliged by the `AXLocalePolicy` change and needs a **Japanese** Logic UI, so Logic was switched to `ja`, run (7/7), and switched back to `en` with the setting restored to its original value.

```
full suite                    ok
live evidence for this head   ok
repo guards                   ok   (42 pass; locale-labels and step-input guards both fired first, legitimately)
SHIP GATE PASS
```


MongLong0214:
**Verified the `edit.redo` claim rather than leaving it asserted.** The PR says redo "moves with it" on the strength of sharing a handler; the live path had only been driven for undo. Measured at this head:

```
rename         -> State A            name: lpm-864-redo
edit.undo      -> State A   'Undo Renaming' -> 'Can’t Undo'    name: Audio 1
edit.redo      -> State A   'Redo Renaming' -> 'Can’t Redo'    name: lpm-864-redo
edit.undo      -> State A                                       name: Audio 1   (restored)
```

Both directions reach State A with the entry moving, and the track name — read from `logic://tracks`, independent of the menu — follows each one. The project is back to its starting state.

Worth noting what `Can’t Redo` confirms: the entry is found while **disabled**, by its shortcut, with a title that carries neither `Redo` as a prefix nor any usable verb. That is the case the shortcut-identity design exists for, observed live rather than only in the fixture.

The harness itself still drives undo only. Extending it to redo would mean another commit, and this branch's Japanese-UI evidence would have to be re-run for it — noted as follow-up work rather than done here.

MongLong0214:
**Refinement to the step-input half of this PR — the fix is right, my stated reason was half of it.**

The commit says Logic's Window menu "offers `Show Step Input Keyboard`". Measured more carefully afterwards, the title is not stable:

```
before the Window menu has ever been opened : "Step Input Keyboard"
while the menu is open                      : "Show Step Input Keyboard"
after closing it again                      : "Show Step Input Keyboard"   <- stays
```

The verb is **lazily populated**: it appears the first time Logic opens that menu and persists from then on. The product's read does not open the menu, so it sees whichever spelling the *human* last left behind.

So `.exact` against the bare name matched on a Logic whose Window menu had never been opened, and failed permanently once anyone opened it. That is why this could sit broken without being noticed, and why the failure looked non-deterministic: **the operation's success depended on whether a person had ever pulled down the Window menu.**

Containment covers both spellings, so the fix stands as merged — it is state-independent, which is what this input requires.

Checked against the prior `[directive]` on `AXLocalePolicy.swift` that ruled out containment for the metronome (`switch metronome matching to containment | it would let 再生 inside another label be taken for Play`). That hazard is real for a two-character label and does not reach here: across **48** Window-menu items, exactly **one** contains any of the three step-input labels. I should have consulted that record before making the change rather than after — noted.

This is the same lazy-population mechanism the #864 observation record left open as a limit for the Edit menu (*"whether it is the FIRST open in the process, or every open, is NOT established"*). For the Window menu it is now established: first open populates, and it stays.

Issue #876, opened 2026-09-12
Feature: Please add compatibility for german version of Logic Pro


Dominik-1980:
### Preflight

- [x] I searched existing issues for related work.

### Priority

P3 - nice-to-have polish or convenience

### Primary area

logic-ui

### Problem

The MCP Server only works for English and Chinese Versions of Logic Pro.

### Proposed outcome

Add compatibility for German Version.

### Safety and contract notes

None

### Done when

It completely works for german Version as well as for the others.

MongLong0214:
Thank you for filing this — you found a real gap, and measuring it corrected two things I had wrong.

**The premise was wrong in both directions.** The server does not work for Chinese: `zh` appears in
exactly two string arrays and in **no** `LabelSet` variant at all. What it actually supports is
**English, Korean and Japanese**. German was at zero, exactly as you said.

**German cannot be translated in; it has to be read off a German Logic.** The first alignment of an
English and a German census makes that concrete:

```
Logic Pro          ->  Logic<U+00A0>Pro      (a NON-BREAKING space between the words)
MIDI File…         ->  MIDI-Datei …          (German puts a space before the ellipsis)
File               ->  Ablage                (not "Datei")
To Playhead        ->  Für Abspielposition
```

Nobody types a non-breaking space, and `Ablage` is not what a translator produces from `File`.
A guessed label fails silently on the one machine its author does not own — which is precisely the
machine that matters here, yours.

**What has been measured and is landing**

Logic's UI language turns out to be a `defaults` write on `com.apple.logic10`, so every language
Logic ships can be measured on one machine. `de-DE` has joined the locale campaign the same way
`ja-JP` did:

| | |
|---|---|
| de-DE records + evidence | 6 surfaces, the same six the English walk produces |
| labels measured in German | **29 of 145**, each with provenance citing the record it was read from |
| menus proven live | File / Edit / View / Window / Navigate / Track, plus Export, Import, Save As, Bounce, Create Marker, Open Marker List, Show Tempo List, Move ▸ To Playhead |

There was also a **structural** blocker, and it is fixed rather than worked around. The guard that
checks a label against a census decided which language a string belonged to by looking at its
**script** — Hangul is Korean, kana is Japanese, ASCII is English. That works only while every
supported language has a script of its own. `Bearbeiten` is ASCII and would have been filed under
English; `Löschen` matches none of the three and would have been skipped entirely. German was
[...]

PR #877, opened 2026-09-12
feat(#876): 독일어는 번역이 아니라 측정이었고, 그 과정에서 제품 결함 하나와 관측 도구의 실명 하나를 찾았다

MongLong0214:
Closes nothing. #876 stays open, with a smaller and better-defined remainder.

## What the report got wrong, in both directions

It says the server works for English and Chinese. It works for **English, Korean and Japanese**;
Chinese appears in exactly two string arrays and in **no `LabelSet` variant at all**. German was at
zero, exactly as reported.

## Why this could not be a translation

The first alignment of an en-US and a de-DE census settles it:

```
Logic Pro     ->  Logic<U+00A0>Pro      a NON-BREAKING space between the words
MIDI File…    ->  MIDI-Datei …          German puts a space before the ellipsis
File          ->  Ablage                not "Datei"
To Playhead   ->  Für Abspielposition
```

Nobody types a non-breaking space. A guessed label fails silently on the one machine its author does
not own, which is the machine that matters.

## The wall that was not one

Logic's UI language is a `defaults` write on `com.apple.logic10`. **Every language Logic ships can
be measured on this machine** — de, en, es, fr, it, ja, ko, pt, zh-Hans, zh-Hant. The repository
already had the whole campaign for it (`locale-campaign.sh` / `locale-align.py` / `locale-propose.py`);
de-DE joined the way ja-JP did.

## A live run, not a table

On a German Logic the menu bar reads `Ablage / Bearbeiten / Spur / Navigieren`, Edit ▸ Move ▸ To
Playhead resolves entirely out of policy labels, and `transport.goto_position` drives the playhead to
bar 9 through the German dialog route with the control bar's position readout visibly changing.

**That run found a product defect the tables could not.** `goto_position` failed
`dialog_unidentified_new_window` — the menu leaf fired, the window opened, and nothing could name it.
Two layers:

1. The German dialog title was unmeasured. It is `Zu Position`, read by **opening** the dialog twice.
   A navigation-free census opens nothing, so the string occurs there 0 times.
2. **Fixing that changed nothing.** The decision lived in a literal title list inside the AppleScript
   — written **twice**, in the position route and the stray-UI sweep — restating what
   `AXLocalePolicy.goToPositionDialogTitle` already declared. The menu path beside it had been moved
   onto the policy for #519; this half was left behind, and its own comment said so.

The handler is rendered from the LabelSet now, and it is deliberately **narrower** than the literal by
one string: `Go to Position` had no provenance, `.exactStrict` is case-sensitive so the Swift side
never accepted it, and two members differing only by case are refused by `check-probe-product-drift.py`.

## The structural blocker, fixed rather than worked around

`check-variants-appear-in-a-census.py` decided a variant's locale from its **script** — Hangul is
Korean, kana is Japanese, ASCII is English. That holds only while every supported language has a
script of its own. `Bearbeiten` is ASCII and would have been filed under en-US; `Löschen` matches
none of the three and would have been skipped. German was unrepresentable.

It derives the locale from the evidence now, which is also **stricter**: 99 absences became 117,
because strings that pinned to no script had been quietly excused.

## Two tests were defused by supporting German

Both used German as their example of an **unknown** language, so supporting it turned them green.
One moved to a language Logic does not ship (Croatian), one to a locale that cannot exist (`zz-ZZ`).
A fixture that uses a real thing as its example of a missing thing dies the day that thing arrives.

## An observation tool that had gone blind

`is_clean` retired the German run for `blocking_modal: cannot_tell`. It looked like timing and was
not: a plain Python process has no run loop, so `NSWorkspace`'s application list still held the
**dead** Logic after a relaunch, and `AXUIElementCreateApplication(dead pid)` answers `-25204`
forever. The check taken before the first quit recorded a clean `null`; all four after it recorded
`-25204`, one of them at the same moment `located_band` read the live tree successfully.

Pumping the run loop and dropping terminated entries fixes it, proven in one process (before:
`null` → `cannot_tell`; after: `null` → `null`), with a regression test. This affected **every
harness that relaunches Logic**, not only this one.

## What is NOT done, and its size is a set rather than a sentence

* **114 of 145 labels are still unmeasured in German** — window titles, dialog buttons, column
  headers, plug-in menu leaves and the editors, none of which a navigation-free census walks.
* **Eight surfaces have no German record.**
* **`tracks.record_sequence` does not work in German.** The SMF import panel is localised too and it
  refuses at `preflight_blocking_dialog` naming a dialog called `Importieren`. The passing run drives
  that refusal itself and records it — a run that quietly picked the operations that pass would be
  measuring its own selection.
* **Six German variants were READ but cannot be CITED.** `locale-propose.py --apply` keys records by
  the census row's raw `surface` field and every window row carries `arrange.window`. That is a
  pipeline defect affecting ko-KR and ja-JP identically, and it is deliberately not fixed while
  adding a locale.

All four are sets in `docs/observations/RATCHETS.json`, so they can only shrink.

## Evidence

| harness | result |
|---|---|
| `live_876_a_german_logic_is_driven_by_measured_labels` | 5/5, every dimension clean |
| `live_778_japanese_ax_reads_resolve` | 7/7 |
| `live_519_region_op_on_a_localized_logic` | 8/8 |
| full suite | 4525 tests |
| repo guards | 42/42 |


MongLong0214:
**Not merging this yet. The pre-merge independent review could not run.**

This repository's rule is that a PR does not merge on my own review plus green CI — an independent
reviewer has to read the diff first. That rule exists because my own gate was measured insufficient:
three wrong defect claims in one day, none of which I caught myself.

Both reviewers are unavailable right now:

| reviewer | state |
|---|---|
| `codex -m gpt-5.6-sol` (xhigh) | usage limit, retries at 2026-09-15 10:24 |
| `grok` | `402 Payment Required — Grok Build usage balance exhausted` |

The blind packet is prepared and unchanged, so the review can run the moment either is back: a
worktree at this exact head with `CHANGED.txt` and a code-only `DIFF-code.patch`, and a
question-form prompt that does not state my conclusions. It asks six things, three of which are
where this change could actually be wrong:

1. does anything here accept a German string that was **not read off a German Logic**;
2. is the new locale rule in `check-variants-appear-in-a-census.py` **weaker than the old one in
   any case** — construct it;
3. for each edited test, is it **adapting to a deliberate change or removing a check that would
   still be catching something** — and if the latter, show what the old version caught.

What is already established, and is not a substitute for the above: SHIP GATE PASS, 4525 tests, 42
repo guards, and three live harnesses clean at this head (German 5/5, Japanese 7/7, Korean 8/8),
each verified with the gate's own `is_clean` rather than by reading a subset of its counters — I
made exactly that mistake earlier today and the gate caught it.

Leaving this open and green.

MongLong0214:
## Self-review, and it is labelled as such

The owner directed a self-review and merge while both independent reviewers are down
(`codex` usage limit until 2026-09-15, `grok` `402 Payment Required`). **This is not the
independent review the rule asks for and does not stand in for one.** It is the same six-question
blind packet, answered by the author, with the evidence attached so a later reviewer can check the
answers rather than repeat the work.

### 1. Does anything accept a German string that was not read off a German Logic?

**No.** Checked mechanically rather than by eye — every variant this branch adds to
`AXLocalePolicy`, diffed against `origin/main`, looked up verbatim in the de-DE measurements:

```
strings added by this branch: 30
NOT found verbatim in a de-DE measurement: 0
```

### 2. Is the new locale rule weaker than the old one in any case?

**Yes, in one, and it is a deliberate trade.** A variant that pins to locale X, is absent from X's
census, but appears in another locale's census is now excused where it used to be reported. It is
reachable: `Mixer` is a word Logic keeps untranslated in German and Japanese, so an English variant
that en-US Logic does not actually show could be excused by a ja-JP sighting.

What buys it: the old rule could not express German **at all** (`Bearbeiten` is ASCII, `Löschen`
matches nothing), and it was silently excusing 18 strings that pinned to no script. Net, the guard
reports more, not less — 99 absences became 117.

Per-locale attribution still lives where it always did, in `coverage` / `provenance` and the
ratchets, and this change does not touch those.

### 3. For each edited test: adapting, or removing a check?

| test | verdict |
|---|---|
| `Issue60LocalePhase4Tests` | **adapting.** Two measured German strings added to a frozen token set; the file already records the same procedure for `リージョン` and for Japanese. Any other drift still fails. |
| `Issue529MenuValidationTests` | **strictly stronger.** It counted one literal twice in the source; it now requires **zero** hand-written predicates anywhere in the source and requires the rendered handler to contain **every** title the LabelSet declares. The old mutation (remove the JA title) still fails it; a returning literal list now fails it too. |
| `Issue523MarkerDeleteMenuTests` | **adapting, and hardened.** Its "unknown locale" fixture was German, so supporting German turned it green. Moved to Croatian, which Logic does not ship. |
| `test_observation_records.py` | same shape; `de-DE` → `zz-ZZ`, which cannot become a locale. |
| `test_variants_appear_in_a_census.py` | **a real rule change, declared.** One-locale-settled no longer suppresses; a new case asserts the all-settled suppression that replaced it. The fixture gained the en-US census it had always implicitly assumed. |

### 4. Does the generated AppleScript behave identically for the shared titles?

```
old literal : ['위치로 이동', '位置の移動', 'Go To Position', 'Go to Position']
new rendered: ['Go To Position', '위치로 이동', '位置の移動', 'Zu Position']
dropped     : ['Go to Position']      deliberate, documented in the LabelSet
added       : ['Zu Position']         measured
quote/backslash in any title: []      nothing to escape, and such a title is filtered out
```

Empty label set renders `if false then return true` — it refuses every window rather than accepting
any. That is the safe direction.

### 5. Can `evidence.py` now report a CLEAR scan where it would have reported a blocker?

**One narrow case, stated rather than denied.** If `NSWorkspace` has not yet listed a
just-launched Logic, `_live_logic_apps` returns empty and the AX signal reports a completed clear
scan. Two things bound it: the CoreGraphics signal runs independently and sees any modal that is
actually rendered, and the run-loop pump this change adds exists precisely to shrink that window.
The residual is a modal CoreGraphics cannot see — inactive Space or off-screen — on a Logic that has
not been listed yet.

Against that: the old behaviour was **permanent blindness** after the first relaunch, which is not a
conservative failure — it is a detector answering cannot-tell forever while a real modal sits there
unnoticed.

### 6. Anything that would let a future run report success while broken?

The one I found and fixed during this review cycle was mine: the run drove `record_sequence`
before its own assertion, leaving Logic's German import panel open, and `is_clean` correctly
retired the whole run. It is driven last now, and the panel is dismissed and the dismissal
confirmed.

`check-variants-appear-in-a-census.py` remains **counted, not gated** — 117 absences do not fail
CI. That predates this branch and is unchanged by it.

### Standing evidence

SHIP GATE PASS · 4525 tests · 42/42 repo guards · three live harnesses clean at this head
(German 5/5, Japanese 7/7, Korean 8/8), each verified with the gate's own `is_clean` rather than a
subset of its counters.

PR #882, opened 2026-09-14
WIP: restore the qualification subsystem — held: the gate is opt-in and off, its verifier is 7,522 lines stale, and its evidence fails its own shape rule

MongLong0214:
> **Scope corrected 2026-09-19.** The product half of this branch — the router change, three LabelSets, eight observation records and the AX sheet census — moved to **#922**, rebased onto today's `main` and passing all 67 guards. What remains here is the qualification subsystem, and it is **not mergeable in its current form**. Three independent reasons, each reproduced from the source rather than repeated from a review.

## 1. The gate is opt-in, and it is off

All seven gate steps in `release.yml` carry:

```yaml
if: ${{ vars.ADR001_QUALIFICATION_ENFORCED == 'true' }}
```

`ADR001_QUALIFICATION_ENFORCED` is set **nowhere** — not in the repository, and not as a GitHub Actions variable (`gh api repos/…/actions/variables` returns nothing). Restoring 9,534 lines would change what a release does by exactly nothing.

This branch's own test file already says it: `Tests/LogicProMCPTests/Issue815OptInGateTests.swift:11` — *"An opt-in gate is not a gate"* — and `:44` asserts all seven steps are skippable.

## 2. Flipped on, the verifier that runs is a July build

```yaml
ref: 67fec5fccf817f9978022f0ca262b80a789e2dce
```

`git diff --stat 67fec5f <this branch> -- Sources/LogicProMCP/Qualification` is **10 files, +7,522/−231**. The pinned tree has no `SemanticOracles.swift` and no `SemanticOracleTable.swift` — the machinery Phase B is built on. It also knows 107 `OperationID` cases where the tree has 113, so six operations can break and still pass the enforced gate.

## 3. The Phase-B evidence cannot pass this branch's own verifier

A mutating case reaches `.passed` only through `expectedZeroWriteRefusalObserved`:

```
QualificationTransport.swift:338   guard expectedZeroWriteRefusalObserved else { return .failed }
                      :373         isError == true && state == "C"
QualificationRunner.swift:383      operationIsError: operationResult?.isError     // → true
                      :1904        && evidence.operationIsError == false          // requires false
```

`true == false` is not satisfiable. Every Phase-B case that reaches `.passed` is structurally rejected by the verifier in the same commit.

**Nothing caught this** because no test puts a Phase-B record through verify: the argument label `mutationRestore:` appears **zero times** in `Tests/`. The records that ARE verified are synthetic fixtures with `preState: "Before", mutation: "After"`.

## Also to settle before this lands

- **The PR's numbers disagree with each other.** The body says `passed 20 → 26`; this branch's own `docs/roadmap/README.md` says `20 → 33`; `Fixtures/qualification/README.md` says `36 … 38`; an observation record says `sweep_passed: 36`. No artifact reproduces 20 or 26, and `--qualify`'s `exit 0` is an unconditional `return` at `QualificationRunner.swift:212`.
- **`--qualify` mutates the open project and the body does not say so** — it renames track 0 to `qualification_phase_b_probe`, creates a marker, writes the tempo and moves a fader. This branch's own records document leaked markers and tracks.
- **`tracks.list_library` was closed as unfixable on a wrong reading.** The body says `scan_library` never writes the cache file. It does: `AccessibilityChannel+Library.swift:270`, and `parseScanMode` returns `.disk` for an empty mode, so the DEFAULT scan writes `library-inventory-disk.json` while the reader looks for `library-inventory.json` — a filename split whose reader half was never written. Nothing in `Sources/` reads the `-disk` name.
- **95 commits stale.** Its last green CI ran a `roadmap` job that no longer exists; the guard count went 31 → 67 while it sat.

## What would make this mergeable

Answer 1–3. A gate whose enforcement is a repository variable nobody sets, whose pinned verifier predates the evidence it must check, and whose evidence its own shape rule rejects, is three separate things that each need a decision — not a rebase.


MongLong0214:
## Independent review (gpt-5.6-sol, xhigh, read-only) at the exact head `fc74d373` — verdict BLOCK

Scope given to the reviewer was the 23 files that are new or changed against `a5ea8674`, the last commit before this subsystem was deleted; the other 16 files were restored byte-identical and were declared out of scope. The reviewer reported opening every file in that list.

I did not forward the verdict on its own. Three of the five blockers I then read myself at this head, and one of them I had already found independently before the review ran. Those notes are marked below.

---

### BLOCKER 1 — an unavailable readback can become a recorded PASS

`QualificationTransport.swift:301` promotes any non-nil mutation record once `readbackFreshness` is admissible, but `:555` evaluates the later generic operation readback rather than the record's own pre/mutation/restore readings. Those recipe readings parse only `data` and ignore `source`, `readable` and `ax_occluded` (`:1794`). And `QualificationReadbackFreshness.swift:131` returns `.admissible` immediately for any policy that is not `.readbackRequired` — which includes `mixer.set_volume` and `mixer.set_pan`, both `.none` at `OperationRegistry.swift:546`.

Scenario: the volume recipe reads 0.4 → 0.5 → 0.4 from cached envelopes; the following `logic://mixer` read answers `data_source:"mixer_not_visible"`; the policy is `.none` so that unavailable read is "admissible"; status becomes `.passed` and `:254` publishes the unavailable artifact as `verified:true`.

**Not verified by me.** This is the one finding I did not re-derive, and it is the most consequential, so it should be checked before it is accepted or dismissed.

### BLOCKER 2 — the Phase-B cycle qualifies a bare index 0 and can corrupt another track on restore — **I verified this**

`QualificationTransport.swift:1826` is literally `let trackIndex = 0`. Both the mutation (`:1837`) and the restore (`:1853`) send `params: ["index": trackIndex, ...]`. `observedTrackValue` reads `rows[trackIndex][field]` and returns only the `Double` (`:1794`) — the row's stable reference is right there and is discarded.

Scenario: track A is index 0 at 0.4, track B is 0.8. The user reorders B to index 0 after the pre-read. The recipe moves B to 0.5, sees a change from A's 0.4, and "restores" B to A's 0.4. Every comparison passes, a qualification record is emitted, and the operator's track B is left at a value it never had.

This is the positional-targeting rule this repository applies everywhere else, and the handoff says the Phase-C track cycles were given a carried identity for exactly this reason. Phase-B volume/pan did not get one.

### BLOCKER 3 — no compensating cleanup between a landed write and its restore — **I verified this**

In `valueRestoreCycle` the write lands at `:1837`, the readback that can throw is at `:1846`, and the restore does not begin until `:1853`. The only `defer`s in that function are the two `nextID` counters. There is no defer that restores the value.

Scenario: `mixer.set_volume` succeeds, then `logic://tracks` times out. The `try` exits before the restore; the outer catch at `:1237` records a refusal and the sweep continues. The operator's volume stays changed. The transport, selection and boolean-toggle cycles have the same mutate → read → restore ordering.

### BLOCKER 4 — a Korean project name can be mistaken for the Key Commands window — **I verified the mechanism**

`AXLocalePolicy.swift:1040` adds the variant `키 명령` to `keyCommandsWindowTitle`, and `ArmKeyCommandSetup.swift:851` matches it with `mode: .contains` against every Logic window title. A project named so that its arrange-window title contains `키 명령` matches. The reviewer traces the consequence to `:493` → `:468`, where a setup that then finds no `AXSearchField` presses that window's close control.

The label's own rationale already says the head is matched "because the preset name and the edited marker vary", and the call site's comment states the case-insensitivity widening deliberately — but neither considers a *document* title carrying the fragment. The added negative test (`ArmKeyCommandSetupTests.swift:1186`) only uses titles with no known token, so it cannot see this collision.

### BLOCKER 5 — CI no longer runs the live Formula-hash gate — **already found independently, and already fixed elsewhere**

This one is **not this branch's regression** and the reviewer could not have known that from the diff it was given. `1fc80e79` deleted the whole `roadmap` CI job on `main`, which is what invoked `Scripts/ci-verify-formula-sha.sh`; the script was never deleted and is still in the tree with no caller. `run-repo-guards.py` discovers only `check-*.py` and `test_*.py`, and the discovered `test_ci_verify_formula_sha.py` uses local fixtures rather than the live release comparison.

I filed this as #887 before the review returned and #888 restores the job. **Do not fix it on this branch** — it would collide with #888. The finding stands; its owner is different.

### P1 6 — the new AX census turns an unreadable tree into a successful empty measurement

`Scripts/livekit/ax_sheet_census.swift:28` maps every AX error to `nil`, `:35` turns a failed child read into `[]`, and `:74` turns an unreadable window list into `[]`. With Logic running but Accessibility permission absent, the script emits valid JSON with `"windows":[]` and exits 0 (`:87`). A failed census then reads exactly like a measured absence — the failure class this repository names most often.

### P1 7 — a new Phase-B assertion is vacuous when every recipe disappears

`QualificationRunnerTests.swift:1743` calls `withEvidence.allSatisfy` without requiring `withEvidence` to be non-empty, and `:1752` accepts every recipe-less mutating operation as `.notQualified`. Delete every Phase-B recipe and the test still passes: the `allSatisfy` succeeds vacuously and the non-empty credit check at `:1887` can be satisfied by read-only operations alone. Separately, the whole real-process test is skipped when `.build/release/LogicProMCP` is absent (`:1639`), and the CI `test` job only runs `swift test` while the release build happens on a different runner.

### Phase C

Nothing reaches either Phase-C track cycle: both call sites (`:1122`, `:1135`) are gated by the `false` constant at `:3625`, as intended. The dormant code can still leak if it is flipped — `trackStagedDeleteCycle` sets `stagedNeedsRemoval = false` at `:3410` *before* creating the replacement at `:3411`, so a create that lands and then times out throws before `:3419` re-enables cleanup and the defer at `:3336` skips the deletion. The initial create cycle likewise invokes its create at `:3126` before installing its cleanup defer at `:3144`. That is worth fixing before the flag is ever flipped, not after.

### What held, checked rather than assumed

- The German variants introduce no second collision: `Erzeugen`, `Neue Spur`, `Importieren` and `Nein` are exact-matched, `Spuren` is a title suffix, and only `Tempo-Informationen` uses containment. Prefix and containment matching is case-insensitive but diacritic-SENSITIVE (`AXLocalePolicy.swift:95`), so accent folding does not widen them.
- `ChannelRouter` preserves the answering channel's state and only adds refusal metadata; the changed track-create path requires a count delta and a settled modal before State A (`AccessibilityChannel+Tracks.swift:2117`). No new success-without-readback and no new positional targeting in those two files.
- A timeout of the operation's own probe does fail, via `failureReason` at `:260`. Blocker 1 is specifically the recipe/generic mismatch plus the `.none` freshness bypass, not timeouts in general.
- No new dead `#expect(x == true)` / `?? false` / `== .some(true)` assertions. Finding 7 is ordinary logical vacuity, not the macro defect.

### One coverage discrepancy in the review itself

The reviewer lists `.github/workflows/release.yml` among the files it opened, while its own text says that file is not in the scope list and has no diff from `a5ea8674`. Both statements cannot describe the same thing as in-scope work. It does not affect any finding above, but it is the kind of claim that should not be inherited silently.

MongLong0214:
### Merge note: this branch and #888 do not conflict textually, but the merged tree fails three guards

Trial-merged this branch into `fix/887-restore-collateral-guards` in a throwaway worktree. `ci.yml` and `docs/roadmap/README.md` auto-merge with **zero conflicts**. The merged tree then fails `check-observation-records.py`, `check-observations-cover-live-walls.py` and `check-observation-ratchets.py`, and the reason is the same one in all three.

#888 restores a guard that requires `issues` entries in an observation record to be integers, because the reader that consumes them does `int(n)`. Every record written before 2026-09-13 uses `306`. Records written after that date use `"#306"` — the convention drifted the day its reader was deleted.

**It is still drifting.** Twelve records on this branch carry the string form, the newest dated 2026-09-15:

```
2026-09-14-a-korean-input-source-silently-breaks-every-key-command-operation      ['#373']
2026-09-14-arm-named-its-own-precondition-and-the-router-threw-it-away            ['#373']
2026-09-14-one-resource-names-its-provenance-differently-…                        ['#373', '#284']
2026-09-14-phase-c-live-needs-a-logic-with-no-open-document                       ['#373', '#284']
2026-09-14-the-arm-key-setup-could-not-see-a-window-that-was-open-in-front-of-it  ['#373']
2026-09-14-the-control-surface-channels-are-not-set-up-…                          ['#373']
2026-09-14-the-freshness-gate-asked-the-mixer-for-a-field-it-does-not-publish     ['#373', '#284']
2026-09-14-the-marker-list-settles-in-seconds-…                                   ['#373']
2026-09-14-the-routing-graph-publishes-nothing-or-23-nodes-…                      ['#291']
2026-09-14-two-menu-items-one-name-and-every-ax-attribute-identical               ['#373']
2026-09-15-an-auxiliary-window-steals-the-track-menu-…                            ['#373', '#284']
2026-09-15-phase-c-live-runs-and-four-destructive-operations-land-…               ['#373', '#284']
```

Three days of a convention decaying with nothing to say so, which is the argument #887 makes, measured on this branch rather than asserted.

**Nothing is required of this branch before it merges** — the guard is not in the tree here, so CI is not wrong to be green about it. Whichever of the two lands second normalises these twelve (only the numeral changes; no other byte of a record moves) and re-runs `check-observation-ratchets.py`, whose ceiling will need to absorb this branch's new records and label variants the same way #888's does for the earlier ones.

Raising it now so it is a known step rather than a surprise red on `main`.

MongLong0214:
### Status: the lockfile blocker is closed; the five review findings are not. The branch is unchanged at `fc74d373`.

**CI is green on this head** — `compile`, `test` and `build` all succeeded, and the step that had been red since this branch was opened now reads `Verify Package.resolved is unchanged: success`. That is the gate confirming the manifest/lockfile pairing, not an argument about it.

**It is not merged and should not be**, for the reason this repository already writes down: green CI plus a self-review is not the bar. Four of the five blockers in the review above were re-read at this exact head and confirmed, and two of them (B2 and B3) can leave the operator's own project changed.

**An attempt at fixing them was made and abandoned honestly.** A worker was given the five findings with the decisions pre-made and got through a first pass — 887 lines across five files — before the account hit its usage limit (retry 2026-09-19). It never reached its own verification step. Measured afterwards rather than assumed:

```
swift build --build-tests   succeeds
swift test                  ABORTS — libc++abi, uncaught NSException, in
                            QualificationSubprocessSession.startReaders -> readAvailable -> NSFileHandle
```

It was mid-way through writing a mock MCP-server fixture into `QualificationRunnerTests.swift` when it stopped. So the patch compiles and destroys the suite, which is the worst shape to leave anywhere near a branch somebody might push. It is preserved on a local `wip/882-blocker-fixes-partial` and **this branch is untouched at `fc74d373`**.

Two things in it look right on reading, and are recorded so the next attempt does not start from zero: the volume/pan cycle reads the row's `track_ref`, refuses rather than falling back to an index, AND re-checks that the reference did not change between readings — stronger than what was asked for. The Key Commands matcher pairs the title containment with the absence of `AXDocument`. Everything else has plausible shape and no evidence, which is not the same as being done.

**One finding is not this branch's to fix.** BLOCKER 5 (CI no longer runs the live Formula-hash gate) is inherited from `main`: `1fc80e79` deleted the whole `roadmap` job, and `ci-verify-formula-sha.sh` has been sitting in the tree with no caller since. #888 restores it. Fixing it here would collide.

**A merge note that has not changed**: this branch and #888 auto-merge with zero conflicts, but the merged tree fails three guards, because twelve observation records here still write `"#373"` where every record before 2026-09-13 writes `373`. Whichever lands second normalises those twelve.

MongLong0214:
## A blind review of the fix found six more, and two of them are my own false statements

Both external review providers are unavailable — codex is rate-limited until 2026-09-19 and grok's balance is exhausted (HTTP 402). So the fix at `28b1b368` went to a BLIND reviewer instead: a fresh agent given the worktree, the diff, and the four claims to break, with no access to this thread, the commit messages, or any of my reasoning. This repository's own record says blind packets beat vendor diversity, and that held today.

**It is same-provider.** That is a real gap against the cross-provider bar and it is stated rather than papered over.

Every finding below I then checked myself in the code. All six are real.

### F1 — I named the wrong operations, and the comment asserting it is in three places

I wrote that `mixer.set_volume` and `mixer.set_pan` are `VerificationPolicy.none`, so `readbackFreshness` short-circuits for them. **They are not.** `OperationRegistry.swift` hardcodes `verification: .readbackRequired` for the whole `.logicMixer` block; the `.none` sitting beside those rows is the **ConfirmationPolicy** — the tuple is `(OperationID, String, ConfirmationPolicy, TargetPolicy, Set<String>)`. I read the third element as the wrong axis and then wrote the mistake into two source comments and a test-suite docstring as the premise for the whole check.

I also wrote "the only two operations this record is built for". Fourteen cycles in that file produce a `QualificationMutationRestoreRecord`.

### F2 — and the check was inert exactly where it was the only defence

The operations where freshness really is vacuous are `transport.toggle_cycle` and `transport.set_tempo`, both `.none`. Those read `logic://transport/state`, whose envelope carries neither `readable` nor `verified_empty` — it says the same thing as `unverified: true` beside `source: "cache"`. So `readingThatDidNotHappen` could only ever fire on `ax_occluded` for that family: a promotion resting on three readings the resource itself labels unverified would have passed.

`unverified` is checked now. Removing that one line turns the new case red.

### F4 — my own catch had the hole it was written to close

`catch let error as QualificationTransportError`. `observedTrackValue` calls `try JSONSerialization.jsonObject(...)`, which raises an `NSError` on a malformed `logic://tracks` body, and that is reached from BOTH post-write readbacks. A landed write could exit the function with no compensation at all, and the sweep's untyped catch recorded a refusal carrying no note. The comment above it asserted a universal the catch was narrower than. It catches everything now.

### F5 — compensating a write that never happened can move a fader the product left alone

`failing()` compensated unconditionally, including on the `phase_b.mutation` refusal path. `mutationSaysItFailed` fires on `state == "C"`, which for these operations is the fail-closed refusal carrying `write_attempted: false` — nothing was written. The compensating write then drives the fader to `pre.value`, read from the poller-backed cache. If the operator has moved that fader by hand since the last poll, the mixer ends up somewhere neither the product nor the operator put it, on a path where doing nothing was correct.

The refusal's own `write_attempted` field is consulted now, and it is the only thing treated as authoritative that nothing moved.

### F6 — "sent" was being reported as "landed"

`compensateValue` returned a success sentence for anything that was not State C. State B is accepted-but-unverified and is not State C, so a compensating write the server took but that moved nothing read identically to one that worked. It re-reads now and answers `LANDED` / `SENT BUT NOT LANDED` / the reference changed / the confirming read failed.

### F8 — the new AXDocument discriminator was fail-open toward closing a project window

`AXHelpers.getAttribute` collapses a failed read into `nil`, and `nil` is the answer that means "this is the Key Commands window". An AXDocument read that times out on a busy Logic — which is the state right after Option+K — selected the operator's arrange window, and on the no-search-field path cleanup presses that window's close control. The status-preserving seam (`getAttributeResult`) existed and is wired in production; I had not used it.

It is used now, a read failure refuses the window, and the fixture gained the ability to make that read fail so the case can actually be driven. Reverting the fix turns it red.

### Two overclaims of mine, also corrected

- The value-cycle positive control proves **a value cycle ran**, not that it addressed the track by `track_ref` or that a failed cycle compensated. Revert the targeting and delete the compensation and it still passes. The comment said otherwise; it now says what it controls for and names what has no regression control below the live gate.
- `readback` / `restore_readback` were compared with `contains`, and `"restore_readback"` contains `"readback"` — so a misattributed middle reading would have passed. `hasPrefix` now.
- A comment 100 lines from my new assertion still said "a mutating operation cannot reach `passed` at all today". My own `#expect(!creditedMutating.isEmpty)` passes on a live run, and credit requires `.passed`. The comment was stale; corrected.

### Not fixed, reported

F3 (`readingThatDidNotHappen` treats `{"data":null}` as data-present while `QualificationReadbackFreshness.isEmpty` treats it as empty — two authorities answering the same question differently), F7 (a `trk_` reference encodes index + name, so two identically-named tracks swapping rows are not distinguished by the mid-cycle identity check — the server-side `liveTrackIdentityGuard` is what stops the write, and it has its own residual), F9 and F10 (the census's top-level `AXWindows` path exits on `.noValue`, the inverse of the rule the descendant walk applies 30 lines earlier; and only the AXChildren axis is instrumented, so `bool(window, "AXModal") ?? false` still publishes a failed read as `modal: false`).

MongLong0214:
## What this branch needs after #894, measured rather than guessed

#894 makes every new observation record carry its relationship to Logic's own data. This branch adds **13 records**, and the waiver list may only shrink, so none can be exempted.

That would have been a problem. Several of these state facts about Logic's **behaviour**, not about its **strings** — "the routing graph publishes nothing or 23 nodes depending on read order" has no key to cite and nothing to prove absent, and forcing a citation there produces a perfunctory one. So schema 3 is gaining a third declaration, `canon_not_applicable`, bounded mechanically: **a record may say the axis does not apply only if none of its readings resolves in the corpus.** Quote a string Logic ships and a citation was available, so the declaration is false.

Running that bound over all 13 records of this branch, against Logic 12.3 (6674):

**Five need citations** — their readings quote strings Logic ships:

```
25 citable   2026-09-15-phase-c-live-runs-and-four-destructive-operations-land
 8 citable   2026-09-14-the-arm-key-setup-could-not-see-a-window-that-was-open-in-front-of-it
 5 citable   2026-09-15-german-logic-blocks-in-the-new-track-sheet-...   'Neue Spur', 'Erzeugen'
 2 citable   2026-09-14-a-korean-input-source-silently-breaks-every-key-command-operation
 2 citable   2026-09-14-two-menu-items-one-name-...   '컨트롤러 할당…', '모든 컨트롤 서피스 바이패스'
```

**Eight may decline** — behaviour, read order, settle time, window focus:

```
arm-named-its-own-precondition-and-the-router-threw-it-away
one-resource-names-its-provenance-differently-...
phase-c-live-needs-a-logic-with-no-open-document
the-control-surface-channels-are-not-set-up-so-six-operations-send-into-nothing
the-freshness-gate-asked-the-mixer-for-a-field-it-does-not-publish
the-marker-list-settles-in-seconds-and-a-poll-that-does-not-wait-reads-a-stale-answer
the-routing-graph-publishes-nothing-or-23-nodes-depending-on-read-order
an-auxiliary-window-steals-the-track-menu-and-that-is-the-sweep-only-failure
```

The split is **derived, not chosen.**

## The rest is smaller than it looks, and one part of it is confirmed rather than corrected

The `locales:` field this branch adds is validated by the corpus: `Toggle Track Record Enable` and `트랙 녹음 활성화 토글` are Logic's own value under **one key**, so the mechanism is right and the value was not a guess.

Five strings are uncitable, and expectedly:

```
Learn by Key Label / Learn by Key Position / Learn New Assignment
    English lives in KeyCommands.nib, not in a .strings overlay — Apple's Base
    internationalization, filed as #895. Their Korean IS citable and should be cited.
Logic Pro Virtual In / Out
    CoreMIDI port names this product creates at runtime. No Logic asset could carry them.
```

Prefer **semantic keys** over Interface Builder object ids when both carry a value: 9.4% of the corpus is `300557.title`-shaped, and those are renumbered when a nib is rebuilt, so such a citation can break without Apple changing the string. Of the values this branch needs, nine of ten have a semantic alternative.

## Logic canon

```text
logic-canon://strings/Contents%2FFrameworks%2FLogic.framework%2FVersions%2FA%2FResources%2FLocalizable.strings/ko/Toggle%20Track%20Record%20Enable#value
  value:    트랙 녹음 활성화 토글
  used for: the `locales:` value this branch types by hand, shown to be Logic's own
```

MongLong0214:
## Canon prep, measured on Logic 12.3 build 6674

After #894, this branch's 13 observation records are all schema 1–2 and rule 6 refuses a new record below schema 3. Rule 13 then decides, mechanically, which of them may decline the axis and which must cite. Running it against each record:

```
must cite   5     their readings quote a string the corpus holds
may decline 8     they measure timing, ordering and plumbing, not a string Logic ships
```

That split is derived, not chosen. The eight are the routing-graph, freshness-gate, marker-list-settling, arm-precondition, provenance-naming, phase-C-preconditions, control-surface-channels and auxiliary-window records.

## The five, with the key and why that key

| reading | reference | why this key |
|---|---|---|
| `Smart Controls` | `Localizable.strings/ko/Smart Controls#value` | see below — the value in **ko** is `Smart Controls` |
| `트랙 녹음 활성화 토글` | `Localizable.strings/ko/Toggle Track Record Enable#value` | the key is the command's English name; `AXLocalePolicy` already carries that as canonical |
| `Neue Spur` | `Localizable.strings/de/New Track#value` | five candidates; the other four are take handling and bounce dialogs. This key gives `New Track` in en and sits beside `Create` and `Cancel` |
| `Absolute Zero` | `Item.strings/…/InsP/Absolute Zero#value` | a plug-in preset name in the content database |
| `컨트롤러 할당…` | **two candidates, unresolved** | `Localizable.strings/ko/Controller Assignments…` or `PrefsCS.strings/ko/100246.title`. A menu item argues for the first; the record is about two menu items sharing a name, so whoever writes it should confirm against the reading |

## `Smart Controls` turned out to be the interesting one

The record lists it among the toolbar customisation checkboxes, the only English entry in a Korean list (`녹음`, `재생`, `대치`, `솔로`, `카운트 인`, `라이브러리`, `인스펙터`, `빠른 도움말`, `도구 막대`, `Smart Controls`, `믹서`).

Every one of the others is in `TransportCustomization.strings` under ko:

```
녹음 100314.title   재생 100312.title   대치 100298.title   솔로 100318.title
카운트 인 100319.title   라이브러리 100333.title   인스펙터 100334.title
빠른 도움말 100336.title   도구 막대 100335.title   믹서 100328.title

Smart Controls — not in that table at all
```

It is in `Logic.framework/Localizable.strings`, and **its ko value is `Smart Controls`**, in six locales that all keep the English:

```
key `Smart Controls`   de en es fr ko pt
```

So this is not a missing translation. Apple ships the English deliberately in those six. If the record reads it as a localisation gap, that reading should change — the citation makes the correct explanation checkable.

Two other things fell out of the same look, both of which are #895:
- `TransportCustomization.strings` has **zero** `en` entries. Apple compiles the English into the nib and gives only translated locales a `.strings` overlay.
- `Learn by Key Label` is absent from all 23 corpora while `키 레이블로 학습` resolves at `KeyCommands.strings/ko/300557.title`. An author citing the Korean succeeds and one citing the English is asked to prove uncitable a string Logic ships.

`Scripts/logic_canon.py locate '<string>'` (#899) prints these candidates grouped by key, and says outright when there is more than one, because choosing between them is judgement the tool does not have.

MongLong0214:
Superseded by #926, merged as `5519079e`.

The qualification subsystem landed there with the three blockers this PR carried fixed at the root rather than rebased past:

- **The evidence could not pass its own verifier.** A mutating pass reaches `.passed` only through a probe that was REFUSED (`isError == true`), and the shape rule it was filed under demands `operationIsError == false`. `verified_write_cycle` is the third kind, bound to its mutation-restore record by digest. No test had ever put such a record through verification — `mutationRestore:` appeared zero times in `Tests/` — which is why nobody noticed.
- **`--qualify` wrote to the operator's open project and said so nowhere.** `--mutate-open-project` is now required and `--help` states the writes; the gate is on the spec list, not on the twelve recipe sites.
- **A failed run exited 0.**

The gate steps are deliberately NOT in #926. All seven were `if: vars.ADR001_QUALIFICATION_ENFORCED == 'true'` and that variable is set nowhere, so restoring them changes what a release does by nothing; deleting the `if:` would stop every tag in `build-release` for want of a secret that does not exist. And the pinned verifier at `67fec5f` answers `"semantic readback binding"` for every case except `system.health`, so turning the gate on today would reject all 21 Phase-A passes. Wiring it is its own change with a run URL as its acceptance.

The product half of this branch landed earlier as #922, #924 and #925.

PR #917, opened 2026-09-18
fix(release): the registry publish has never fired, and the record it would have published linked to the wrong release

MongLong0214:
The MCP registry advertises **3.7.1**, published 2026-06-23. This repository has shipped through 3.16.0.

```
$ curl -s 'https://registry.modelcontextprotocol.io/v0/servers?search=io.github.MongLong0214/logic-pro-mcp&version=latest'
version    : 3.7.1
release URL: https://github.com/MongLong0214/logic-pro-mcp/releases/tag/v3.7.1
isLatest   : True   published: 2026-06-23T04:19:11.605537Z
```

Three defects, each of which alone would have been enough.

## 1. The trigger has never fired

`publish-mcp.yml` ran on `release: [published]`. Its entire run history:

```
$ gh run list --workflow publish-mcp.yml
2026-06-23 workflow_dispatch success
2026-06-23 workflow_dispatch success
2026-06-23 workflow_dispatch failure
```

Three runs, all manual, all on the day the workflow was added. The reason is in the release list:

```
$ gh api 'repos/.../releases' -q '.[]|"\(.tag_name) by=\(.author.login)"'
v3.16.0 by=github-actions[bot]
v3.15.0 by=github-actions[bot]
...
v3.7.4  by=github-actions[bot]     <- first bot-created release, 2026-06-30
```

`release.yml` creates the release with softprops/action-gh-release and the default `GITHUB_TOKEN`, and **GitHub does not start workflow runs from events a `GITHUB_TOKEN` created**. v3.7.4 is the first release after the last manual dispatch, and the registry stops exactly there.

**Fix:** `release.yml` now calls `publish-mcp.yml` as a reusable workflow, `needs: validate-install` — i.e. after the published assets have been downloaded and run through the installer on macos-14 and macos-15. The registry record names a release, so it goes out downstream of the check that the release works. `release: [published]` stays as a fallback for a human-published release (those do trigger) and `workflow_dispatch` stays as the manual recovery path.

## 2. The record would have linked to the previous release

The sync step was one line of jq:

```
jq --arg version "$VERSION" '.version = $version' server.json
```

`server.json` carries the version in **two** places — `.version` and `distribution.release`. That line writes the first. So a publish of version N would have advertised N and sent every reader to the download page for N−1. Nothing could catch it: the rewrite happened on a runner, against a file the workflow published and threw away, so the wrong record would have existed only in the registry.

**Fix:** `Scripts/sync-server-json-version.py` writes both fields or writes neither, and has `test_sync_server_json_version.py` beside it. It also refuses a version that is not strict SemVer — the version reaches it from a tag name and the job holds `id-token: write` — and refuses a document whose `distribution.release` is missing or is not a release URL, rather than half-writing it.

## 3. The committed file had drifted on its own

```
tag        server.json  distribution.release  Formula
v3.14.0    3.14.0       v3.14.0               3.14.0
v3.15.0    3.15.0       v3.15.0               3.15.0
v3.16.0    3.15.0       v3.15.0               3.16.0   <- nobody noticed
```

v3.14.0 and v3.15.0 were right because somebody remembered. That is the whole mechanism.

**Fix:** `Scripts/check-registry-metadata-matches-the-release.py` pins both fields to the Homebrew Formula's version, which `ci-verify-formula-sha.sh` already compares against the tarball attached to that release. The record is then chained to a verified artifact — release → Formula (hash verified) → registry record. Both fields are checked separately, because a guard that reads one of two copies is how the other one drifts.

`server.json` is corrected to 3.16.0 in this PR, by running the sync script rather than by hand.

## Verification

```
python3 Scripts/run-repo-guards.py                 exit 0, all 65 passed (was 62)
actionlint                                         exit 0, all workflows
swift test --filter 'ReleasePipelineContract|InstallScriptContract|VersionConsistency'
                                                   46 tests passed
```

Both new guards are `Scripts/check-*.py` / `Scripts/test_*.py`, so `run-repo-guards.py` discovers them and the `guards` CI job runs them with no change to `ci.yml`.

### Mutations

The guard — each produced real failures, not crashes:

```
version check removed            3 FAIL
release-URL check removed        5 FAIL
tag compared by prefix only      2 FAIL   (v3.16.0-rc.1 accepted as v3.16.0)
CANNOT-DETERMINE removed         2 FAIL
```

The sync script:

```
writes .version only             5 FAIL   <- the original defect
SemVer check removed            14 FAIL
half-write with no release URL   2 FAIL
```

The first attempt at that last one **exited 1 with zero FAIL lines** — the mutation had made the script crash, so it proved nothing about whether the test can see a half-write. It was rewritten to genuinely half-write before it counted.

## NOT RUN

- **The registry is still nine releases stale.** Nothing here republishes it. That is a `workflow_dispatch` against the registry, and publishing is not mine to trigger — the committed record now names 3.16.0, so a dispatch would correct it.
- **The reusable-workflow call has not been executed.** `release.yml` fires only on a tag push, and no tag was pushed. It is verified by `actionlint` (which does check context availability and reusable-workflow calls) and by YAML parsing — not by a run.
- `validate-install` now gates the registry publish, but a registry publish cannot be un-published. If it fails after `validate-install` passes, the release stands and the record does not.

## Canon

This change states no fact about Logic. It touches the release pipeline, the MCP registry record and two new guards; every version string in it is read from this repository's own `Formula/logic-pro-mcp.rb`, from GitHub's release and workflow-run APIs, and from the public MCP registry, and none of it is a claim about anything Logic Pro displays, names or ships.


MongLong0214:
Superseded by #918, which carries this commit as its first and eight more from the same four production reviews. The branch here is behind main and the work has moved rather than changed.

PR #932, opened 2026-09-20
fix(guards): a rule that cannot be applied stops the run, instead of noting it beside exit 0

MongLong0214:
## What exited 0 while a rule was off

`check-policy-literals-against-canon.py` classifies every UI literal in the product against the pinned corpus. One classification — `composed_value` — exempts a literal from the `nowhere` ledger without any corpus holding it, so it is the one a wrong guess escapes through. The rule that keeps it honest asks whether anything beyond the decomposition itself says Logic composes the string: an observation record that names it, or a LabelSet that declares the composition.

That rule switched itself off when its evidence could not be read — `docs/observations/`, `LABELSETS-WITHOUT-A-ROW.json`, or either of the two readers loaded by path. It printed a note to stderr, returned `False`, and **the run exited 0 with fifteen literals unchecked**. Recorded as a limit at `014cbb05`: *"The rule abstains when it cannot read its evidence, and says so on stderr. A tree where the readers break silently stops enforcing it — the note is the only thing that would tell anybody."*

## Why the abstention was there, and why it is not needed

It was reasoned, not sloppy: the self-test used to build a tree of `Scripts/` and a classification and nothing else, and the rule's first version turned its fifteen legitimate literals into fifteen findings.

That fixture no longer exists. The self-test copies the whole of `docs/canon`, the whole of `docs/observations` and both readers — *"because a fixture that is a subset of what the guard reads tests nothing"*, its own comment. The only tree that reaches the branch now is one where an artifact this repository ships has **gone**, and that is a broken tree, not a clean one.

It refuses once, naming what is missing and how many literals the switched-off rule covers, rather than once per literal.

## Evidence

- The real tree still passes: `1204 literals … composed_value 15, nibstrings_value 6, nowhere 85, quickhelp_title 217, strings_value 881`.
- Four new cases: a tree carrying everything passes (**the control**), and losing `docs/observations/`, losing `LABELSETS-WITHOUT-A-ROW.json`, or losing `check-labelsets-are-derived.py` each exits 1.
- `Ran 32 tests … OK`; `python3 Scripts/run-repo-guards.py` → `all 71 passed`.

## Ruled out

Keeping the abstention and making the note louder. It was already loud and it was on stderr next to exit 0 — a thing that fails to stop anybody cannot be fixed by wording.

This pull request body states no fact about Logic: it changes when a guard refuses, and the numbers in it count literals classified in this repository's own `POLICY-LITERALS.json` and cases in a self-test.


MongLong0214:

Issue #896, opened 2026-09-15
Twelve python drives are outside run-repo-guards.py's discovery, and only a release script runs any of them


MongLong0214:
## What is true

`Scripts/run-repo-guards.py` discovers `Scripts/check-*.py`, `Scripts/test_*.py` and
`Scripts/livekit/test_*.py`. Twelve files are named `*_test.py` instead, so it finds none of them:

```
live_e2e_bridge_test.py                 logic_session_bootstrap_test.py
logic_bounce_test.py                    logic_session_bootstrap_action_test.py
logic_bounce_main_test.py               logic_ui_jxa_test.py
logic_bounce_support_test.py            logic_free_tempo_modal_test.py
logic_bounce_ui_test.py                 logic_controller_learn_mode_test.py
logic_variants_test.py                  logic_key_event_test.py
```

The only thing that runs them is `Scripts/release-stable.sh`, which no workflow in
`.github/workflows/` invokes. `live_e2e_bridge_test.py` is referenced by nothing at all — not even
by the `py_compile` line in that script.

So a pull request can break any of them and CI stays green. The break surfaces at release time,
which is the worst moment to find it.

## Measured now

All twelve pass, headlessly, without Logic, in under a minute:

```
live_e2e_bridge_test.py                 exit=0  OK
logic_bounce_test.py                    exit=0  OK
logic_variants_test.py                  exit=0  OK
logic_ui_jxa_test.py                    exit=0  OK
logic_session_bootstrap_test.py         exit=0  OK
logic_session_bootstrap_action_test.py  exit=0  OK
logic_free_tempo_modal_test.py          exit=0  OK
logic_controller_learn_mode_test.py     exit=0  OK
logic_key_event_test.py                 exit=0  OK
logic_bounce_main_test.py               exit=0  (no output)
logic_bounce_support_test.py            exit=0  (no output)
logic_bounce_ui_test.py                 exit=0  (no output)
```

That is exactly the criterion the runner's own docstring gives for what belongs in it: *"Both are
plain Python needing neither Xcode nor Logic."*

## The three silent ones are not a widening bug, they are the point

`logic_bounce_main_test.py`, `logic_bounce_support_test.py` and `logic_bounce_ui_test.py` are
`TestCase` modules with no `unittest.main()`. Run directly they exit 0 having asserted nothing;
they are imported by `logic_bounce_test.py`. So **widening the glob to `*_test.py` would add three
[...]

PR #934, opened 2026-09-20
fix(tools): a suite that stops contributing cases now fails, and the count is of distinct cases

MongLong0214:
## Two numbers that were not what they said

`run-helper-suites.py` exists because `run-repo-guards.py`'s globs do not match `*_test.py`, so twelve suites could break with CI green (#896). It drives them.

**It refused only when the total was zero.** One suite losing every case — a class renamed, an import guard that stops being satisfied — lowers a number nobody reads and passes. That is the shape of a guard that checks nothing, sitting inside the drive written to close exactly that gap.

**And the total was not the number of cases.** `logic_bounce_test.py` imports three `TestCase` classes from its neighbours to re-export them, and `unittest` discovery loads a class wherever it finds it. 32 cases were being run — and counted — twice: `167 case(s) pass` over **135** distinct ones.

## What changed

- Cases are grouped by the module that **defines** them, so an aggregator re-exporting a class cannot hide a loss somewhere else.
- `Scripts/helper-suite-cases.json` carries a **floor per suite**. Below it, undeclared, or declared-and-not-found each refuse. Growing is free; lowering one belongs in the diff with the reason.
- An empty or unreadable floor file is **refused**, not passed — otherwise the rule evaporates the moment its input does.
- The suite that runs is the deduplicated one: **135 cases in 30s** instead of 167.

## Evidence

```
135 distinct case(s) pass across 11 *_test.py suite(s) the runner does not discover,
each at or above its committed floor
```

Seven new cases in `Scripts/test_helper_suite_floors.py`, driven against fixture directories so they run anywhere — the real suites compile Swift and import AppKit, which is why the runner does not discover them:

- a tree at its floors passes (**the control**)
- a suite that lost a case → exit 1
- one suite of two vanishing entirely → exit 1 (`discovery found NONE of it`)
- an undeclared suite → exit 1
- an empty floor file → exit 1
- a re-exported class counted **once**, under the module that defines it (`Ran 2 tests`, not 4)
- every real `*_test.py` that defines cases is declared

`python3 Scripts/run-repo-guards.py` → `all 72 passed`.

## Ruled out

Renaming the twelve to `test_*.py` so the runner discovers them — #896's end state. It makes the guards job platform-dependent: `logic_key_event_test.py` compiles Swift that imports AppKit and the guards job runs on ubuntu. That rename needs the job split first.

## Where the floors live

Beside the drive, not in `docs/canon/` with this repository's other ratchets. That prefix is Logic-facing, so a change touching anything under it must cite a row of Apple's data in its pull request body. This file counts test cases and rests on no such row — putting it there would make every future edit manufacture a citation, which is exactly what the citation rule exists to stop.

This pull request body states no fact about Logic: every number in it counts test cases in this repository's own helper suites.


MongLong0214:

PR #935, opened 2026-09-20
fix(canon): the census guard had its own parser, and three LabelSets lived in its blind spot

MongLong0214:
## Three LabelSets in the guard's own blind spot

`check-labelset-census-only-shrinks.py` read the policy with its own parser: a regex for `static let X = LabelSet(`, then `source.find("\n    )")` for the end of the block. A declaration that closes on the **same line as its last argument** is not delimited by that search — so the span the guard examined ran on into a *later* declaration, found **that** one's `derivedFrom:`, and concluded the set named a row.

Measured 2026-09-20: that parser sees **58** undeclared LabelSets. The balanced parser in `check-labelsets-are-derived.py` — which raises rather than skipping — sees **61**. The three in the blind spot are `eventListColumnM`, `eventListColumnName` and `eventListColumnPosition`, each carrying one or two locales and naming no row. A census of 58 called itself complete over them.

## What changed

- The guard reads the policy through the **sibling's parser**. A second copy of a parser is a second copy of the truth, and this one went stale in the direction where the checking is missing.
- The three are in the census with `derive_label_variants.py`'s verdicts — all three `ambiguous` — plus what today's measurement actually found (below).
- **Growth is redefined**, because the old rule made the census unfixable. It was "a name the base census did not list"; repairing a parser blind spot *is* such an addition. It is now "a name that named a row, or did not exist, **at the base**" — read from the base's own policy source. A repair is accepted and said out loud; a genuinely new undeclared LabelSet is still refused.

## What the measurement found

| set | candidates | what they say |
|---|---|---|
| `eventListColumnName` | 18 | **17 share one ten-locale signature**; the eighteenth is MAMixer's `name`, differing only by case in en/es/fr/it. The key is ambiguous, the values are not. |
| `eventListColumnPosition` | 7 | Two signatures differing **in exactly one locale**: ja reads `ポジション` in four, `位置` in three. Offline cannot decide it. |
| `eventListColumnM` | 4 | Four signatures. Two are chord labels from ChordsKit (`Maj`/`maj`). The other two read `M` in English and split in Chinese: MAMixer's reads `静音` — *mute* — in zh_CN, Logic.framework's reads `M`. |

None is closed here. That is the point of the file: a row that resolves is not the row a control is named by.

## Canon

The census entry this change adds names, for contrast, the two Event List columns that DO resolve to a row — one row, ten locales, nothing typed. That is what the three cannot do yet:

```
logic-canon://strings/Contents%2FFrameworks%2FLogic.framework%2FVersions%2FA%2FResources%2FLocalizable.strings/en/Trk#value
Trk
```

## Evidence

- Five new cases: a declaration closing on one line is **seen** (and named), an unreadable declaration **stops the run**, a census short of a real entry is refused (and names it), growth-as-repair is accepted and said, and a **new** set filed in the census is refused.
- The refusing direction needed base **content** seams (`LPM_LABELSET_BASE_POLICY`, `LPM_LABELSET_BASE_JSON`): `git show <base>:<path>` can only reach a path the base carries, so a fixture census degrades to the bootstrap note and that direction had never been driven.
- `python3 Scripts/run-repo-guards.py` → `all 71 passed`.


MongLong0214:

PR #936, opened 2026-09-20
fix(guards): the label rule stopped at four characters, so `OK` and every three-letter label walked through

MongLong0214:
## A blind spot two characters wide

`check-shipped-python-has-no-ui-literals.py` refuses a shipped Python helper that spells one of Logic's own labels — the rule that exists because `logic_bounce_ui.py` carried six hand-typed English-and-Korean tables in a file the installer ships.

It skipped every literal shorter than four characters, and its own comment named the cost:

> Four is where `OK` and `Off` fall out and `Save` stays in.

That is reasoning, not counting. A three-letter label in any language walked straight through, in the direction that hides a finding.

## The measurement

Run the whole rule at **2** instead of 4, across every shipped helper: **exactly one** literal of length two or three is a value Apple ships anywhere. It is `...` — the ASCII ellipsis `logic_variants.py` appends when it truncates an `osascript` error. A format token, in an expression that says so, which is what `PROTOCOL_LITERALS` is for.

So: the floor is **2**, and the marker is exempt **in the expression that builds it** (`detail[:237] +`), not on the line and not everywhere. One exemption bought a two-character blind spot.

## Evidence

Four new cases (17 total):

- a two-character label Apple ships (`OK`) → exit 1
- a two-character string Apple ships nowhere (`qz`) → exit 0 — **the control**, passing for the right reason
- the truncation marker in the expression that builds it → exit 0
- the same literal elsewhere (`LABEL = "..."`) → exit 1

`python3 Scripts/run-repo-guards.py` → `all 71 passed`.

## Ruled out

Lowering to 3 rather than 2. The only literal the measurement found at either length is `...`, which *is* three characters — stopping at 3 would need the same exemption and still keep `OK` and `Off` out, which is the hole being closed.

## What is still skipped

One character. `M`, `L`, `R` in a column header collide with too much to separate by value alone, and the measurement was not run at 1. That is recorded rather than closed.

## Canon

The case above rests on `OK` being a value Logic ships — that is why a two-character literal spelling it is a finding, and the new case in `Scripts/test_shipped_python_has_no_ui_literals.py` writes exactly that string:

```
logic-canon://strings/Contents%2FFrameworks%2FLogic.framework%2FVersions%2FA%2FResources%2FLocalizable.strings/en/OK#value
OK
```

`qz`, the control, is absent from every corpus — which is what makes it a control and not a second copy of the same case.


MongLong0214:

Issue #937, opened 2026-09-20
The canon citation rule covers four files that state no fact about Logic, so correcting them needs a manufactured citation


MongLong0214:
## What is wrong

`docs/canon/LOGIC-FACING.json` lists path **prefixes** whose contents are claims about Logic, and `docs/canon/` is one of them. A change touching any path under it may not use the `states no fact about Logic` opt-out: its pull request body must carry a canonical citation that **bears on a changed file**.

Four files under that prefix carry no claim about Logic at all. They are repository tooling:

| file | what it holds |
|---|---|
| `docs/canon/CI-GATE.json` | which jobs and commands the required CI gate must carry |
| `docs/canon/CI-SKIPS.json` | how many cases each guard may SKIP under CI |
| `docs/canon/GUARD-TESTS-BLIND-TO-THEIR-GUARD.json` | guards whose tests do not notice their gate being removed |
| `docs/canon/PROSE-NUMBERS.json` | numbers `README.md` may state with no artifact behind them |

A change to one of these has two honest options and neither works: the opt-out is barred because of the prefix, and there is no row of Apple's data the change rests on. The third option — pasting a citation that resolves and quoting its value somewhere in the diff — is **manufacturing evidence**, which is the exact failure the citation rule exists to prevent.

## It is not hypothetical

Measured 2026-09-20, twice in one afternoon:

1. A per-suite floor file for `run-helper-suites.py` was written to `docs/canon/HELPER-SUITE-CASES.json`, beside the other ratchets. `check-canon-citations.py --text` refused the pull request body: *"this change may not opt out: it edits 1 file(s) whose contents are claims about Logic."* The file counts test cases. It was moved to `Scripts/helper-suite-cases.json` — away from the convention — for no reason other than this rule.

2. `GUARD-TESTS-BLIND-TO-THEIR-GUARD.json` now carries two stale statements: `how_to_reproduce` says `--fast`, which after #933 explicitly does **not** sweep the two slow guards, and `not_swept` names them as unswept when a full sweep covers them (measured today: `29 of 29 guard(s) have a test that notices the gate being removed`, both slow guards included). The correction is four lines and is not being made, because making it requires a citation that does not exist.

## Two ways out

[...]

Issue #910, opened 2026-09-17
A composed LabelSet has no row, so nothing offline checks the composition


MongLong0214:
## The gap

`Scripts/check-labelsets-are-derived.py` verifies a LabelSet against ONE row: for each of the ten locales, some member must be the value Apple pins there. A COMPOSED label has no row.

`showLibraryMenuItem` is the first one. `Show Library` is in no corpus in any locale, because Apple ships `Show %@` and `Hide %@` as templates and Logic assembles the item — measured live on a Korean Logic 12.3, whose View menu reads `라이브러리 가리기`. Its members are the Show template composed with the Library noun in each locale, and **nothing offline checks that composition**.

Two checks each see half of it:

| check | what it can say |
|---|---|
| `check-policy-literals-against-canon.py` | that Apple composes a string like this one, via `docs/canon/TEMPLATES.json` |
| `check-labelsets-are-derived.py` | nothing — there is no row to verify against |

So a typo in a composed variant is caught only if it stops being composable at all. `Afficher Bibliothèqu` would still decompose to a noun that is not a corpus value and be refused; `Afficher Bibliothèque` paired with the WRONG noun for that locale would not.

## What to do

- [ ] A way for a LabelSet to name a composition — the template row and the noun row — as `derivedFrom` names one row.
- [ ] `logic_canon.py build` pins what the check needs. Digests of the composed strings are enough and keep the artifact small; the plaintext already lives in `TEMPLATES.json` for the templates a classification rests on.
- [ ] The guard then requires, per locale, that some member IS the composition — the same shape the row check already has.
- [ ] Self-test by mutation: swapping one locale's noun for another's must fail.

## Size

One LabelSet today. Filed rather than built because a mechanism for one label is how a mechanism gets built wrong; the second composed label is the one that should pay for it, and `%@ contents`, `%@ header` and `on Track %@` are already in `TEMPLATES.json`, so it will not be long.

## Logic canon

This issue states no fact about Logic that is not measured by `docs/observations/2026-09-18-the-plugin-menu-is-english-on-a-korean-logic.json`.

MongLong0214:
Done on `main` (`ac67c770`), landed by **#915** (`6aa4add2 feat(#910): the composition nothing checked is two rows, and now both are`). Checked against the four boxes this issue wrote, not against the commit title:

[...]

Issue #909, opened 2026-09-17
Region bar positions are parsed by four hand-written regexes; Apple ships the sentence in ten locales


MongLong0214:
## What is hand-written

`AccessibilityChannel+Regions.swift` parses a region's bar positions out of its AXDescription with
four regexes, one per language somebody read off a running Logic:

```
리전은\s*(\d+)\s*마디.*?시작.*?(\d+)\s*마디.*?끝
(?i)region\s+starts\s+at\s+(?:bar\s+)?(\d+)(?:\s*bars?)?.*?ends\s+at\s+(?:bar\s+)?(\d+)(?:\s*bars?)?
リージョンの開始位置は\s*(\d+).*?終了位置は\s*(\d+)
Region\s+beginnt\s+bei\s*(\d+).*?endet\s+bei\s*(\d+)
```

Four of ten. On the other six, `enumerateRegions` returns `startBar: -1, endBar: -1` — the comment
above those patterns records exactly that failure for Japanese before its row was added.

## Apple ships the sentence

One row, `Localizable.strings` key `Region starts at %@ and ends at %@`, carries all ten locales:

```
en     Region starts at %@ and ends at %@
ko     리전은 %@에서 시작하여 %@에서 끝납니다.
ja     リージョンの開始位置は%@、終了位置は%@です
de     Region beginnt bei %@ und endet bei %@
es     El pasaje comienza en %@ y termina en %@
fr     La région commence à %@ et se termine à %@
it     Inizio regione: %@; fine regione: %@
pt     A região começa em %@ e termina em %@
zh_CN  片段开始于 %@，结束于 %@
zh_TW  區段起始於 %@，並結束於 %@
```

Each hand-written regex is that locale's sentence with the placeholders turned into capture groups.
The generated form of the Korean row is the same anchors the hand-written one uses — `리전은`, the
first number, then the literal text between the two placeholders.

## What to do

- [ ] Derive the ten sentences from the row rather than carrying four regexes.
- [ ] Build each pattern by escaping the literal segments and replacing each `%@` with a digit
      capture, so a locale added by Apple arrives without anybody writing a regex.
- [ ] Keep the two properties the current comments record and were measured the hard way: the UNIT
      is mixed (`1 bar ` beside `2 小節`) and INFLECTED (`1 Takt` beside `2 Takte`), so the pattern
      must anchor on the sentence's own literal text and never on a unit word.
[...]

Issue #904, opened 2026-09-16
Locale migration SSOT — finish existing bundle-derived labels without claiming unmeasured runtime support


MongLong0214:
# Locale derivation migration — current implementation and remaining work

> **Execution SSOT r2 — 2026-09-21. OPEN / bounded existing migration.** #892 owns parent acceptance, this issue owns implementation, and #308 owns programme selection. This is the existing ADR-019-related continuation, not a new locale/observation framework. Source checkpoint: `06e5d37c8bd8fb9db62431618b0d386595f63ba9`. This revision is not a live qualification of every language or a fresh count of every remaining literal.

## 1. Decision and first remaining deliverable

Use the existing pinned corpus and `LabelSet.derivedFrom` machinery to derive the relevant row's available language values, preserving justified variants and the existing matcher semantics. Do not start a per-label/per-language live measurement campaign for values already established by the pinned bundle. Do not rebuild corpus ingestion, ten-language detection, the derived-value guard or the new-LabelSet gate that already exist.

First deliverable: run the existing derivation/inventory tools against current source, identify the first unresolved **matching site or compound/ambiguous LabelSet**, and migrate that one real consumer with a correct row/composition/absence disposition and a regression test. Publish current remaining identifiers rather than copying the historical 159/168, 112, 48 or 15 counts as today's truth.

## 2. Implementation history that supersedes the original nine-step plan

| Recorded work | Actual result and limit |
|---|---|
| #894 and #903 | Existing corpus/derivation foundations; inspect current output and pinned build rather than rebuilding them from the proposal. |
| [PR #905](https://github.com/MongLong0214/logic-pro-mcp/pull/905), merge `1f5d878de09ab9e23f4f7d6adb375e9f42cecd60` | Ten-language detection, `derivedFrom`, offline digest guards and cross-locale pinning foundation. It explicitly did not complete bulk migration. |
| [PR #906](https://github.com/MongLong0214/logic-pro-mcp/pull/906), merge `e075f1e054ca1acde5cda6ce546755c1d906bc1d`; implementation `d33f97586f5f2f28f03f55c1faa91a53287cb3c5` | Seventy LabelSets were migrated with 426 added strings in that historical diff. Compound/tolerance cases were deliberately not all applied. Cross-source `strings`/`nibstrings` rows and case-folded matching required explicit corrections. |
[...]

PR #939, opened 2026-09-20
feat(locale): the Event List names its columns in code, and the exception mechanism says what it proves

MongLong0214:
Two commits, both from putting the remaining open work in front of a second reader instead of leaving it as a backlog.

## 1. The Event List names its columns in code

`eventListColumnM`, `eventListColumnPosition` and `eventListColumnName` carried one or two locales each and named no Apple row. `derive_label_variants.py` answers **ambiguous** for all three — 4, 7 and 18 rows in the bundle carry those values — and this repository's own warning is that a row whose VALUE matches is not the row a control is NAMED by (`trackTypeMaster`'s best candidate is a Bass Amp label).

The bundle decides it, with no running Logic:

- **No nib mentions `EventList`** — 159 `Base.lproj` nibs scanned, zero hits. The columns are built in code.
- `-[EventListView createTableColumns]` (0x3fa468) titles each header from `-[EventListView columnTitle:forMenu:]` (0x3f8c24), which this reading identifies as looking the column identifier up in **Logic.framework's own `Localizable.strings`** under `M`, `Position` and `Name` -- a reading of the disassembly, not a re-derivation of the callee or the bundle it resolves against; see the observation record's `limits`.

What that buys, read from the pinned corpus:

| set | what the row says |
|---|---|
| `M` | nine locales spell it `M`; **zh_CN spells it 静音 — mute**. That is also what separates this row from MAMixer's `M`, whose zh_CN value stays `M`. |
| `Position` | ja is **ポジション**, not 位置 — three of the seven candidate rows would have given 位置. |
| `Name` | eighteen rows carry `Name` in English and seventeen share one ten-locale spelling, so the values were never in doubt and the row was. |

**137 of 195** LabelSets now name a row, each proved per locale offline. The census falls **61 → 58**. The reading is recorded with its evidence in `docs/observations/2026-09-20-the-event-list-names-its-columns-in-code.json`.

## 2. What the Logic-facing exceptions prove — and three ways the first version did not

The exception mechanism landed in #938 this afternoon. A second reader took it apart:

1. **The file claimed a ratchet it does not have.** `LOGIC-FACING.json` said "THIS LIST MAY ONLY SHRINK" while `exceptions` is deliberately excluded from `RATCHETS`. A sentence that contradicts the code is worse than no sentence.
2. **"quotes nothing citable" could mean "could not check".** `_citable_strings_in` catches `CanonError` and carries on. Reading a PR body that is advisory; as **evidence for an exemption** it is the shape of every silent pass this repository removes. Found by simulating a failure of every lookup — `"Audio Units"` came back clean. It takes `strict` now and rule 15 uses it.
3. **The body path trusted a list it had not checked.** Rule 15 runs in the tree check, and `--text` returns before it. In CI the tree check is a required command and does run — but a rule that is sound only because another step happened has an undeclared dependency. `--text` proves the exceptions itself now (control: put `docs/canon/README.md` in the list and the body check exits 1).
4. **And the limit, which is the real answer to #937.** No check over free prose can establish "states no fact about Logic": appending *"The bounce dialog has three buttons."* to a note passes rule 15, and the alternatives are worse — a corpus word scan flags `deleted`, `merged` and `control` in CI policy explanations. The file now separates what is **proved** (no reference, no citable quote, re-derived every run, unknown refuses) from what a reviewer is **asserting**, and names the one thing that would close it: these files carrying no free prose at all, which means rewriting what `CI-GATE`, `CI-SKIPS` and `PROSE-NUMBERS` store.

A case pins the limit, so a later reader cannot mistake it for a proof — and if somebody strengthens the rule, that case goes red and the sentence has to be rewritten in the same change.

## Evidence

- `check-labelsets-are-derived.py` → `137 of 195 LabelSets name a row, and every one of them is that row's own values in every locale the corpus carries` -- case-folded (`:149`): at least one member matches case-insensitively, not exact identity, and extras beyond the pinned row are allowed by design
- `check-labelset-census-only-shrinks.py` → `the census is complete and gained no new set: 58 LabelSet(s) name no row, of 195`
- `check-observation-records.py` → `128 observation record(s) obey the schema`; the new record carries its evidence file
- `test_canon_citations_guard.py` → `Ran 79 tests … OK`

## Canon

The three rows the Event List's headers are titled from, and the one that separates a mute column from a mixer's:

```
logic-canon://strings/Contents%2FFrameworks%2FLogic.framework%2FVersions%2FA%2FResources%2FLocalizable.strings/en/Name#value
Name
```

```
logic-canon://strings/Contents%2FFrameworks%2FLogic.framework%2FVersions%2FA%2FResources%2FLocalizable.strings/zh_CN/M#value
静音
```

Closes nothing on its own; #937 stays open with its answer written down and #904 counts this as three more.

---

## Added 2026-09-20: the roadmap rows, cherry-picked from #940

`main` is **red** at `ac67c770` — `Scripts/test_labelset_census_only_shrinks.py` fails there, and this branch is what fixes it. #940 (nine roadmap rows) could not merge, because its `guards` job hit that same failure; this branch could not merge, because its `status` job failed on the four issues closed today, which #940 fixed. One green pull request is worth more than two blocked ones, so #940's rows are now here.

The one worth naming: the `#896` row stated as fact that `Scripts/test_helper_suites_are_run.py` "is discovered and runs all twelve through `unittest` discovery, so CI now executes their 167 cases".

```
git cat-file -e ac67c770:Scripts/test_helper_suites_are_run.py   -> ABSENT
git merge-base --is-ancestor 2a5a0268 origin/main                -> NO
git merge-base --is-ancestor eb8c0ff2 origin/main                -> NO
```

That file has never been on `main`. The merge that wrote the row (#920, `014cbb05`) describes it in its commit message and did not land it. What does run those cases is the helper-suite drive in `ci.yml`'s macos-15 `test` job, deliberately **not** discovered by `run-repo-guards.py` — that job runs on ubuntu and one of the twelve compiles Swift that imports AppKit.

Also corrected: **#373/#816** (the subsystem was restored by #926 `5519079e`; what is missing is the *gate* — `ADR001_QUALIFICATION_ENFORCED` is set nowhere and the `if:` needs a `QUALIFICATION_EVIDENCE_URL` secret that does not exist), and **#883** (all three LabelSets carry German now, from `d33f9758`, and all three name an Apple row — but that German is *derived* from Apple's rows, not read off a running German Logic, which is why the issue stays open). **#885, #898, #909, #910 and #937** are marked closed with the evidence each was verified against.

`python3 Scripts/roadmap-table-matches-github.py --pr 939` → 89 rows, 20 open issues all listed, table agrees with GitHub. `python3 Scripts/run-repo-guards.py` → **all 75 passed**.

#940 is now a subset of this branch and will be closed rather than rebased.


MongLong0214:
## Review findings fixed — `07755d77`

An independent review at `ff94093c` returned five claims this branch made that its own code and committed evidence do not support. All five are fixed. Chasing the last one turned up a factual error in the observation record.

### The cfstring was read wrong

`docs/observations/evidence/2026-09-20-event-list-column-titles.json` recorded three addresses and the section they live in, **not their contents**; the record said a co-reader had identified them as `Muted`, `Position` and `Name`.

Read out of the binary the evidence file's own sha256 already pins — 32-byte CFString record, data pointer through the chained-fixup word, into `__TEXT,__cstring`:

```
0x2322ba0  len=1  -> 0x1d54e4a  "M"
0x231fdc0  len=8  -> 0x1d53499  "Position"
0x2316f20  len=4  -> 0x1d4be18  "Name"
```

The first is **`M`**, one character. `Muted` is a *fourth* cfstring at `0x2315880`, referenced once at `0x3f8c74` and held in `x22`; in the `forMenu: NO` branch (reached by `cbz w19, 0x3f8f28`) it is stored at `[sp, #0x128]`, one slot before `@"Position"` at `0x130` and `@"Name"` at `0x138`, while each of the three loaded cfstrings goes to the callee at `0x9fd880` and its result lands at `[sp, #0x198]`/`0x1a0`/`0x1a8`.

That makes the conclusion **stronger**: the key really is `M`, read rather than inferred from the ten-locale agreement. The evidence file now carries both reads and the correction.

### The other four

| Where | Claimed | Actually |
|---|---|---|
| `docs/canon/LOGIC-FACING.json` `exceptions_note` | rule 15 refuses an excepted file that quotes a corpus value | it has a **six-character floor**, so `M` and `Name` — rows *this branch adds* — pass. Measured, now stated. |
| observation `reverify.expected` | the guard proves every member is the row's own value | it does not; `check-labelsets-are-derived.py:15-17` allows extras on purpose, which is what `variants` is for |
| observation `limits[2]` | a Logic rename would show as `check-labelsets-are-derived.py` failing | it would not: that guard reads the **committed** corpus index and never opens an installed Logic, so the break is silent until somebody rebuilds |
| `check_text`'s comment | the exception list is no longer trusted unchecked | the proof is gated on `changed_paths` — correct, since with none nothing rests on the list, but the comment did not say so |

### Why the floor is stated rather than lowered

Measured at `CITABLE_QUOTE_MIN = 1` against all six excepted files: **every one is refused for its own schema key `note`**, which Logic ships as a value (verbatim in `strings/it`, as `Note` in twelve more rows across four corpora). Lowering the floor would empty the list, not tighten rule 15.

Two cases pin it, and both were mutation-tested:

- `test_rule_15_has_a_six_character_floor` — `M` and `Name` scan clean, `Position` (same change, one character over) is the control so the case cannot pass on a scanner that finds nothing. Goes **red** when the floor is set to 1.
- `test_the_floor_is_not_movable_for_this_list` — all six refused at a patched floor of 1. Goes **red** when `_citable_strings_in` is stubbed to return nothing.

`python3 Scripts/run-repo-guards.py` → **all 75 passed**, including `test_labelset_census_only_shrinks.py`, which is red on `main` and fixed by this branch.

### Force-push

The six inherited commits carried `X-Claude-Session:` footers and the pre-push gate refuses them. Their messages were stripped; **the trees are byte-identical** (`git diff --stat f4581eeb 07755d77` is empty), and the CommitLore trailers survived. Old head `ff94093c` → `b0242ea0`; the reviewed content is unchanged.

Issue #884, opened 2026-09-14
Feature: Add verified automation-point editing for deterministic fades


Dominik-1980:
### Preflight

- [x] I searched existing issues for related work.

### Priority

P1 - unlocks a high-value Logic Pro workflow

### Primary area

mixer

### Problem

The current automation surface can switch Logic’s automation mode and move a fader through MCU, but it cannot directly create or inspect automation points. A simple request such as ‘fade the last five seconds to silence’ therefore becomes a fragile real-time Write/Touch/Latch workflow.

In a live test, the attempted operation switched Logic to **Write**, opened automation mode, and triggered Logic’s warning that Write is dangerous and that Touch or Latch should be used instead. After dismissing the warning, there was still no deterministic way to place two volume points at exact positions or verify the resulting envelope.

This is independent of UI language: the same missing primitive remains with English Logic. Mixer volume writes are not equivalent to editing track automation data.

### Proposed outcome

Expose track-automation data as an explicit, verified MCP surface instead of emulating envelope editing through real-time fader movement. For example:

- `automation.get_points(target_ref, parameter, range)`
- `automation.add_point(target_ref, parameter, position, value)`
- `automation.set_point(...)` / `automation.delete_point(...)`
- optionally `automation.replace_range(...)` or a higher-level `automation.fade(...)`

Positions should support Logic musical time (bar/beat/tick) and, where reliable, seconds. Volume values should accept dB including silence / negative infinity. Each mutation should return the resolved track, parameter, requested position/value, observed position/value, and verification state.

A canonical use case is: on a 20-second MIDI drum track, preserve the current level until 15 seconds, place a point there, place a second point at 20 seconds at silence, and verify the two-point fade.

### Safety and contract notes

- Require an explicit stable track target (`target_ref` or an unambiguous index/name) and an explicit automation parameter.
- Do not enable global **Write** mode as an implementation side effect.
- Use State A only when point count, track identity, parameter, time, and value are independently read back from Logic.
- Return State B/C when the automation lane is unreadable or the target is ambiguous; do not infer success from a click or fader command.
- Keep the change undoable as one Logic undo step where possible.
- Avoid leaving automation mode, warning dialogs, or playback active after failure.
- The operation should be locale-neutral or have measured `en-US` and `de-DE` UI paths.

### Done when

- An MCP client can create a volume fade over the last five seconds of a 20-second track by specifying exactly two points.
[...]

PR #946, opened 2026-09-20
obs(884): the route the earlier reading called a wall was a submenu parent, not a command

MongLong0214:
## What this corrects

`docs/observations/2026-09-15-the-automation-lane-is-legible-and-inert.json` concluded that no route measured there can create an automation point. It reached that by actuating `Mix > Create Track Automation` and finding Logic's undo stack unmoved.

That item is a **submenu parent**. The en-US and ko-KR menu censuses already in this repository show seven leaves beneath it, and the parent row carries no `globalMenuItemCall:` identifier while every leaf does. The earlier run drove a container, not a command — its silence was an unaimed instrument, not a wall.

## What was measured

Driving the leaf `리전 경계에 2개의 오토메이션 포인트 생성` with a region present moves Logic's own undo stack to `리전 경계에 2개의 트랙 오토메이션 포인트 생성 실행 취소`, the name Logic ships for that operation under its own `#und` key. **An automation point can be created without a coordinate.**

Three controls bracket that, all in the same run:

| control | reading |
|---|---|
| `tracks.create_instrument` first | the undo-label reader detects a change at all |
| the same leaf with **no** region | label unmoved — state matters, the leaf is not unconditional |
| the committed script mutated to click the **parent** | stack unmoved, a menu left open, exit 2 |

The last one re-measures the 2026-09-15 finding rather than arguing with it.

## What is still missing

Readback. With two points in the lane and the automation view on, a census of the arrange window to depth 14 found **543 elements and 0 that represent a point**. The nine `AXHandle`s are region-border and loop handles. So a created point cannot be located or valued through this surface, and the verdict is `partial`, not `yes`.

`오토메이션 값` now carries an `AXValue` where the 2026-09-15 reading found none, but it is one number for the whole track, it is not settable, and its meaning is **not** established here — the 90/127 ratio to `AXMaxValue` invites a MIDI-scaled reading and no control could be run, because this host has no control surface so `mixer.set_volume` refuses. That is recorded as unverified, not as a measurement.

The Event List and `Mix > 오토메이션 변환` are the two untried routes to a per-point readback; this record rules neither in nor out.

## The reverify

`Scripts/observations/reverify-automation-point-creation.sh` brackets the run with Logic's own undo instead of comparing labels. A bare before/after comparison cannot tell "nothing happened" from "the same operation happened again" — a second run of the same leaf writes the identical label — and the first version of this script failed exactly that way, reading a working leaf as dead. The stack is now walked **off** the operation first, and that is checked.

Its canon references are literal and Korean rather than assembled from a locale variable: an interpolated reference is unresolvable to `check-canon-citations.py`, and splitting the scheme to hide references from the checker would be worse than having none. Adding a locale means adding that locale's checkable references.

## Verification

- `python3 Scripts/run-repo-guards.py` — all 75 pass on this branch
- reverify exits 0 on the real tree; the parent-clicking mutant exits 2; `grep -c MUTATION` is 0 after restore
- `Scripts/test_labelset_census_only_shrinks.py` fails identically on a clean worktree of `HEAD` and is not from this change

Refs #884

## Canon

The values this change's derivation compares are the record's own, and the record cites them. The
undo-stack label the reconciliation now matches state-by-state is the undo form of this command,
which is why forging it must be caught rather than grouped around:

`logic-canon://strings/Contents%2FFrameworks%2FLogic.framework%2FVersions%2FA%2FResources%2FLocalizable.strings/ko/Create%202%20Track%20Automation%20Points%20at%20Region%20Borders%23und#value`

리전 경계에 2개의 트랙 오토메이션 포인트 생성

The automation-value rows the census counts, whose readings M2 zeroes:

`logic-canon://strings/Contents%2FFrameworks%2FLogic.framework%2FVersions%2FA%2FResources%2FLocalizable.strings/ko/Automation%20Value#value`

오토메이션 값

The menu command the actuation issues:

`logic-canon://strings/Contents%2FFrameworks%2FLogic.framework%2FVersions%2FA%2FResources%2FLocalizable.strings/ko/Create%20Track%20Automation#value`

트랙 오토메이션 생성

No Korean label is typed anywhere in the derivation or the self-test; both read these values from
the record and compare them field to field against the raw evidence.


MongLong0214:
# Response to round 2 — AUTO-CENSUS-001

## The repair

`reconcile_undo_stack_boundaries()` now runs before any grouping. It requires the
raw `undo_stack_boundaries.readings` and the record's census states to cover each
other one-for-one: every record state must have a raw reading, every raw reading
must have a record state, roles must be unique on both sides, and for each role
the raw `item_1` — and `item_2` whenever either side carries it — must equal the
record's `undo_item_1`/`undo_item_2`. `state_groups()` runs only if reconciliation
produced no failures, and it consumes the reconciled raw pairs, not the record's
own summary. No Korean undo label is hard-coded anywhere; the comparison is field
to field between raw evidence and record.

Your reproduction — replacing all five raw `item_1` values with one forged label —
now returns 1. It is committed as mutant M7.

## What I found while adding the witness you asked for, which you should check

You asked for "a mutation witness proving forged raw boundary readings return
nonzero". Adding one is not sufficient, and I can show that, because the first
witnesses I had did return nonzero for the wrong reason.

Every mutant in the self-test edits **data**. A data mutant cannot witness a
predicate that data does not reach, so I mutated the derivation **code** instead,
four mutants, each reporting its applied-edit count:

| code mutant | self-test exit | verdict |
|---|---|---|
| C1 `item_1` reconciliation predicate replaced by constant `True` | 0 | **SURVIVED** |
| C2 missing-reading check neutered | 1 | KILLED |
| C3 duplicate-reading-role check neutered | 1 | KILLED |
| C4 `item_2` reconciliation predicate replaced by constant `True` | 0 | **SURVIVED** |

So the two predicates that are the whole point of this repair had no witness,
while seven data mutants all reported KILLED. M7 and M8 look like the `item_1`
witnesses and are not: with C1 applied they still exit 1. They forge `item_1` into
a shape that collapses the undo-stack grouping, and the **grouping** rejects them.
Collateral damage was being read as a kill.

My first attempt at an honest witness also failed, and failed differently. Swapping
two readings' `item_1` labels preserves group *sizes*, but it changes group
*membership*, so the census equality comparison rejects it and C1 still survived.
The witness has to preserve the grouping exactly:

- **M11** relabels *every* applied-state reading to one new string. The three
  applied states still equal each other and still differ from the two undone
  states, so the partition is byte-identical in shape. Only the state-by-state
  record comparison can see the disagreement.
- **M12** forges one reading's `item_2` to another reading's value and touches no
  `item_1`, which no grouping step reads.

With M11 and M12 present, all four code mutants die: C1 1, C2 1, C3 1, C4 1.

## Commands and results

- `selftest-derive-automation-census-difference.py`: exit 0 — unmutated data passes,
  nine data mutants rejected, every mutation printing its edit count.
- `derive-automation-census-difference.py`: exit 0 on the committed censuses.
- `check-observation-records.py`, `check-canon-citations.py`,
  `check-guards-have-self-tests.py`, `run-repo-guards.py`: all exit 0.

## Limitations, unchanged

Nothing here observes live Logic. The censuses are consumed exactly as committed and
no evidence JSON was edited — a number that cannot be reproduced from the censuses
is a finding, not a reason to edit the census. Census capture provenance and live
automation readback remain unverified, as the record's own `Limit:` already states.

MongLong0214:
Automatic review budget for this PR is exhausted (3/3). Attempt 1 and attempt 3 both ended with no artifact — attempt 3 failed on an account-level executor usage limit, not on this change — so the repair above has **not** had an independent follow-up review. The round-2 review is the last real one, and its blocker `AUTO-CENSUS-001` is what the commit above addresses.

This needs a maintainer decision rather than another automatic round.

Issue #947, opened 2026-09-21
swift test aborts the whole process on main: FileHandle.fileDescriptor raises an ObjC exception the reader cannot catch


MongLong0214:
`swift test` aborts the whole test process on `main`: `FileHandle.fileDescriptor` raises an ObjC exception the reader's `do`/`catch` cannot see.

## What happens

A full `swift test` run on `main` (8c6681b7feac4c14dab42cff57ed788928b1ac50) terminates with

```
*** Terminating app due to uncaught exception 'NSFileHandleOperationException',
    reason: '*** -[NSConcreteFileHandle fileDescriptor]: No such file or directory'
...
4  LogicProMCPPackageTests  ...$s11LogicProMCP13readAvailable...
8  LogicProMCPPackageTests  ...QualificationSubprocessSession...startReaders...
libc++abi: terminating due to uncaught exception of type NSException
```

The process aborts (signal 6), so the run produces no result line at all — not a failure count,
not a pass count. Every test that had not yet reported is simply lost.

The second reason string `'*** -[NSConcreteFileHandle fileDescriptor]: unknown error'` also occurs;
both are the same call.

## Why

`readAvailable` reads `handle.fileDescriptor` on **every** iteration of its loop:

```swift
private func readAvailable(_ handle: FileHandle, upTo limit: Int = 64 * 1024) throws -> Data {
    var buffer = [UInt8](repeating: 0, count: limit)
    while true {
        let n = buffer.withUnsafeMutableBytes { Darwin.read(handle.fileDescriptor, $0.baseAddress, limit) }
```

`Sources/LogicProMCP/Qualification/QualificationTransport.swift:4432`

`FileHandle.fileDescriptor` is an Objective-C property that **raises an `NSException`** when the
handle is closed. It does not return -1 and it does not throw a Swift error, so neither the
`do`/`catch` in the stdout reader nor the one deliberately added to the stderr reader can catch it.
An `NSException` crossing a Swift frame terminates the process.

The shutdown path closes exactly those handles while the reader threads are still in that loop:

```swift
if readers.wait(timeout: .now() + shutdownGrace) == .timedOut {
    try? outputPipe.fileHandleForReading.close()
    try? errorPipe.fileHandleForReading.close()
```

`Sources/LogicProMCP/Qualification/QualificationTransport.swift:4588-4589`

That close is the intended way to unblock a reader stuck in `read(2)`, and it works — but if the
[...]

PR #948, opened 2026-09-21
fix(dialog): the receipt named a write that never happened, and an unparsed result released the fallback (#943, #945)

### LINKED ISSUE #943: goto_position receipt contradicts itself: menu_could_not_be_closed_write_attempted_true beside write_attempted: false
Found by an independent review of #941 and confirmed to pre-date it.

## What

`AccessibilityChannel+Transport.swift` classifies a failed menu cleanup as
`.menuCouldNotBeClosed(writeAttempted: Bool)`. That associated value does **not** mean a position was written. It is decoded from whether the script appended the literal `" after menu actuation"` to its refusal string — that is, whether a **menu click** happened. `menuActuationAttemptedBeforeUnsafeRefusal` reads it correctly with that meaning, and #941's new reconciliation trigger depends on that same reading.

`diagnosticLabel` does not. It emits

    menu_could_not_be_closed_write_attempted_true

into a response whose top level also carries

    "write_attempted": false

Both are correct under their own definitions and they contradict each other on the wire. A consumer reading the diagnostic label concludes a position write was attempted when none was.

## Why this was not fixed in #941

The label is a published receipt token. Renaming it is a public-surface change that deserves its own change and its own review, not a rider on a blocker fix.

## Suggested shape

Rename the associated value to `menuActuationAttempted` and the diagnostic token to `menu_could_not_be_closed_menu_actuation_attempted_<bool>`, in one commit that also updates every consumer and every test asserting the old spelling. Check `Scripts/` and `docs/` for anything pinning the old token before changing it.

### LINKED ISSUE #945: An unrecognised script result is safe_to_retry, so any sentinel typo fails open
Raised by round 6 of the #941 review, declined there as out of scope, recorded here because the underlying question is real and is not specific to any one sentinel.

## What

`classifyGotoPositionDialogResult` maps the generated script's result string to a classification. Anything it does not recognise becomes `.unexpectedResult`, whose `performedDialogSafetyObservation` defaults to true and whose `requiresUnsafeUIRefusal` is false. The route then returns an ordinary `safe_to_retry: true` result, and a later fallback can actuate.

That means a single character wrong in ANY sentinel — a case change, a missing underscore, a renamed constant on one side of the boundary only — silently converts a terminal safety refusal into a retryable failure. The reviewer demonstrated it with a lower-cased `menu_validation_unreadable: menu_actuation_attempted=true`, but nothing about that sentinel is special: `MENU_DISABLED`, `DIALOG_NOT_READY` and every other exact-match case have the same shape.

## Why it was not fixed in #941

Making one sentinel case-insensitive while its neighbours stay exact would be an inconsistency dressed as a fix, and the script cannot emit a mis-cased literal — it returns a compile-time constant string. The reachable version of this defect is not "the script lied about its own spelling"; it is "the two sides of this boundary drifted", which is a property of the whole classifier.

## The actual question

Should an UNRECOGNISED result be safe-to-retry at all? An unparsed result is not an observation that nothing happened — it is the absence of an observation, and this repository's own rule for that is that unreadable is not absent. The conservative reading is that `.unexpectedResult` should refuse, and that a route which cannot classify its own script's output has no basis for releasing a fallback.

## Suggested shape

- Decide `.unexpectedResult`'s safety posture once, for every sentinel, and write the decision where the enum is defined.
- If it becomes a refusal, count the currently-reachable paths that depend on it being retryable before flipping it — measure, do not assume the count is zero.
- Add a test that the sentinel strings the script emits and the strings the classifier matches are the same set, so a drift on one side fails the build rather than failing open at runtime.

MongLong0214:
## What this changes

**#943 — the receipt named a write that never happened.** The dialog route's
`menuCouldNotBeClosed` payload was called `writeAttempted`, but the value it
carries is whether the *menu leaf was actuated*. The diagnostic label it built,
`menu_could_not_be_closed_write_attempted_<bool>`, therefore said
`write_attempted` about a route that had not reached a position write. Renamed
the associated value to `menuActuationAttempted` and the label to
`menu_could_not_be_closed_menu_actuation_attempted_<bool>`.

The envelope key `write_attempted` is deliberately untouched. Measured before
deciding: `writeAttempted`/`write_attempted` occurs 672 times across
`Sources`/`Tests`/`Scripts`/`docs`; split by pattern,
`menu_could_not_be_closed_write_attempted` = 0,
`menuCouldNotBeClosed(writeAttempted` = 0, and `writeAttempted:` = 61 — those 61
are the server-wide public receipt field emitted by the export executor, saga
journal, dispatchers and qualification transport. Renaming it would turn a
payload-label fix into a wire-protocol break.

**#945 — an unparsed result released the fallback.** `.unexpectedResult` was
reaching the generic `safe_to_retry: true` return, so a script sentinel one
character wrong released the bar-slider fallback as though the dialog route had
cleanly declined. An unparsed result is the *absence* of a dialog-safety
observation, not an observation that nothing happened. It is now terminal.

**A parity test, and what it found.** The new test reads the result sentinels the
generated script actually emits and the sentinels the classifier actually matches,
both from production source, and requires the two to agree. It immediately found a
classifier-only `DIALOG_NOT_READY` arm the script never emits; that arm, its enum
member and its entries in `requiresUnsafeUIRefusal` and
`dialogActuationMayHaveOccurred` are gone, and a regression test pins the stale
value to `.unexpectedResult`.

## Evidence

Four mutants, each carrying an applied-edit count so a substitution that silently
matches nothing cannot read as a surviving mutant. Every one compiled before it
ran, so every exit code is a test result rather than a build failure.

| mutant | edits | build | `swift test --filter Issue529` | verdict |
|---|---|---|---|---|
| M1 restore the old `write_attempted` diagnostic token | 1 | 0 | 1 | KILLED |
| M2 drop `.unexpectedResult` from the `safeToRetry` false list | 1 | 0 | 1 | KILLED |
| M3 drift one classifier arm to `SELECT_ALL_ARMEDX` | 1 | 0 | 1 | KILLED |
| M4 re-add a classifier-only `DIALOG_NOT_READY` arm | 1 | 0 | 1 | KILLED |

M3 and M4 were both killed by `generated dialog result sentinels and classifier
sentinels remain in lockstep`, attributed by name from the failure output. M3 is
the one that matters: it leaves the sentinel's first token unchanged, so the
set-equality half of the test cannot see it, and only the literal probe kills it.

That literal probe is a repair this PR carries. Written against the *name* set, the
probe fed the classifier a bare `DIALOG_INPUT_ISSUED` and got `.unexpectedResult`.
That is not production drift: the script emits that sentinel at 18 sites
(`AccessibilityChannel+Transport.swift:1972`–`:2047`) and every one carries a
second segment — `: SELECT_ALL_ARMED` or `: POSITION_INPUT_ARMED` — which the
classifier's `hasPrefix` arms at `:2545`/`:2550` include. The name extractor stops
at the first `[A-Z_]` run, which is right for comparing two sets of names but is
not a value production ever emits. `issue529EmittedDialogResultLiterals` captures
the literal head up to the first interpolation instead.

`swift test --filter Issue529`: 59 tests, exit 0.
`Scripts/ci-forbid-dead-expect.sh` and
`Scripts/ci-forbid-hardcoded-menu-bar-item.sh`: exit 0.

## What this does not show

No live Logic drive. The dialog route was exercised through the generated script
text and the classifier only.

The full suite is red under parallel load on this path, and it is not this branch.
Baseline measured at `eadb5f4a` in a separate worktree: **20** distinct failing
tests there against **4** here, and the 4 are a subset. `TraceClearAuditTests` and
`OperationTraceCoverageTests` both pass in isolation and fail together, and the
process also aborts on the `FileHandle.fileDescriptor` `NSException` filed as #947.

Closes #943
Closes #945

## Canon

This change states no new fact about Logic's interface: it renames an enum payload, makes an
unparsed script result terminal, and removes a classifier arm for a sentinel the generated script
never emits. No label in any language is added, removed or re-spelled, and the route's label
resolution is untouched.

The claims the edited file makes about Logic, and which this change's classifier decides results
for, rest on the Navigate → Go To → Position… menu chain. Cited here because the file is
Logic-facing, not because the diff moves any of them.

The menu-bar item:

`logic-canon://strings/Contents%2FFrameworks%2FLogic.framework%2FVersions%2FA%2FResources%2FLocalizable.strings/en/Navigate%23mti#value`

Navigate

The submenu parent:

`logic-canon://strings/Contents%2FFrameworks%2FLogic.framework%2FVersions%2FA%2FResources%2FLocalizable.strings/en/Go%20To#value`

Go To

The leaf:

`logic-canon://strings/Contents%2FFrameworks%2FLogic.framework%2FVersions%2FA%2FResources%2FLocalizable.strings/en/Position%E2%80%A6#value`

Position…


MongLong0214:
## Implementer response to round 3 (`354a2c24`)

**GOTO-DIALOG-MALFORMED-SAFETY — accepted, and repaired at the cause rather than the site.**

Confirmed exactly as reported: `.malformedPayload` was absent from both lists, so
`requiresUnsafeUIRefusal` answered `false` and the terminal receipt said `safe_to_retry: true`
for stdout that is not even the `{"result": …}` shape.

I did not take the "add it to the false list" option the finding offered as the small mechanical
fix, because it closes this instance and leaves the mechanism: `performedDialogSafetyObservation`
listed the outcomes that had *not* read the dialog state and defaulted the rest to observed, so a
case became fallback-safe by saying nothing — which is exactly how this one shipped, and exactly
what your `.menuPickFailed` note says is waiting for the next one. The lists are inverted and
exhaustive now. The switch names the outcomes that *did* read the state, anything unlisted answers
`false`, and there is no `default` arm, so adding a case to the enum is a compile error until
somebody declares which side it is on.

Every current answer is preserved — I enumerated them from the old shape rather than re-deciding
any. `.menuPickFailed` in particular stays on the observed side, on your trace of its one reachable
bare emission site (after `cleanupState == "CLOSED"`, strictly before the leaf click). That trace
is now a written limit on the commit (`Limit:` on `r-948malformed`) rather than a fact in a review
thread, because it stops being true the moment a `MENU_PICK_FAILED` return is added after the click.

**Tests.** `malformedDialogPayloadDoesNotFallThroughToRetryableRoute` reproduces your scratch
result through the public entry point and asserts the envelope shape.
`everyDialogClassificationDeclaresItsFallbackSafety` pins `requiresUnsafeUIRefusal` for all
seventeen `Failure` cases plus `.driven`.

That second test first measured nothing, which is worth recording: written as
`#expect(classification.requiresUnsafeUIRefusal == expected)` it **passed** under the mutant that
puts `.malformedPayload` back on the observed side, while the reproduction test failed. A top-level
`Bool == Bool` inside `#expect` passes unconditionally on this toolchain — the same footgun already
recorded against this file as `r-941round11`. Rewritten as `#expect(flag)` / `#expect(!flag)` under
an `if`, that mutant fails it at line 1407, and a second mutant moving `.menuPickFailed` to the
refusing side fails it too, so both columns of the table are live rather than only the `true` one.

**NIT 1 (`sliderWrites == 0`) — accepted, answered in the comment.** You are right that it cannot
fail for the claim the comment made: `gotoPositionViaBarSlider` writes the bar/beat slider on no
path at all. The discriminator is `safe_to_retry`, and the comment now says that instead of
claiming more than the assertion proves. I left the assertion in place as the shape this file's
other 18 cases use rather than changing them here.

**NIT 2 (parity granularity) — noted, not changed here.** The parity check compares at the leading
`[A-Z][A-Z_]+` run, so the three `MENU_PICK_FAILED` messages collapse to one token on both sides.
That is a real coverage gap and it is the same blind spot your `.menuPickFailed` note describes.
Widening it changes what the guard asserts about every sentinel, not just that one, so it is
tracked separately rather than folded into a blocker repair.

`swift test --filter Issue529`: 61 tests pass, up from 59.

Issue #955, opened 2026-09-21
Backlog: plug-in parameter reads, snapshots, batch compensation and automatic EQ (ADR-009/013 remainder)


MongLong0214:
# Deferred plug-in reads, snapshots, batch application and compensation

> **Execution SSOT r2 — 2026-09-21. OPEN / `scope:backlog`.** This issue explicitly owns the remaining generic read/snapshot/batch interfaces from #292 and full-plan/reset/comparison/application requirements from #301. It is not a new numbered ADR. #308 selects a concrete stage. Source checkpoint: `1b36b90bb8f339c82a3d606c2f44cab08ef18da6`; no implementation or live qualification is claimed by this issue revision.

## 1. Scope and activation

Preserve the useful single-operation surface. #292 owns shared stock capability correctness; #301 owns the specific EQ state reader and bounded existing writer. This issue owns the generic parameter/read/snapshot/diff interface and later multi-parameter application, reconciliation, verified inverse and explicitly confirmed recommendation application.

Activate only a named user workflow and its required stage. **A needed B0 parameter reader does not wait for the entire B2/B3 batch engine.** Batch activation requires a complete before-state for its supported subset and an independently demonstrated inverse. A generic request to finish every ADR is not authority to build a platform.

Do not create a second catalog, a separate batch engine per plug-in, a vendor SDK, distributed jobs or durable recovery. #299 and #306 own their model/vendor-specific qualification; neither is a prerequisite for a small stock EQ batch.

## 2. Original requirements retained, with no unnamed follow-up

| Original proposal | Authoritative owner / stage |
|---|---|
| #292 `get_param_verified` and parameter-state read | **B0 here**, reusing current per-control reads; after-read inside a writer is not already this public API. |
| #292 snapshot read, `diff_snapshot`, per-instance supported-state resource | **B0/B1 here**. #301's EQ-specific state reader is reused, not reimplemented. |
| #292 `apply_snapshot_verified` and accurate partial-snapshot failure | **B2/B3 here**. |
| #301 `apply_channel_eq_plan_verified`, multi-band before-state and inverse | **B1–B3 here**. Existing named scalar writing is not full-plan application. |
| #301 reset/compare-state APIs | Compare uses **B1**; reset uses an explicitly qualified known-state plan in **B2**, not arbitrary zeros or an unverified plug-in reset/Undo button. Extra aliases are added only if they expose a distinct selected contract. |
[...]

Issue #951, opened 2026-09-21
Canon: ready after #952 — preserve evidence checks, separate CI-only policy and improve issue guidance


MongLong0214:
# Canon maintenance — contributor-friendly guidance without weakening evidence checks

> **Execution SSOT r3 — 2026-09-21. OPEN / READY. The #950 dependency is satisfied by merged PR #952 (`06e5d37c8bd8fb9db62431618b0d386595f63ba9`).** Start from current main containing that change. Do not perform the old CI migration again or wait for #950 to reopen. This issue owns the remaining forms, advisory bot, diagnostics, documentation and CI-only policy separation. #308 schedules work; #956 owns repository navigation and milestone metadata, not this implementation.

This issue states no fact about Logic. It specifies the repository's evidence-checking and contributor workflow; it does not assert that a particular Logic UI or audio operation works. Existing corpus/observation requirements on a future implementation PR remain applicable to its actual changed paths.

## 1. Outcome, first task and actual baseline

A contributor can submit an issue without building our evidence corpus. A supplied declaration survives the issue form's rendering. The advisory bot distinguishes a correctable evidence-format problem from its own inability to evaluate, maintains one positively identified managed note, and resolves that note without claiming runtime correctness. CI-only policy is owned by the existing CI-integrity execution path rather than the Canon checker.

**First bounded deliverable:** repair Canon textareas in the existing issue forms, with rendered-output tests, and make evidence optional for initial intake. This does not wait for the entire policy extraction or for live Logic. Then implement the shared result/comment path, and finally perform the coordinated CI-only extraction. Small PRs are allowed; each uses `Refs #951` until the whole accepted scope is complete.

The initial investigation was at `eadb5f4`; those filing links are history, not a reset target. The current handoff includes #949 and #952. On 2026-09-21 ruleset 19060230 was read back with `build`, `compile`, `test`, `pr-policy`, app 15368, strict freshness and no bypass actors. Re-read actual files/configuration when implementation starts; unavailable permissions mean unverified, not absent. This editorial revision did not run new workflow or Logic tests.

### Concrete remaining defects

1. Canon issue textareas use `render: text`, which wraps an author's declaration in a fence that the checker intentionally ignores. Fix the form, not the rule that hidden/example text cannot count as an active declaration.
[...]

Issue #956, opened 2026-09-21
SSOT alignment: finish repository navigation, milestone cleanup and native roadmap pinning


MongLong0214:
# Complete repository navigation and native GitHub metadata alignment

> **Execution SSOT r4 — 2026-09-22. OPEN. D documentation can proceed; M native metadata remains partly unapplied.** #308 owns selection; each owning body is the execution contract. This issue does not authorize runtime changes or another full ADR audit. Current comparison anchor: `ffcf0bf8ac0f93a7615fdce6b57d0b2fd5370575`.

This issue states no fact about Logic. It concerns repository navigation, historical ownership and GitHub metadata; future PR checks still apply to actual changed paths.

## 1. Separately verifiable delivery

| Track | Actual status at this reconciliation | Done only when |
|---|---|---|
| D — repository Markdown/navigation | Earlier CI/Canon PRs changed some rows; the complete authority migration is not claimed done here. | Actual merged entry files point to #308/current owners, preserve technical evidence and pass applicable existing checks. |
| M — native milestones and pin | Earlier #290/#816 assignment to milestone11 was applied. This overlap-review pass did not create/rename milestones, clear memberships or pin #308. | Native objects/memberships/pin are read back after real successful changes. |

Updating this body or a target mapping is not native milestone completion. A missing metadata action does not block D, #291/#965 or other independent product work. Use `Refs #956` until both tracks are complete or a remaining part is explicitly transferred to an open owner.

## 2. Current owners and concrete newly found drift

Read #308 and only the relevant current owner bodies, plus actual current states when editing rows. Preserve all original ADR owners, ADR-019's observation-ledger document and #892/#904, the #955/#957/#958 remainders, and new #965–#971.

Important current dispositions:

- #291 R0–R3 is now **selected**, not blanket backlog. R4 sidechain mutation remains later scope.
- #965–#971 are incremental extensions over #28/#95/#102 and current Saga, not a replacement audit/workflow platform.
- #448 is a non-completion transfer: observation #965, naming #968, exact layout #969, color #970 and composition #966/#971. Its shipped sort is retained.
- #293 is a transfer to #302, not completed MIDI reading. Do not change states to satisfy a table.
- #301 was read closed while its body retained E1/E2; preserve the true owning requirement and reconcile actual closing evidence, not an invented transfer to #955/#308.
- #950/#952 are delivered CI work. #951 must begin from current merged #959 code and remaining evidence, not redo old form/bot/extraction work.
[...]

PR #959, opened 2026-09-21
Canon: forms that carry a declaration, one advisory note, and CI policy with its own owner (#951)

MongLong0214:
Refs #951. This does not close it: I02 (a real deployed issue event showing create → update → resolve → recurrence on one owned comment) cannot be observed until this is on the default branch, and #951 §10 says to keep the issue open through that observation rather than claim deployed completion.

## What a contributor sees now

A Canon declaration typed into an issue form arrives as visible prose instead of inside a fence the checker deliberately ignores, evidence is optional at intake, and the advisory bot keeps one note that it edits rather than a new accusation per failing edit. None of that weakens a committed-artifact check: a fenced or HTML-commented declaration is still ineligible, a supplied invalid citation is still invalid next to a no-fact declaration, and the protected-path opt-out is still refused.

## Workstream A — the forms

`render: text` came off every Canon evidence/declaration textarea in `bug_report.yml`, `feature_request.yml` and `live_qa_failure.yml`, and stayed on the log and reproduction fields, where a fence is the point. Field IDs are unchanged. `Scripts/test_issue_form_canon_field.py` renders the Issue Form Markdown GitHub would produce — heading included — and runs the real checker over it, because YAML that parses proves nothing about what the checker reads.

## Workstream B — one result, two callers

`Scripts/check-canon-citations.py --format json` exposes the diagnostics the human output already computed. There is no second citation engine. The three categories are evaluation completed/satisfied, evaluation completed/actionable, and evaluation failed — and the third is never success: an empty or malformed result, a crash, an unreadable input or an unavailable corpus is an error, not an empty issue body and not proof the author was wrong. `--text` keeps its narrower route; adding `--format json` does not turn `--changed` into the weaker mode.

## Workstreams C and D — the note

`Scripts/canon_issue_bot.py` decides; `canon-issue.yml` carries out. Ownership is the marker `<!-- logic-pro-mcp:canon-guidance:v1 -->` **and** the authenticated bot identity, because a marker is something any commenter can type. The comment list is paginated in full, so an owned note on page two does not cause a duplicate. Identical content writes nothing. A tooling error never marks a note resolved and never posts an accusation. Copied input is escaped for Markdown, `<details>`, backticks, workflow annotations and mentions, and displayed diagnostics are truncated while the evaluation input is not.

## Workstream E — the CI lists get their own owner

Four files under `docs/canon/` said nothing about Logic. They held job names, guard names, test counts and skip allowances, and lived there only because the merge-base ratchet happened to be written inside `check-canon-citations.py`. The cost was measured, not argued: `docs/canon/` is a Logic-facing prefix, so a correction to one of them could not use the `states no fact about Logic` opt-out and had no row of Apple's data to cite. #937 is that record.

`Scripts/ratchet.py` is the comparison now — `Ratchet`, the member extractors, `History`, `check()` — neutral about both Logic and CI, with no dependency back on either owner. `check-canon-citations.py` keeps the Logic-facing lists; `check-every-ci-job-is-required.py` declares the four CI lists over `.github/ci/` and calls the same function. Two owners, one comparison, no second authority on what `shrink` means.

| Moved list | Ratcheted key | Compared members | Direction | External base | Owner | Test |
|---|---|---|---|---|---|---|
| `.github/ci/CI-GATE.json` | `required_commands` | list of strings | grow: shrink refused | `legacy="docs/canon/CI-GATE.json"` at `git merge-base` | `check-every-ci-job-is-required.py` | `TheRatchetsOverTheCiLists`, `TheMigrationIsComparedNotBootstrapped` |
| `.github/ci/CI-GATE.json` | `not_required` | map — its KEYS | shrink: grow refused | same | same | `TheRatchetsOverTheCiLists` |
| `.github/ci/CI-SKIPS.json` | `allowed` | map — key with its `skips` count | shrink: raising an allowance refused, lowering allowed | same basename under `docs/canon/` | same | `test_raising_a_skip_allowance_fails`, `test_lowering_a_skip_allowance_passes` |
| `.github/ci/GUARDS-WITHOUT-A-TEST.json` | `guards` | map — its KEYS | shrink: grow refused | same basename under `docs/canon/` | same | `test_a_guard_arriving_without_a_test_fails` |
| `.github/ci/GUARD-TESTS-BLIND-TO-THEIR-GUARD.json` | `guards` | map — its KEYS | shrink: grow refused | same basename under `docs/canon/` | same | `test_a_guard_arriving_blind_to_its_own_gate_fails` |

`CI-GATE.json`'s `workflows` key is NOT a ratchet entry: it is a map the guard checks structurally, job by job, against the workflow files themselves. An earlier revision of this table listed it beside `required_commands` and left `not_required` out, which named the wrong half of what a merge-base comparison actually protects there.

`legacy` is declared by the OWNER and never read out of the moved file: a policy that names its own predecessor chooses what it is compared against, and a PR-controlled `migrated_from` field inside a file would let the change under review pick its own baseline. That field is still checked; it is not believed. A list absent at the merge base is unratcheted on the branch that introduces it, and the branch that moves a requirement is exactly the branch where losing one is easiest.

`LOGIC-FACING.json`'s `exceptions` went from six entries to two. The four CI lists are outside every Logic-facing prefix now, so they need no exemption at all — the stronger remedy, because an exemption asks rule 15 to re-prove on every run what a location settles once, and it is one rename away from lapsing. `check_exceptions_state_no_fact` fails on an exception naming a file that does not exist, so a stale entry could not have been left behind quietly.

### Separation demonstrated, not asserted

`TheTwoOwnersFailIndependently` carries BOTH guards over ONE tree, so each assertion is about the same repository state seen twice rather than two fixtures chosen to agree. A CI-topology defect fails the CI owner and leaves Canon at zero (M07); a citation defect fails Canon and leaves the CI owner at zero (M08); and M09 imports `run-repo-guards.py` and asserts that its discovery actually returns both owners and both suites, since a moved JSON file with no invocation under a required path is not protected.

Fifteen mutants over two sweeps, every one killed, control green at both ends of each. The sweeps are in the record; the cases are committed, because a control proved once by hand and thrown away goes quietly wrong.

The move exposed a real defect in three fixtures: they copy `check-canon-citations.py` into a temporary root, that file now imports `ratchet.py` by path, and a root without it does not run a weaker guard — it runs no guard at all, and the case reads the traceback as the refusal it wanted. Caught by `test_policy_literals_against_canon.py` going red on its own control. Fixed in all three.

## Workstream F — the documentation

`docs/canon/README.md` gained an entry section: what a contributor provides, what a committed digest establishes, what stays human or runtime, that an issue is advised while a pull request is gated, and where CI-integrity policy lives. Two statements there were stale and are corrected: the enforcement table named a `canon-citations-in-the-pull-request` job that #952 removed, and it said nothing checks an issue mechanically yet, which workstreams C and D made false. The job names in the new table were read out of the workflows rather than remembered — the tree-wide sweep runs in `ci.yml`'s macOS `test` job, not in `guards`.

## Canon

This change touches two files under `docs/canon/`, so it may not use the opt-out, and the rule is derived from the paths rather than from what this body says about itself. The row it cites is the one the edited document now rests on: workstream F's entry section showed the citation form only as a placeholder, `logic-canon://<source>/<locale>#value`, which is the shape and not an example anyone can check. It now carries a worked one, and this is that row:

```
logic-canon://strings/Contents%2FFrameworks%2FLogic.framework%2FVersions%2FA%2FResources%2FLocalizable.strings/en/Trk#value
Trk
```

The honest limit, and it is #937's open remainder rather than something this PR fixes: `docs/canon/LOGIC-FACING.json` is the second of the two files, and its contents are path prefixes and notes — no row of Apple's data. The prefix rule counts it as Logic-facing because of where it sits. The route out of that is the `exceptions` list rule 15 proves, and I did not take it here: adding the prefix file to its own exception list to pass one correction is exactly the widening #951 §9 forbids, and this change shrinks that list rather than growing it.

## Limits

A matching committed digest establishes consistency with that artifact in this tree. It does not authenticate Apple provenance, select the right UI target, validate prose, prove a live write, or defeat an actor able to rewrite both checker and evidence. No pull request has been driven through the relocated gate on GitHub, and no real issue event has exercised the bot — I02 stays open. Every measurement below is local.

```
git diff --check                                  clean
Scripts/test_issue_form_canon_field.py            OK (11 tests)
Scripts/test_canon_citation_diagnostics.py        OK (26 tests)
Scripts/test_canon_issue_bot.py                   OK (63 tests)
Scripts/test_canon_citations_guard.py             OK (77 tests)
Scripts/test_every_ci_job_is_required.py          OK (80 tests)
Scripts/test_pull_request_event_classifier.py     OK (20 tests)
Scripts/test_third_party_imports_are_installed.py OK (12 tests)
Scripts/check-canon-citations.py                  exit 0
Scripts/check-every-ci-job-is-required.py         exit 0
Scripts/run-repo-guards.py                        all 90 passed — 89 that can refuse, 1 that only counts
```

## What the review changed

Four findings, each about two answers that could not be told apart.

**A readable merge base with unreadable history.** Reproduced against a real `--depth=1` clone over `file://`: `at_base` returned `None` both for "no ancestor carries this file" and for "the history that would say so cannot be read", and `_before` read that single `None` as a first introduction -- the path where no comparison runs. A `grow` list that had LOST a required command passed with no failures. `Unknown` is the third answer, returned when `rev-list` fails, when an ancestor's copy is unparsable, and when `rev-parse --is-shallow-repository` reports truncation. Six mutants killed.

**Whose fault an unresolved reference is.** `load_index` returns an empty table for a MISSING index file, so every reference read as "not in `docs/canon/index/<source>.tsv`" -- indistinguishable from an invented key, and filed as `invalid_reference`, a sentence addressed to the contributor. On a checkout without the corpus built, every citation in every body became the author's fault. `_require_readable_index` raises instead, so the category is `error` and the status is 2; a key the run could look up and did not find stays actionable. Five mutants killed.

**The checker's exit status was never read.** A status is what a killed or wedged process cannot fake; a category is what the note is worded from. A checker killed by a signal after printing `satisfied` was read as a pass, and through `main()` that RESOLVED a standing warning. `_validated` now enforces the status/category pair, the result's shape, each diagnostic's field types, and refuses an actionable verdict with nothing behind it; the child has a bounded timeout, and a launch failure or a timeout is the same "did not finish" answer.

**The freshness check ran before the slowest thing this does.** The body was compared once, before a comment walk of up to a hundred pages. `body_moved` is asked again immediately before any write and again after the stale-note recovery re-lists. Driven through the real paging method, with a full first page and the managed note on page two. Twelve bot mutants killed in total.

**The value branch asked whether the index exists and never whether it parses.** The third review's
one nonblocking finding. `_require_readable_index` asks two questions on the key path -- is the file
there, does it parse -- and asked only the first on the value path, so a `<source>.values.tsv` row
with the wrong field count raised the base `CanonError` past `CanonIndexUnavailable`. Not a
regression: `check_text`'s outer except still catches it, so the category and the status were the
same either way and only the message changed. Closed anyway, because two branches of one function
reaching the same answer by different routes is what that function exists to remove. Two cases, not
one -- a value citation against the committed value index, then the same body with it wrecked.

Also corrected: `pr-policy.yml` said closing and reopening re-runs this workflow only -- `reopened` is a **code** event to `classify-pull-request-event.py`, so it re-runs the full validation, and a body edit is the cheap recovery. `docs/roadmap/README.md`'s ADR-013 row assigned #301's own E1/E2 scope to #955 and #308; #955 holds the deferred batch, #308 owns scheduling, and the row now reports the closed-state/unchecked-rows mismatch instead of moving the work. `docs/canon/README.md` says outright that an issue's citation is asked for rather than required, and that a maintainer can map an observation to a row.


Scripts/check-canon-citations.py
```diff
@@ -1273,46 +1138,57 @@ def check_text(path: str, changed_paths=None, *, require_changed: bool = False)
...
+            f"  reason -- outside any code block or HTML comment. See docs/canon/README.md."))])
+
+    findings = []
     for ref_text in references:
         try:
             ref = canon.CanonRef.parse(ref_text)
             canon.resolve_offline(ref)
         except canon.CanonError as exc:
-            failures.append(f"{path}: {exc}")
+            findings.append((INVALID_REFERENCE, f"{label}: {exc}"))
```
MongLong0214, line 1256: **[R1 / P2] Do not classify a corpus-loader failure as an invalid author citation.**

`load_index()` can raise the same `canon.CanonError` for malformed repository TSV data; missing index data also reaches `CanonResolveError`. This broad catch turns those failures into `INVALID_REFERENCE`, and the later broad catch skips them, so the outer tooling-error handler never sees them. A valid reference plus a broken index therefore becomes `actionable`, despite the bot's new error-vs-author distinction.

Please separate unreadable/malformed required corpus data from a bad reference in a healthy index. Add controls for a valid pinned quote with normal/missing/malformed index, plus an unknown key in a healthy index. The tooling cases must produce `error` and no author-accusation/resolved note. The current missing-manifest/no-fact-body test reaches a different branch and does not cover this. See R1 in the review summary for the bounded fix and verification scope.

MongLong0214 (commented):
## Production-readiness review — changes needed before merge

Reviewed head **`2765bb911424218af4269e5a8a6646cf498781b8`** against base **`06e5d37c8bd8fb9db62431618b0d386595f63ba9`**, the current #951 implementation contract, all 27 changed-file diffs, and the affected production/test call paths. The four policy JSON files are content-preserving renames; the substantive risk is their executing owner and history comparison, not a changed allowance in those files.

**Recommendation: do not merge this head yet.** The direction is good, but four correctness gaps remain in the new error/history/comment paths. There is also one narrow ownership correction needed in the roadmap change. These are bounded fixes to this PR, not a request for another framework, a broader ADR audit, stronger release governance, or unrelated runtime work.

### Evidence and verification boundary

I independently exercised isolated regressions against exact copies of the new production modules. Their Git blob IDs matched the connector-fetched originals:

- `Scripts/ratchet.py`: `fb5ba11a0416108bd57ee26a92b617ec89aa1842`
- `Scripts/canon_issue_bot.py`: `903dd93815ae8879ee3a39cf1fbe3f84b0621518`

The history reproduction used real temporary Git repositories and a real `git clone --depth=1 file://...`. The bot reproductions used the unmodified production module with injected subprocess/API responses; they are **not live GitHub bot executions**. The corpus diagnostic finding below is confirmed by the checker/loader exception paths, not by claiming a full repository checker run here. I did not rerun the complete repository suite, Swift tests, or live Logic. The PR's reported test results remain author-provided evidence.

At the latest CI read, run [35587823299](https://github.com/MongLong0214/logic-pro-mcp/actions/runs/35587823299) had `classify`, shell `guards`, and `formula` successful; `test` was running and `compile` queued. The standalone Python guard stage and coverage inside `test` were still pending. [PR policy 35587823469](https://github.com/MongLong0214/logic-pro-mcp/actions/runs/35587823469) and [Maintenance 35587823317](https://github.com/MongLong0214/logic-pro-mcp/actions/runs/35587823317) were successful. Therefore this review does not call the full CI green.

## Required corrections

| ID | Severity | Finding | Required bounded result |
|---|---|---|---|
| R1 | P2 | A referenced corpus/index failure becomes an author-facing `invalid_reference` | Distinguish an invalid reference in a healthy index from inability to load/evaluate that index |
| R2 | P2 | The bot accepts `satisfied` JSON from a failed/killed checker and incompletely validates the report | Validate process outcome and report structure before any resolved/no-problem verdict |
| R3 | P2 | A readable merge base with unreadable/shallow policy history is treated as genuine first introduction | Preserve an explicit unknown-history failure instead of bootstrapping away the comparison |
| R4 | P2 | Body freshness is checked before comment pagination/recovery, not at the publishing boundary | Recheck the relevant body after discovery and before a body-dependent write; bounded recovery must do the same |
| R5 | P2 | The changed ADR-013 row assigns the remaining EQ reader/writer work to owners that do not own it | Preserve #301's E1/E2 ownership and explicitly reconcile its closed-state/body mismatch without silently transferring scope |

### R1 — error classification still depends on which body branch was taken

In `diagnose_text`, the reference loop catches **every `canon.CanonError`** and records `INVALID_REFERENCE`. The second loop also catches every such error and continues. However, `logic_canon.load_index()` raises that same base error for a malformed TSV. A missing index also becomes an empty lookup and then `CanonResolveError`. Thus a valid reference whose repository-side index is unavailable can end up as `actionable / invalid_reference`, and the bot tells the author to repair their citation. The outer `check_text` error handler cannot classify an exception that the inner loop already consumed.

The new missing-manifest test uses a no-fact body. That exercises the other branch through `load_manifest`; it does not protect a body containing a legitimate citation.

**Fix:** separate index/corpus availability/decoding failures from genuine syntax/unpinned-row/value problems. Use the existing loader with a small typed distinction or explicit prerequisite validation, not a second citation implementation. Preserve actionable errors for an actually bad reference against a readable index.

**Regression:** in an isolated fixture, a known-valid pinned reference and quote pass with the normal index; with that source's index missing or malformed they produce `category=error` and a nonzero tooling result; an unknown key in the intact index remains actionable. Through the bot, the tooling case must not create an author-error note or resolve an existing warning. Include malformed manifest/report handling without disguising exceptions as an empty body.

### R2 — process failure can resolve an existing warning

`evaluate()` never checks `done.returncode`; it only checks the JSON category. I injected the following child results into the actual production function:

```text
returncode=0,  valid satisfied report  -> satisfied  (control)
returncode=2,  valid satisfied report  -> satisfied  (incorrect)
returncode=-9, valid satisfied report  -> satisfied  (incorrect)
{"category":"satisfied"} without diagnostics -> satisfied
[] as stdout JSON -> uncaught AttributeError
```

Driving the failed-process case through actual `main()` with an existing owned note produced an update of that note to **`resolved`**. This violates #951's explicit requirement that a crashed/failed/malformed evaluation must not become success. The current test called a crash test only injects empty/non-JSON/unknown-category output with a successful return code, so it misses this condition.

**Fix:** enforce the checker's documented exit/category contract and minimally validate the top-level object and required diagnostic fields/types. A signal termination, inconsistent category/exit pair, invalid structure, launch failure or timeout is `error`. Add an explicit bounded child timeout; do not rely on an entire workflow eventually being killed to implement an unknown-result transition. An ordinary `TemporaryDirectory` also suffices to clean up the temporary input; no job service or new protocol is needed.

There is a related diagnostics omission in the same flow: a legitimate structured `error` normally carries its cause in stdout `diagnostics`, with empty stderr. `render(error, ...)` intentionally emits a generic note, and `main()` never logs those structured findings. It only prints “see the log above”, where the cause is absent. I reproduced that with a sentinel error message. Log a bounded, safely escaped diagnostic code/cause for maintainers, without copying raw private input or interpreting it as an Actions command.

**Regression:** test matching normal category/exit pairs, nonzero and signal-terminated `satisfied`, missing/wrong-type diagnostics, list/null JSON, timeout/launch failure, and a valid structured error with empty stderr. Assert no erroneous resolved note and an actionable maintainer log of the tooling cause.

### R3 — the shallow-history test does not model a shallow repository

`History.at_base()` returns `None` both for proven absence and for history it explicitly says is truncated. `_before()` then tries the legacy path and, when both calls return `None`, labels the situation a genuine first introduction. A `grow` requirement is skipped. `base_or_refuse()` does not prevent this when `git merge-base` succeeds, which is perfectly possible in a shallow clone.

The real-Git reproduction was:

1. Commit an old-path policy with `required_commands = ["keep", "must-not-disappear"]`.
2. Commit its deletion while leaving another repository file.
3. In a full-history working copy, introduce the new path with the original contents: the comparison passes (control).
4. Remove `must-not-disappear` in that new path: full history correctly rejects it via the last ancestor.
5. Clone the same repository with `--depth=1` over `file://`, retaining a readable `origin/main` merge base, and introduce the shortened new-path policy.
6. With `CI=true`, the actual helper reports truncated history, then “genuine first introduction”, and returns **no failures**.

```text
FULL HISTORY + EQUIVALENT RESTORE: []
FULL HISTORY + REMOVED REQUIREMENT: [ ... may only GROW ... ]
SHALLOW: true
MERGE BASE EXISTS: True
SHALLOW HISTORY + REMOVED REQUIREMENT: []
```

The added `test_a_shallow_checkout_under_ci_fails_rather_than_degrading` deletes `.git` altogether. That tests *no readable merge base*, not *a readable base with unavailable earlier policy history*. The default full-depth workflows reduce exposure, but do not satisfy the extracted helper's declared M05 failure behavior. This is not a claim that the guard defeats a malicious committer who can rewrite it.

**Fix:** distinguish found policy, positively established first introduction, and unknown/failed history. Unknown history must record a failure, not grant bootstrap treatment. Apply the distinction to legacy lookup as well; do not silently turn failed Git/JSON reads into absence. Retain legitimate first introductions and valid relocations.

**Regression:** add the real shallow-clone case above beside the full-history equivalent and rejected-change controls. Keep the no-`.git` test, but name it for what it actually covers.

### R4 — freshness can become stale while finding the comment

`main()` re-reads the body, then performs the comment listing (potentially many pages), decides, and writes. The update-recovery path performs another listing and another write without rechecking the body. I reproduced a body change **inside comment discovery, after both existing body reads**, and the actual function still updated the note to the old actionable result over a now-valid declaration.

**Fix:** place a body-content check after comment discovery and immediately before the primary create/update; repeat it after stale-ID recovery. If the body changed, abandon that obsolete verdict or perform the one already-authorized bounded refresh. No polling loop, distributed lock or claim of an atomic cross-object GitHub transaction is needed. Concurrency cancellation remains useful, but is not a replacement for validating the snapshot at this boundary.

**Regression:** advance the body during a later comments page and during update recovery, not only during the initial evaluator call. Both should avoid publishing the old result. Exercise the actual pagination method with a full first page and an owned note on page two; handing an already-complete list to `FakeGitHub.comments()` does not test pagination.

### R5 — keep the EQ implementation owner instead of making the roadmap absorb it

The new `docs/roadmap/README.md` ADR-013 row correctly notices that GitHub closed #301 again. I also re-read #301: it is closed, but its current body still owns **E1 exact-instance EQ state reading and E2 bounded writer completion**. #955 owns generic/batch/reset/compensation extensions, and #308 is the scheduling index, not an implementation destination for E1/E2.

The added wording that the remainder is owned by “#955 and #308” therefore introduces a scope transfer that neither receiving contract supplies. A closed-state observation is not evidence that the selected EQ work was completed or transferred.

**Fix only this touched row and its explicit handoff:** retain #301 as the E1/E2 owner, #955 as the deferred batch owner, and #308 as scheduling. Flag/reconcile the issue's state/body mismatch through the authorized issue-maintenance path; reopening or a real requirement-preserving transfer is an administrative decision, not completion. Do not implement EQ in this Canon PR or expand this change into #956's whole-document migration. A status guard should not be made green by assigning work to a navigation issue.

## Small documentation corrections, not new workstreams

- `.github/workflows/pr-policy.yml` says closing/reopening reruns only policy and never restarts macOS jobs. The current classifier explicitly treats `reopened` as a **code** event. Prefer a body edit for metadata-only recovery; describe reopening as a full-validation trigger. Do not change the correct classifier to match this comment.
- Correct the PR's moved-policy table: CI-GATE's ratchets cover `required_commands` and `not_required`; `workflows` is structurally checked, not that same ratchet entry. The blind-test file uses `guards`, not `tests`; the guard-debt files are maps whose keys are compared. This matters for the next agent reading the handoff.
- In the new Canon README entry, explicitly preserve optional initial issue evidence and maintainer help instead of presenting one evidence sentence as mandatory to submit every issue. A placeholder is not a submitted default value; do not require extra form machinery based on that mistaken test comment.

## What should remain as implemented

The corrected form fields, single shared checker, explicit non-success tooling category, marker-plus-bot ownership, deterministic comment contents, content-preserving policy moves, existing executing owners, and reuse of current guards are appropriate. Preserve the #952 metadata/code separation, protected-path rules, coverage floors and token boundaries. Do not replace this with a new service or broaden exemptions to make a test green.

Using `Refs #951` and keeping **I02 post-deployment observation open** is correct. A live issue workflow running from the default branch cannot be honestly demonstrated by a local renderer. Once these bounded fixes and existing required CI pass, merging the implementation need not wait for an impossible pre-merge default-branch demonstration; #951 should remain open until the controlled deployed create/update/resolve/recurrence observation is recorded.

## Exit criteria for the next review

1. Fix R1–R4 in their existing owners and add the specified focused positive/negative regressions.
2. Correct the touched roadmap ownership and the small recovery/handoff wording above without reopening the entire ADR design.
3. Run the affected Canon diagnostics/bot/forms/history/CI-integrity suites, then the existing complete required CI on the new head. Record actual results, not copied previous counts.
4. Keep source merge, live bot deployment verification, and #951 completion separate. No stable release, issue auto-closure, lower coverage, new permission, or unrelated runtime change is required by this review.

This review is published as a comment, not an approval or merge action. The findings are anchored to the exact head above; later changes need a focused delta check rather than a new full-programme specification.

Scripts/canon_issue_bot.py
```diff
@@ -0,0 +1,337 @@
...
+    path = os.path.join(workdir, "issue-body.md")
+    with open(path, "w", encoding="utf-8") as handle:
+        handle.write(body)
+    done = subprocess.run([sys.executable, CHECKER, "--text", path, "--format", "json"],
+                          capture_output=True, text=True, cwd=REPO)
+    try:
+        result = json.loads(done.stdout)
+        # A category this version does not know is an evaluation that did not finish, not a pass.
+        if result.get("category") not in ("satisfied", "actionable", "error"):
+            raise ValueError(f"unknown category {result.get('category')!r}")
```
MongLong0214: **[R2 / P2] Validate the child process and report before accepting `satisfied`.**

`done.returncode` is ignored. Against this exact module, injecting a valid `satisfied` JSON with return code `2` or `-9` still returned satisfied; through `main()` it updated an existing owned warning to `resolved`. `{"category":"satisfied"}` without diagnostics is also accepted, and list JSON raises an uncaught `AttributeError`.

Enforce the documented category/exit mapping and minimal required object/diagnostic types; treat launch failure, signal termination, timeout and inconsistent output as tooling errors. Give the subprocess a bounded timeout. Add those cases to the real evaluate/main tests, not only invalid JSON with returncode=0. Also log sanitized structured error diagnostics: when the checker emits JSON error data and empty stderr, the current generic 'see the log above' message loses the actual cause. None of these fixes needs a new reporting framework.

Scripts/ratchet.py
```diff
@@ -0,0 +1,299 @@
...
+        for sha in (history or "").split():
+            prior = self.show_json(sha, path)
+            if prior is not None:
+                self.note(f"{path} is absent at the merge base {base[:8]}; ratcheted against "
+                          f"{sha[:8]}, the last ancestor carrying it.")
+                return prior
+        if self.git("rev-parse", "--is-shallow-repository") == "true":
+            self.note(f"{path}: history is truncated (shallow clone), so 'no ancestor carries it' "
+                      f"is not a reading anyone can trust. Check out with fetch-depth: 0.")
+            return None
```
MongLong0214: **[R3 / P2] Unknown history must not share the `None` used for confirmed first introduction.**

I reproduced this with a real depth-1 `file://` clone and `CI=true`: the merge base is readable, but the policy existed before a deletion outside the shallow boundary. Both new and legacy lookups return this `None`; `_before()` then reports 'genuine first introduction' and skips a `grow` requirement even when a previously required command was removed. The full-history control correctly rejects the same removal.

Propagate unknown/failed history as a failure, separately from proved absence, through the legacy fallback as well. The newly added test named for shallow checkout deletes `.git`, so it only tests missing merge-base handling and misses this case. Add an actual shallow clone with a valid merge base, full-history equivalent-relocation control, and full-history rejected-removal control. Preserve genuine bootstrap cases; no new policy engine is needed.

Scripts/canon_issue_bot.py
```diff
@@ -0,0 +1,337 @@
...
+    comments = api.comments(number)
+    note, extra = active_note(comments, expected_login)
+    action, why = decide(result["category"], rendered, note)
+    print(f"category={result['category']} action={action} ({why})")
+
+    if result["category"] == "error":
+        print("::warning::the Canon check could not evaluate this issue body; see the log above")
+
+    if action == "create":
+        api.create_comment(number, rendered)
```
MongLong0214, line 408: **[R4 / P2] Recheck the body after comment discovery, immediately before publishing.**

The existing freshness check runs before `api.comments()`, which can paginate many requests. If the author repairs the body during that interval, this code still publishes the old finding. I reproduced that interleaving through the actual `main()` with the API seam: both early body reads agreed, discovery advanced the body, and the bot edited the note to the stale actionable result. The retry path at lines 322–325 has the same gap.

Recheck the relevant body content after discovery and before create/update, including after stale-ID recovery; abandon or boundedly refresh an obsolete result. Test edits during later-page discovery and recovery, not only inside initial evaluation. This closes the avoidable window without promising an atomic GitHub transaction or introducing locking/polling infrastructure.

docs/roadmap/README.md
```diff
@@ -58,7 +58,7 @@ v3.15.0 published, 207 commits since · ADR-019 landed · de-DE joined the local
...
...he live instance. The write is a walk, not a set — the sliders expose only `AXIncrement`/`AXDecrement`, and a direct `AXUIElementSetAttributeValue` returns success while landing a different value. Original wording: Measured 2026-08-30, the recorded wall is not where it was said to be: Channel EQ's native view exposes 8 named band enables and 26 `AXSlider`s, including all 24 named band parameters; every slider reports `settable=yes` and carries `AXValueDescription` in Hz/dB. **No write was attempted, and AX settability lies** — what is established is that the controls are addressable and readable. The opacity claim came from a census that instantiates the AU *outside* Logic, which ADR-018's Important Boundary says cannot speak for the live instance. The ADR's C4 Decision was revised on the issue the same day **Reopened 2026-09-21 by the owner: a real single-parameter feature shipped, the whole ADR did not.** The remaining scope is truthful same-instance EQ state readback and the bounded single-operation contract, with `logic_plugins.set_eq_band_verified` preserved; the earlier C4 setup/codec/port/doctor instructions are superseded and must not be implemented. #308 owns scheduling. |
```
MongLong0214: **[R5 / P2] A closed-state observation does not transfer #301's E1/E2 work to #955/#308.**

The new sentence assigns the remaining scope to '#955 and #308'. Current #301 still owns exact-instance EQ state reading and bounded existing writer completion. #955 explicitly owns the deferred generic/batch/reset/compensation scope and reuses the EQ reader; #308 is the scheduling index, not its implementation owner.

Keep those owners explicit in this touched row. The actual closed-state/body mismatch should be reconciled by an authorized state correction or an explicit requirement-preserving transfer, not by inferring delivery or assigning implementation to the index. Do not implement EQ or the entire #956 migration in this Canon PR. A narrow accurate state/owner note is sufficient here.

.github/workflows/pr-policy.yml
```diff
@@ -27,9 +27,23 @@ name: PR policy
...
+# ------------------------------------------------
+# A required context is satisfied by a check run on the head commit, and this workflow only
+# produces one when it is TRIGGERED. A pull request whose head predates 2026-09-21 therefore has
+# no `pr-policy` run at all: branch protection reports the context missing and the merge waits --
+# correctly, because nothing has read that body under the rule now in force.
+#
+# The recovery is one event, and none of them is code churn: edit the description, or close and
+# reopen the pull request. Both are in `types:` below, both re-run THIS workflow only, and neither
+# restarts a macOS job or disturbs a code verdict already on the commit. A rebase or an empty
+# commit also works and is the expensive way round -- it moves the head, so the code jobs run
```
MongLong0214: **[P3 / documentation] Reopening does trigger code validation.**

The existing classifier returns `CODE` for every non-`edited` action, including `reopened`, and `ci.yml` subscribes to that event. Thus close/reopen is not a policy-only recovery and can restart full macOS validation. A title/body edit runs policy plus the lightweight classifier without rerunning heavy code jobs.

Correct this comment rather than change the intentional classifier behavior. Also avoid inferring missing policy evidence solely from a head commit's date: an older commit may already have a later valid policy run.

Issue #950, opened 2026-09-21
CI maintenance contract — implemented by #952; preserve contributor and required-check behavior


MongLong0214:
# CI contributor experience and metadata handling — implemented maintenance contract

> **Execution SSOT r3 — 2026-09-21. CLOSED / implementation merged in #952.** This is no longer the next implementation task. #951 owns the remaining Canon form/bot/CI-policy separation. #308 owns programme scheduling; #956 owns navigation/milestone reconciliation. Do not rerun the original additive migration or rebuild the classifier merely because this issue is searchable.

This issue states no fact about Logic. It records this repository's CI event handling, contributor documentation, required-check configuration and their verification limits; it makes no new DAW behavior claim.

## 1. Applied state and evidence boundary

[PR #952](https://github.com/MongLong0214/logic-pro-mcp/pull/952) merged as `06e5d37c8bd8fb9db62431618b0d386595f63ba9` on 2026-09-21. Main was re-read at that commit for this reconciliation. Its ten implementation commits and actual run references remain the implementation history; this issue revision did not rerun them.

Ruleset `19060230` was independently re-read on 2026-09-21: active on the default branch, required contexts `build`, `compile`, `test`, `pr-policy`, each integration ID `15368`; strict freshness true and bypass actors empty. That confirms the inspected ruleset, not an assertion that all possible inherited rules were enumerated. No ruleset was changed by this editorial update.

The old filing observations at `eadb5f4`—especially `pr-policy is not required`, a pending #949 and a duplicate body-check job—are historical, not the present instructions. #949 and #952 have landed.

## 2. Actual delivered design

- `Scripts/classify-pull-request-event.py` classifies positively identified metadata-only edits; unknown/incomplete/malformed payloads conservatively require code validation.
- `ci.yml` retains the `edited` subscription because base retargeting is also an edited event. Code and metadata cancellation domains are separate.
- Real code validation emits canonical required job names. Metadata-only skipped jobs must not emit `compile`, `test` or `build` in a form that can replace a failed/pending code verdict.
- The measured skipped-job behavior differs from the original explanatory sketch: GitHub emitted the raw name expression for skipped jobs, while executing jobs used the evaluated canonical name. Preserve the invariant that metadata-only names are noncanonical; do not promise a particular decorative display name.
- `pr-policy.yml` is independently required and owns changed-file/body Canon validation. Code guards remain in the required code path; maintenance remains outside the merge gate.
[...]

PR #962, opened 2026-09-21
build(package): expose LogicProMCP target as LogicProMCPKit library product (#944)

### LINKED ISSUE #944: Feature: expose a `.library` product so other Swift apps can embed the AX readback layer
### Preflight

- [x] I searched existing issues for related work.

### Priority

P1 - unlocks a high-value Logic Pro workflow

### Primary area

setup-release

### Problem

My name is Rico, I'm working on a native AU3 plugin for Logic Pro that manages level relationships through a Mackie Control bridge.

Basically you guys solved the one problem I couldn't. Reading Logic's live state through the accessibility tree. I've been through the ugly parts of trying to infer it from meters and MCU echoes, and it doesn't hold. Your AX readback does.

So I'd rather depend on this project than reimplement it, and keep taking your updates every time Logic shifts under everyone.

Today this fails because `Package.swift` only declares the `LogicProMCP` executable product. The `LogicProMCP` target can't be pulled in as a Swift package dependency, so the only way in is running the server as a separate process and talking MCP to it. For an in-process plugin that's a second process to sign, launch, and keep alive, with IPC latency on every read.

### Proposed outcome

Add one product declaration next to the executable:

`.library(name: "LogicProMCPKit", targets: ["LogicProMCP"])`

(Edited: the product needs a name distinct from the executable. SwiftPM accepts a duplicate name but reports "ignoring duplicate product" and silently drops the library. See comment below.)

Nothing about routing, safety, or the tool surface changes. Internals stay internal. Consumers only get what you choose to mark `public`.

The client should be able to `import LogicProMCP` and use the existing AX readback in-process. What I'd do with it: walk the mixer strips and insert slots to find my own plugin's instances, then read a `kAXIdentifier` my plugin's view advertises. That makes the link between a surface fader and a plugin instance deterministic instead of guessed.

I have a small read-only facade on top of that already working on my side. Happy to contribute it as a follow-up if you'd want it in the tree. That's probably a conversation for another day.

Two questions:

- Is a library product something you're open to, or do you want the executable to stay the only door?
- Any part of the target you'd want kept deliberately non-public?

The branch is ready. I'll open the PR on your word, or shoot back a few lines and I'll adjust.

### Safety and contract notes

- Mutating operation? No. Manifest-only change; no runtime behavior added or altered.
- Required explicit target? N/A. No new tool or MCP surface.
- Readback source: unchanged. Consumers use the existing AX readback; the library exposes only what is already marked public.
- Failure/uncertain state: unchanged. Embedded callers inherit the existing fail-closed semantics.
- Destructive confirmation: N/A.

### Done when

- [ ] `Package.swift` declares a `.library` product named `LogicProMCPKit` for the `LogicProMCP` target alongside the executable.
- [ ] `swift package describe` lists the library product and the build emits no "ignoring duplicate product" warning.
- [ ] A consumer package can `import LogicProMCP` and build with `swift build`.
- [ ] Existing test suite passes unchanged.
- [ ] README or docs note the library product and its public surface.
- [ ] No live Logic evidence required: the change does not write to Logic.

### Logic canon

states no fact about Logic. This issue proposes a Package.swift product declaration only; it asserts nothing about Logic Pro behavior, UI, or data.

Brujo369x:
## Summary

- Adds one product declaration to `Package.swift`, `.library(name: "LogicProMCPKit", targets: ["LogicProMCP"])`, so a Swift application can link the AX/state/channel layer as a SwiftPM dependency instead of running the server as a separate stdio process.
- The product is named `LogicProMCPKit` because a library sharing the executable's name is reported by SwiftPM as "ignoring duplicate product" and silently dropped (measured on this branch; the issue text was corrected accordingly).
- Adds a README section, "Using LogicProMCP as a library", with the dependency snippet, the import, and the exact public surface reachable today: `PluginInspector` and its data types, plus the Library inventory data model. The AX readers, channels, state cache, and dispatchers stay `internal`; no access level was changed. Every public entry point that touches a window or menu takes a caller-supplied closure runtime, so the library exposes nothing that reaches Logic on its own.
- The read-only facade discussed in #944 is deliberately not in this PR and will come as its own issue.

## Related issue

- Closes #944

## Verification

```text
swift build
  Build complete, no warnings about products.

swift package describe --type json
  products: LogicProMCP (executable), LogicProMCPKit (library, automatic)
  No "ignoring duplicate product" in describe, build, or test output.

Separate consumer package (.package(path:) to this branch, .product(name: "LogicProMCPKit", package: "logic-pro-mcp")),
main.swift: import LogicProMCP; PluginInspector.decodeAUVersion(0x00010203)
  swift build: Build complete! (13.31 sec), exit 0
  output: LogicProMCPKit linked; PluginInspector.decodeAUVersion(0x00010203) = 1.2.3

swift test --no-parallel   (on the manifest commit ce6a59d8)
  4657 passed, 0 failed, 7 skipped
```

Not verified (and why):

- The serial suite was not re-run on the README-only commit c4170010. On this Mac two CoreMIDI smoke tests currently fail with "Could not read CoreMIDI endpoint name (-50)" on an untouched checkout of main as well, from a stale endpoint in the local MIDIServer, so a rerun on the docs commit would not add signal. The commit touches README.md only.

## Logic evidence

This change states no fact about Logic: it is a Package.swift product declaration and README documentation only, and it edits no Logic-facing path.

## Risks and limitations

- User-visible impact: none for server users. A new SwiftPM product is available to consumers; the executable, tools, and resources are unchanged.
- What this change does not do: it does not change any access level, expose the AX readers, add a tool or resource, or alter runtime behavior.
- Known limitation: the public surface reachable through the library is small (PluginInspector and data types). Consumers who need Mixer, track, or plug-in slot readback will need the separate facade change tracked from #944.


MongLong0214:
Reviewed. The manifest change is right and the README's public-surface enumeration is accurate — I
checked it rather than read it. One thing was broken and I corrected it on this branch rather than
send you a round trip; `maintainerCanModify` was on. Details so you can disagree:

**The manifest.** `swift package describe --type json` on your branch reports exactly two products,
`LogicProMCP (executable → LogicProMCPCLI)` and `LogicProMCPKit (library, automatic → LogicProMCP)`,
with empty stderr. No "ignoring duplicate product". Your reason for the `Kit` suffix holds.

**The public-surface list.** This is the claim most likely to outrun its measurement, so I measured
it. Every `public` top-level declaration in `Sources/LogicProMCP` is 22 names; five of them are not
in your list — `Miss`, `ScanOrchestration`, `PathResolution`, `PathRuntime`, `ScanResult`. All five
are nested inside `enum LibraryAccessor` (LibraryAccessor.swift:213), which is declared without
`public`, so Swift caps them at internal and a consumer cannot reach them. Your section already says
that in prose. The ten `PluginInspector` entry points you list are exactly the file's
`public static func` declarations — grep returns those ten and nothing else. Nothing overstated.

**What I changed (`0f357525`).** The dependency snippet is the one thing a reader copies, and it
could not work:

```
.package(url: "https://github.com/MongLong0214/logic-pro-mcp", from: "3.16.0"),
```

`git show v3.16.0:Package.swift` carries `LogicProMCP` and `trusted-verifier` and no library, and
`from:` resolves the highest tag in `[3.16.0, 4.0.0)` — v3.16.0 today. I wrote a consumer package
exactly as the section said and built it:

```
error: 'consumer': product 'LogicProMCPKit' required by package 'consumer' target 'Consumer'
not found in package 'logic-pro-mcp'.
```

The snippet now uses `branch: "main"`, which is true from the moment this merges, and the section
says why the version form fails, quotes that error, and states that `from: "<version>"` is the
preferred form once a release carries the product. I did not write `from: "3.17.0"`: there is no
such tag and no commitment that the next release is a minor rather than a major, and a snippet
naming an unreleased version is a promise this repository has not made.

I also merged `main` into the branch — it was behind, and the required checks are strict.

One note on this repository rather than on your change: workflow runs on a fork wait for a
maintainer to approve them, which is why this sat with no checks. That is the barrier #950 was
about, and I have approved them now.

Thank you — this is a well-evidenced contribution, and the "not verified, and why" section is the
part I wish more pull requests had.

MongLong0214:
Two more commits on your branch (`ab514acf`), one found by this PR's first CI run and one by an
independent review of its head. Both are on me, not on you — neither was visible from what you
could run.

**1. `Package.resolved` had to move.** The `compile` job ends with
`git diff --exit-code Package.resolved`, and it failed. The reason is your change, and it is
interesting: adding the library product makes CI's toolchain (Xcode 16.4) resolve **25 pins where
the committed file has 9**. The 16 additions are all transitive through `swift-sdk` —
`async-http-client`, `swift-certificates`, `swift-crypto`, the NIO extras, `swift-service-lifecycle`
and the rest — and `originHash` changes with them.

I did not reproduce it locally, because I could not: on Swift 6.2.4 the same manifest resolves to
the same 9 pins as before. So the file committed here is the runner's own output, lifted from the
job log and applied — `git hash-object Package.resolved` gives `e347b333e92bed9df34329b39f97ded9224c68ef`,
which is the blob id in that run's own `diff --git` index line.

Why 16.4 resolves more is **not** established, and I have not written a mechanism into the tree.
What is measured is the part that decides whether the file is safe to commit: Swift 6.2.4 adds none
of the 16 to the 9-pin file and removes none of them from the 25-pin file, so the 25-pin file is a
fixed point under both toolchains and the 9-pin file is a fixed point under only one.

**2. The roadmap row.** `docs/roadmap/README.md` is this repository's source of truth for what is
open, and its `#944` row still read "NOT TRIAGED … nothing has decided that it wants one" — while
this PR is the deciding. The file's own rule is that the closing PR updates the row, so the row now
says closed and records what settles the ADR-class question it raised: the surface this repository
takes on was measured, not promised.

Checks run locally on the new head: `roadmap-table-matches-github.py --pr 962` agrees with GitHub
(101 rows, 28 open), `run-repo-guards.py` 90/90, `check-canon-citations.py --changed` resolves 162
references.

Nothing in your three files changed in this push beyond the README correction I described above.

docs/roadmap/README.md
```diff
@@ -146,7 +146,7 @@ Also open, outside the ADR set:
...
...an reach are `PluginInspector`'s ten entry points and its data types, because everything in `LibraryAccessor` is nested inside an `internal` enum and Swift caps a nested `public` at its container, so the obligation taken on is the one already documented. Two things the pull request found rather than assumed: the README's `from: "3.16.0"` form resolves v3.16.0, which predates the product, and a consumer package written that way fails with `product 'LogicProMCPKit' ... not found in package 'logic-pro-mcp'` (measured), so the documented form is `branch: "main"` until a release carries it; and adding the library product changes what `swift build -c release` resolves on CI's Xcode 16.4 toolchain from 9 pins to 25 — `async-http-client`, `swift-certificates`, the NIO extras and 13 more, all transitive through `swift-sdk` — so `Package.resolved` is updated in the same pull request. Why that toolchain adds them is not established here; what is measured is that the local Swift 6.2.4 toolchain neither adds them nor prunes them away once present, so the committed 25-pin file is stable under both. The small read-only facade the requester offered is not part of this and has no issue of its own |
```
MongLong0214: **[P2 / SSOT] This undercounts the public API that the new library product makes reachable.**

`LibraryAccessor.swift` has four top-level public declarations *before* the internal `enum LibraryAccessor`: `LibraryNodeKind`, `LibraryNode`, `TreeProbe`, and `LibraryRoot`. Only declarations nested inside the internal enum are capped by that container.

The README in this same PR already lists those four correctly. Because this row is recording the API-stability obligation #944 creates, please make the SSOT say the same thing: PluginInspector + its public companion types **and** those four top-level Library types are externally reachable; the Library scanner/selection implementation remains internal.

README.md
```diff
@@ -205,6 +205,51 @@ bash <(curl -fsSL https://raw.githubusercontent.com/MongLong0214/logic-pro-mcp/v
 
 See [SECURITY.md §Installer trust model](SECURITY.md#installer-trust-model) for the trust tiers and threat model.
 
+## Using LogicProMCP as a library
+
+The package also exposes the `LogicProMCP` target as a library product named `LogicProMCPKit` (#944), so a Swift application can link the same code the server runs instead of driving the server as a separate stdio process. This is a manifest-only addition: it adds no MCP tool or resource, and it changes no runtime behavior of the server or its executables.
```
MongLong0214: **[P3 / accuracy] “manifest-only” stopped being true after the necessary lockfile/roadmap correction.**

The important behavioral statement is still right: no MCP operation, source access level, executable behavior, or Logic runtime path changes. But the current PR now also changes `Package.resolved`, CHANGELOG and the roadmap.

Please phrase this as a **package-product/API-surface-only** addition (or simply “no runtime behavior/access-level change”) rather than “manifest-only,” and mirror that cleanup in the changelog/PR body so the final receipt describes the diff that will actually merge.

MongLong0214 (changes requested):
## Production-readiness review — one SSOT correction before merge

Reviewed the **current head `ab514acfea2b354eccdead39bfde8aa96826d4b5`**, not the earlier head that failed CI. I checked the five-file diff, #944's accepted scope, the actual public declarations in `PluginInspector.swift` and `LibraryAccessor.swift`, the consumer-facing documentation, and the current Actions state.

The core change is appropriately small: the distinct SwiftPM product name is correct, the module remains `LogicProMCP`, no access level changed, the requested mixer/slot facade was correctly kept out of this PR, and the README's own public-surface list matches the source. The separate consumer build is exactly the kind of verification #944 asked for.

The first CI run also found a real repository contract issue rather than a library-build failure: `swift build -c release` succeeded, then the compile job failed because the manifest edit caused SwiftPM to rewrite `Package.resolved`. The new head commits the exact expanded lockfile instead of weakening that check. That is the right direction; the replacement CI must still finish successfully before merge.

### Required: correct the #944 roadmap SSOT row

The new roadmap row undercounts the API contract this PR actually exposes. It says the only reachable public declarations are `PluginInspector` and its types because “everything in `LibraryAccessor` is nested inside an internal enum.” That is not the source layout: `LibraryNodeKind`, `LibraryNode`, `TreeProbe`, and `LibraryRoot` are top-level `public` declarations **before** the internal `LibraryAccessor` enum. The README already describes this correctly.

This matters more than ordinary prose because the row explicitly justifies the compatibility obligation this new library product takes on. Please make the roadmap agree with the README/source: the externally reachable surface includes the PluginInspector surface **plus those four Library types**; only the implementation/nested declarations inside `LibraryAccessor` are capped internal.

### Small accuracy cleanup in the same update

- `README.md` and `CHANGELOG.md` still call the PR “manifest-only.” After the necessary `Package.resolved` and roadmap updates that is literally false. Keep the important claim — no source access-level, MCP, executable, or runtime-behaviour change — but call it a package-product/API-surface addition rather than manifest-only.
- Update the PR body's Logic-evidence/risk wording too: it currently says the change is only Package.swift + README, while the current diff also contains `Package.resolved`, CHANGELOG and roadmap metadata. The no-Logic-fact opt-out is still appropriate; only the changed-file description is stale.
- Non-blocking: “v3.16.0 is the current published stable release” will age. “v3.16.0 and earlier do not contain this product” is an evergreen way to explain why a branch/revision is needed until a tagged release carries it.
- Non-blocking and deliberately **not** part of this PR: #944 can close for the product declaration, but the offered mixer/insert-slot read-only facade currently has no follow-up issue. Please track it separately if we want that requester workflow to remain visible; do not implement it here.

### CI boundary

At review time on this head:
- `pr-policy`: **success**
- `guards`: **success** in the new code CI
- `formula`: **success**
- `compile`: **running**
- `test`: **running**
- Maintenance on the fork still shows `action_required`; it is not a merge gate, but this PR directly edits the roadmap, so its status check or the equivalent authenticated roadmap validation should be observed before calling the documentation verified.

Once the roadmap/API-surface statement is corrected, the stale “manifest-only” scope wording is cleaned up, and the current-head required CI is green, I do not see a reason to expand this contribution or block it on the follow-up facade. The optional duplicate-product regression check we discussed on #944 remains optional; do not add governance or a new packaging framework for this PR.

MongLong0214 (approved):

MongLong0214 (approved):
Approving as maintainer: the base ruleset has require_extra_approval_for_unattributed_changes on, and this branch now carries commits by someone other than its author. build, compile, test and pr-policy are green on e0532221.

